PR #5, opened 2025-11-03
Refactoring

jherr:
This PR adds

* Full agentic workflow with stop strategies
* Full tool support including approval process and client tools
* A new `connection` system that abstracts the ai-client from `fetch`
* Tool state support in the message types
* Now the message types are split between ModelMessage and UIMessage with converters between formats
* PHP server example
* PHP library for stream formatting (not sufficient, needs a full agentic workflow system)
* Python server example
* Python library for stream formatting (not sufficient, needs a full agentic workflow system)
* Vanilla JS client example on Vite (connects to PHP or Python)
* A Radix style ai-react-ui library that has a base chatbot implementation that you can style
* A new unit tested stream processing system
* Stream processing now includes a strategy pattern for chunk output timing (can chunk by sentences, chunk on punctuation, etc.) for smooth updates
* A refactored ts-chat application that removes unnecessary demos, adds debugging tools to the chat test system which is now the home row, and a demo of the `ai-react-ui` restylable components on /demo


packages/ai/src/tool-call-manager.ts
```diff
@@ -0,0 +1,174 @@
...
+  }
+
+  /**
+   * Get all complete tool calls (filtered for valid ID and name)
+   */
+  getToolCalls(): ToolCall[] {
+    return Array.from(this.toolCallsMap.values()).filter(
+      (tc) => tc.id && tc.function.name && tc.function.name.trim().length > 0
+    );
+  }
```
AlemTuzlak: if you did hasToolCalls and it returned true and then getToolCalls right after that it would return an empty array if there were only invalid tool calls inside of here

packages/typescript/ai-fallback/README.md
```diff
@@ -0,0 +1,143 @@
...
+for await (const chunk of stream) {
+  if (chunk.type === 'content') {
+    console.log(chunk.delta);
+  }
+}
+```
+
+## API
+
+### `withModel(ai, options)`
```
AlemTuzlak, line 47: i think this API should be moved inside of the ai wrapper itself maybe

TOOL_STATES_MIGRATION.md
```diff
@@ -0,0 +1,274 @@
...
+  {
+    id: "1",
+    role: "user",
+    parts: [{ type: "text", content: "Hello" }]
+  }
+];
+
+// Automatically converted to ModelMessages when sent
+const connection = fetchServerSentEvents("/api/chat");
+const stream = connection.connect(messages); // Converts internally
```
AlemTuzlak: a comment with an example converted output would be great here

packages/typescript/ai-react-ui/src/index.ts
```diff
@@ -0,0 +1,61 @@
...
+import { ChatMessages } from "./chat-messages";
+import { ChatMessage } from "./chat-message";
+import { ChatInput } from "./chat-input";
+import { ToolApproval } from "./tool-approval";
+
+// Attach subcomponents
+(Chat as any).Messages = ChatMessages;
+(Chat as any).Message = ChatMessage;
+(Chat as any).Input = ChatInput;
+(Chat as any).ToolApproval = ToolApproval;
```
AlemTuzlak, line 46: I wouldn't do this, would kill shedding

PR #22, opened 2025-11-25
Moving to functions

jherr:
* Removing the AI class and moving to functions `chat`, `embedding` and `summarize`
* Removing `chatCompletion`
* Cleaning up the file structures in `ai`
* Removing the CLI
* Updating docs and tests to match


docs/adapters/openai.md
```diff
@@ -80,9 +87,14 @@ export async function POST(request: Request) {
 ## Example: With Tools
 
 ```typescript
-import { tool } from "@tanstack/ai";
+import { chat, tool } from "@tanstack/ai";
+import { openai } from "@tanstack/ai-openai";
 import { z } from "zod";
 
+const adapter = openai({
+  apiKey: process.env.OPENAI_API_KEY!,
```
AlemTuzlak, line 95: should we leave this in like this in the docs? these utilities check for the apiKey under the hood with this exact name, it would just look cleaner if it was `const adapter = openai()`

PR #48, opened 2025-11-29
WIP: Tool overhaul

jherr:
`toolFactory` because `toolDefinition` and now you can use `.server((args) => ...)` or `.client((args) => ...)`. I didn't see the need for the `{ execute: ... }`.

There is also now a `createServerFnTool` which creates both a server function and a tool from the sample implementation.

There is also a `createChatClientOptions` function that makes it easier to get typed messages:

```ts
const chatOptions = createChatClientOptions({
  connection: fetchServerSentEvents('/api/tanchat'),
  tools: clientTools,
})

type ChatMessages = InferChatMessages<typeof chatOptions>
```

Which have all the information about the tools.


README.md
```diff
@@ -40,10 +40,36 @@ A powerful, type-safe AI SDK for building AI-powered applications.
...
 ### <a href="https://tanstack.com/ai">Read the docs →</b></a>
 
+## Bonus: TanStack Start Integration
+
+TanStack AI works with **any** framework (Next.js, Express, Remix, etc.).
+
+**With TanStack Start**, you get a bonus: share implementations between AI tools and server functions with `createServerFnTool`:
+
+```typescript
+import { createServerFnTool } from '@tanstack/ai-react'
```
AlemTuzlak, line 55: I would make this a sub export rather than mix it in with the react primitives, form did it this way and then had to refactor everything due to different bundling requirements later on

docs/api/ai.md
```diff
@@ -95,37 +95,57 @@ const result = await embedding({
...
 });
+
+// Use directly in chat() (server-side, no execute)
+chat({
+  tools: [myToolDef],
+  // ...
+});
+
+// Or create server implementation
+const myServerTool = myToolDef.server(async ({ param }) => {
```
AlemTuzlak, line 131: I would put examples here of passing it in to chat and what not so its clear to the user where he puts this after he defines the definition

docs/getting-started/overview.md
```diff
@@ -2,30 +2,82 @@
 
 TanStack AI is a powerful, type-safe AI SDK for building AI-powered applications with React, Node.js, and other JavaScript frameworks.
 
+## Works With Any Framework
+
+TanStack AI works great with:
+- **Next.js** - API routes and App Router
+- **TanStack Start** - React Start or Solid Start (recommended!)
+- **Express** - Node.js server
+- **Remix** - Loaders and actions
```
AlemTuzlak: I would say Remix / React Router v7 here

docs/guides/client-tools.md
```diff
@@ -167,8 +260,43 @@ const displayPreferences = tool({
 - **Browser APIs** - Access geolocation, camera, clipboard
 - **State Management** - Update React/Vue/Solid state
 - **Navigation** - Change routes, scroll to sections
+- **Analytics** - Track user interactions
+
+## Comparison to Old API
+
+**Before (onToolCall):**
```
AlemTuzlak: Remove the comparison

docs/guides/server-tools.md
```diff
@@ -139,23 +153,29 @@ You don't need to manually handle tool execution - it's automatic!
...
-  execute: async ({ userId }) => {
-    try {
-      const user = await db.users.findUnique({ where: { id: userId } });
-      if (!user) {
-        return { error: "User not found" };
-      }
-      return { name: user.name, email: user.email };
-    } catch (error) {
-      return { error: "Failed to fetch user data" };
+  outputSchema: z.object({
```
AlemTuzlak, line 162: Maybe make the schema here a discriminated union to make it even a better example so they have to check for the error before accessing data

PR #60, opened 2025-12-02
Feat/ modality for chat inputs

AlemTuzlak:
## 🎯 Changes

Chat now supports:
- audio, video, document and image inputs
- typesafe messages with autocomplete
- warnings about invalid modularities based on model
- typesafe metadata per modality

## ✅ Checklist

## 🚀 Release Impact

### DIFF
--- a/packages/typescript/ai-anthropic/tests/model-meta.test.ts
+++ b/packages/typescript/ai-anthropic/tests/model-meta.test.ts
@@ -68,8 +68,7 @@ describe('Anthropic Model Provider Options Type Assertions', () => {
     })
 
     it('claude-haiku-4-5 should support thinking options', () => {
-      type Options =
-        AnthropicChatModelProviderOptionsByName['claude-haiku-4-5']
+      type Options = AnthropicChatModelProviderOptionsByName['claude-haiku-4-5']
 
       expectTypeOf<Options>().toExtend<AnthropicThinkingOptions>()
       expectTypeOf<Options>().toExtend<AnthropicServiceTierOptions>()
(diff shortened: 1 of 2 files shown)


packages/typescript/ai-anthropic/src/anthropic-adapter.ts
```diff
@@ -232,6 +253,66 @@ export class Anthropic extends BaseAdapter<
     return requestParams
   }
 
+  private convertContentPartToAnthropic(
```
jherr, line 256: Can we move this to anthropic as an exported helper function?

packages/typescript/ai-anthropic/src/anthropic-adapter.ts
```diff
@@ -292,9 +376,28 @@ export class Anthropic extends BaseAdapter<
         continue
       }
 
+      // Handle user messages with multimodal content
+      if (role === 'user' && Array.isArray(message.content)) {
```
jherr, line 380: this all feels like it should either be in the anthropic adapter and/or normalized through the client. i don't think we should be doing much in terms of massaging message formats way up at the client layer.

packages/typescript/ai-anthropic/tests/model-meta.test.ts
```diff
@@ -464,3 +475,170 @@ describe('Anthropic Model Provider Options Type Assertions', () => {
...
+ * content parts based on each Anthropic model's supported input modalities.
+ *
+ * All Claude models support: text, image, document
+ * No Claude models support: audio, video
+ */
+describe('Anthropic Model Input Modality Type Assertions', () => {
+  // Helper type for creating a user message with specific content
+  type MessageWithContent<T> = { role: 'user'; content: Array<T> }
+
+  describe('Claude Opus 4.5 (text + image + document)', () => {
```
jherr, line 485: these seem kind of redundant, testing wise. maybe just test one or two. also, these dates on the end of the models change, which could be a PITA to maintain.

packages/typescript/ai-anthropic/src/anthropic-adapter.ts
AlemTuzlak: Rewolving

packages/typescript/ai-anthropic/tests/model-meta.test.ts
AlemTuzlak: great idea, from their docs i got the idea that we HAVE to use dates, but that's not true, removed them completely! I agree they are a bit redundant but I like redundant tests for these cases, I'd rather check for everything than not check 1 model and it turns out that one doesn't work 😥

PR #64, opened 2025-12-03
Ollama adapter

harry-whorlow:
Strongly types ollama models and configuration.


packages/typescript/ai-ollama/src/model-meta-llama.ts
```diff
@@ -0,0 +1,971 @@
+// originates -> ollama/llama/llama.cpp/include/llama.h
+// located -> ollama/llama/llama.go
+interface _GoLlamaContextParams {
```
AlemTuzlak, line 3: can we remove these go interfaces

harry-whorlow: yeah, they will be removed before merge

PR #84, opened 2025-12-05
feat: Add Vue `useChat()` support

wobsoriano:
## 🎯 Changes

This PR adds `useChat()` for Vue! The UI components will come in a separate PR.

## ✅ Checklist

## 🚀 Release Impact

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,6 +7,11 @@
   },
   "packageManager": "pnpm@10.17.0",
   "type": "module",
+  "pnpm": {
+    "overrides": {
+      "abbrev": "^3.0.0"
+    }
+  },
   "scripts": {
     "clean": "pnpm --filter \"./packages/**\" run clean",
     "test": "pnpm run test:ci",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -4,6 +4,9 @@ settings:
   autoInstallPeers: true
   excludeLinksFromLockfile: false
 
+overrides:
+  abbrev: ^3.0.0
+
 importers:
 
   .:
@@ -3094,10 +3097,6 @@ packages:
   '@zkochan/js-yaml@0.0.7':
     resolution: {integrity: sha512-nrUSn7hzt7J6JWgWGz78ZYI8wj+gdIJdk0Ynjpp8l+trkn58Uqsf6RYrYkEK+3X18EX+TNdtJI0WxAtc+L84SQ==}
     hasBin: true
-
-  abbrev@2.0.0:
-    resolution: {integrity: sha512-6/mh1E2u2YgEsCHdY0Yx5oW+61gZU+1vXaoiHHrpKeuRNNgFvS+/jrwHiQhB5apAf5oB7UB7E19ol2R2LKH8hQ==}
-    engines: {node: ^14.17.0 || ^16.13.0 || >=18.0.0}
 
   abbrev@3.0.1:
     resolution: {integrity: sha512-AO2ac6pjRB3SJmGJo+v5/aK6Omggp6fsLrs6wN9bd35ulu4cCwaAU9+7ZhXjeqHVkaHThLuzH0nZr0YpCDhygg==}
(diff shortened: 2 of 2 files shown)


packages/typescript/ai-vue/src/use-chat.ts
```diff
@@ -0,0 +1,106 @@
+import { ChatClient } from '@tanstack/ai-client'
+import { onScopeDispose, readonly, shallowRef, useId } from 'vue'
+import type { AnyClientTool, ModelMessage } from '@tanstack/ai'
+import type { UIMessage, UseChatOptions, UseChatReturn } from './types'
+
+export function useChat<TTools extends ReadonlyArray<AnyClientTool> = any>(
+  options: UseChatOptions<TTools> = {} as UseChatOptions<TTools>,
+): UseChatReturn<TTools> {
+  const hookId = useId() // Available in Vue 3.5+
```
wobsoriano, line 9: If we want to support Vue <3.5, we can use a random ID generator instead

AlemTuzlak:
@wobsoriano for the provenance issues I suggest copying the lock from main over to your branch and running pnpm install

packages/typescript/ai-vue/package.json
```diff
@@ -0,0 +1,61 @@
+{
+  "name": "@tanstack/ai-vue",
+  "version": "0.0.1",
```
AlemTuzlak: this should be 0.0.0 so the patch releases it as 0.0.1

packages/typescript/ai-vue/src/use-chat.ts
DamianOsipiuk: I guess this should be fine unless someone explicitly asks. Upgrading 3.3 to 3.5 should not be a problem though?

packages/typescript/ai-vue/src/use-chat.ts
```diff
@@ -0,0 +1,106 @@
+import { ChatClient } from '@tanstack/ai-client'
+import { onScopeDispose, readonly, shallowRef, useId } from 'vue'
+import type { AnyClientTool, ModelMessage } from '@tanstack/ai'
+import type { UIMessage, UseChatOptions, UseChatReturn } from './types'
+
+export function useChat<TTools extends ReadonlyArray<AnyClientTool> = any>(
+  options: UseChatOptions<TTools> = {} as UseChatOptions<TTools>,
```
DamianOsipiuk, line 7: @AlemTuzlak 
Do we need options to be reactive via `getter`?
Is there anything important we could `change` dynamically without breaking the integration?

DamianOsipiuk (approved):

packages/typescript/ai-vue/src/use-chat.ts
AlemTuzlak: I could see some possible scenarios where we'd want to maybe change the option parameters between requests, this would require changes here right?

DamianOsipiuk: Yeah, we would need to add a `getter` around the options and handle that in code. Or go for specific `MaybeRef` option parameters instead of all of them.

AlemTuzlak: I think that's a good thing to add in case we need it later on, if it's not a big technical feat

wobsoriano: Marking the options as reactive and putting the client in a computed property will recreate the instance. Do we want that? The React implementation has this comment:

```ts
// Create ChatClient instance with callbacks to sync state
// Note: Options are captured at client creation time.
// The connection adapter can use functions for dynamic values (url, headers, etc.)
// which are evaluated lazily on each request.
```

and the client is only recreated when `clientId` is updated.

packages/typescript/ai-vue/src/use-chat.ts
wobsoriano: yep! no problems here

packages/typescript/ai-vue/src/use-chat.ts
AlemTuzlak: Ahh okay, nvm you're right ✅️

AlemTuzlak (approved):

PR #93, opened 2025-12-05
feat: context aware tools

ahmedrowaihi:
## 🎯 Changes

Supporting context in tool execution..

to make tools behaviour context aware, the only way is to reconstruct them everytime you want to invoke a new trigger..
this PR suggests supporting context.

## ✅ Checklist

- [x] I have tested this code locally with `pnpm run test:lib`.

## 🚀 Release Impact

## Release Notes

* **New Features**
  * Tools now receive context during execution through an options parameter, available in both client and server environments
  * Context can be passed via chat client options and is propagated to all tool executions

* **Tests**
  * Added and updated tests to verify context propagation to tool executions

### DIFF
--- a/packages/typescript/ai-client/src/chat-client.ts
+++ b/packages/typescript/ai-client/src/chat-client.ts
@@ -32,6 +32,7 @@ export class ChatClient {
   private uniqueId: string
   private body: Record<string, any> = {}
   private pendingMessageBody: Record<string, any> | undefined = undefined
+  private context: unknown = undefined
   private isLoading = false
   private isSubscribed = false
   private error: Error | undefined = undefined
@@ -81,6 +82,7 @@ export class ChatClient {
   constructor(options: ChatClientOptions) {
     this.uniqueId = options.id || this.generateUniqueId('chat')
     this.body = options.body || {}
+    this.context = options.context
     this.connection = normalizeConnectionAdapter(options.connection)
     this.events = new DefaultChatClientEventEmitter(this.uniqueId)
 
@@ -202,7 +204,7 @@ export class ChatClient {
             // Create and track the execution promise
             const executionPromise = (async () => {
               try {
-                const output = await executeFunc(args.input)
+                const output = await executeFunc(args.input, { userContext: this.context })
                 await this.addToolResult({
                   toolCallId: args.toolCallId,
                   tool: args.toolName,
@@ -948,6 +950,7 @@ export class ChatClient {
   updateOptions(options: {
     connection?: ConnectionAdapter
     body?: Record<string, any>
+    context?: unknown
     tools?: ReadonlyArray<AnyClientTool>
     onResponse?: (response?: Response) => void | Promise<void>
     onChunk?: (chunk: StreamChunk) => void
@@ -985,6 +988,9 @@ export class ChatClient {
     }
     if (options.body !== undefined) {
       this.body = options.body
+    }
+    if (options.context !== undefined) {
+      this.context = options.context
     }
     if (options.tools !== undefined) {
       this.clientToolsRef.current = new Map()
--- a/packages/typescript/ai-client/src/types.ts
+++ b/packages/typescript/ai-client/src/types.ts
@@ -185,6 +185,7 @@ export interface UIMessage<TTools extends ReadonlyArray<AnyClientTool> = any> {
 
 export interface ChatClientOptions<
   TTools extends ReadonlyArray<AnyClientTool> = any,
+  TContext = unknown,
 > {
   /**
    * Connection adapter for streaming.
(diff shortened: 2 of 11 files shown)


packages/typescript/ai/src/tools/tool-definition.ts
```diff
@@ -100,20 +105,22 @@ export interface ToolDefinition<
   /**
    * Create a server-side tool with execute function
    */
-  server: (
+  server: <TContext = unknown>(
     execute: (
       args: z.infer<TInput>,
+      context?: TContext,
```
jherr: I'd prefer it if we had an optional `options` argument here where one of the potential keys is `context`. That keeps the door open for more things later on without adding more arguments.

packages/typescript/ai/src/tools/tool-definition.ts
```diff
@@ -177,27 +184,31 @@ export function toolDefinition<
   const definition: ToolDefinition<TInput, TOutput, TName> = {
     __toolSide: 'definition',
     ...config,
-    server(
+    server<TContext = unknown>(
       execute: (
         args: z.infer<TInput>,
+        context?: TContext,
```
jherr: Again, `options`.

packages/typescript/ai/src/tools/tool-definition.ts
```diff
@@ -177,27 +184,31 @@ export function toolDefinition<
...
-        execute,
+        execute: execute as any,
       }
     },
 
-    client(
+    client<TContext = unknown>(
       execute?: (
         args: z.infer<TInput>,
+        context?: TContext,
```
jherr: And once more `options`.

jherr (changes requested):
One small API change, otherwise I love it!

packages/typescript/ai-client/src/chat-client.ts
```diff
@@ -298,10 +310,13 @@ export class ChatClient {
       // Call onResponse callback
       await this.callbacksRef.current.onResponse()
 
-      // Include conversationId in the body for server-side event correlation
+      // Include conversationId and context in the body for server-side event correlation
       const bodyWithConversationId = {
         ...this.body,
         conversationId: this.uniqueId,
+        ...(this.options.context !== undefined && {
+          context: this.options.context,
```
AlemTuzlak: i am just wondering if this is a good idea to do, if you put localStorage in context and we try to send it over the wire what would even happen? do we want to serialize/deserialize this? or just send simple things or not send it at all?

ahmedrowaihi:
@AlemTuzlak , do you think we better also pass current _serialisable_only_ within the request ?

tombeckenham:
Hi @ahmedrowaihi — thanks for the contribution. Unfortunately this branch is significantly stale (~50 commits behind `main`) and several things have changed underneath it that make the diff hard to land as-is. Posting a review here so the work isn't lost, but I'm going to close this in favor of a fresh patch against current `main`.

## What `main` already has

- `chat({ context })` exists today — `ChatOptions.context?: unknown` flows into `ChatMiddlewareContext.context` (see `middleware.test.ts:793`). The field name you chose already collides with this.
- `ToolExecutionContext` exists with **required** `emitCustomEvent` (relied on by production code in `executeToolCalls`).
- `ClientTool` is intentionally a *sibling* of `Tool` (no `extends`) — narrower contract, different `execute` signature.
- Strict TypeScript + double-cast ban shipped in #564.
- `ChatClient` was reworked (`fetcher` option #512, `forwardedProps`/`body` split, `threadId`, `resolveTransport`).
- Tool-execution flow was reworked by #596 and #609.

## What this PR got right

The core idea — forwarding user-supplied context to tool `execute` so tools don't need to capture dependencies in closures — is still a real gap and worth landing.

## Issues with the current diff vs `main`

**Architectural / API**
1. `ClientTool extends Tool` is a regression. Main keeps these as siblings on purpose; widening client tools to accept the server `execute(args, context)` signature blurs the discriminant.
2. Making `ToolExecutionContext.emitCustomEvent` optional is unnecessary and harmful. The production `executeToolCalls` on main always provides it; this change only exists so `ToolCallManager.executeTools` can build a half-context (`{ toolCallId, userContext }`), which silently disables `emitCustomEvent` for any tool routed through that path.
3. `ChatClientOptions<TTools, TContext>` adds a generic that's never threaded into `ChatClient` itself, so `TContext` is always erased to `unknown` — dead type parameter.
4. Asymmetry: `TextOptions.context: unknown` (no generic) vs `ChatClientOptions<…, TContext>`. After #564 the bar for new public surface is end-to-end typing; either both sides get the generic or neither does.
5. `chat({ context })` is *added* as if new, but main already has it for middleware — these should be unified, not duplicated.

**Code quality**
6. `(args: any, context: any) => execute(args, context)` in `tool-definition.ts` — would fail #564's strict-TS / no-double-cast rules on rebase. Also unnecessary: the wrappers `(args, ctx) => execute(args, ctx)` are pure no-ops; just assign `execute` directly.
7. `executeToolCalls` signature would become 7 positional args after this — past the point where it should be an options object.
8. `updateOptions({ context: undefined })` can't clear context (guarded by `!== undefined`). Minor, but worth documenting or using `'context' in options`.

**Test coverage gaps**
9. No end-to-end test that `chat({ context })` → server tool receives `userContext` via the real `TextEngine` path. The new `tool-definition.test.ts` tests pass a hand-built context directly to `execute()` — they validate the type signature but wouldn't catch a regression in the wiring.
10. No `ChatClient` test that the constructor's `context` reaches a client-tool execute.

## Recommended path forward

Open a new PR against current `main` with a minimal patch (~50 lines, not 250):

1. Reuse the existing `ChatOptions.context` field — don't add a new one.
2. Add `userContext?: unknown` to `ToolExecutionContext` (keep `emitCustomEvent` required).
3. In `executeToolCalls`, set `userContext: ctx.context` on the per-tool context alongside the existing `emitCustomEvent`.
4. Add an optional `context?` param to `ClientTool.execute` directly — keep it as a sibling of `Tool`, no `extends`.
5. Thread `ChatClient.context` (constructor + `updateOptions`) into the client-tool execute call.
6. Add two integration tests: one end-to-end via `chat()` for server tools, one via `ChatClient` for client tools.

If you'd like to take a stab at the rebased version, happy to review. Closing this one to keep the queue clear — please reopen as a fresh PR.

PR #116, opened 2025-12-07
(WIP) feat: aws bedrock adapter

chunterb:
## ✅ Checklist

## 🚀 Release Impact

### DIFF
--- a/packages/typescript/ai-bedrock/src/bedrock-adapter.ts
+++ b/packages/typescript/ai-bedrock/src/bedrock-adapter.ts
@@ -1,0 +1,1157 @@
+import { AwsV4Signer } from 'aws4fetch'
+import { EventStreamCodec } from '@smithy/eventstream-codec'
+import { fromUtf8, toUtf8 } from '@smithy/util-utf8'
+import { BaseAdapter } from '@tanstack/ai'
+import {
+  BEDROCK_EMBEDDING_MODELS,
+  BEDROCK_MODELS,
+} from './model-meta'
+import { convertToolsToProviderFormat } from './tools/tool-converter'
+import type {
+  ChatOptions,
+  ContentPart,
+  EmbeddingOptions,
+  EmbeddingResult,
+  ModelMessage,
+  StreamChunk,
+  SummarizationOptions,
+  SummarizationResult,
+} from '@tanstack/ai'
+import type {
+  BedrockChatModelProviderOptionsByName,
+  BedrockModelInputModalitiesByName,
+} from './model-meta'
+import type { BedrockProviderOptions } from './text/text-provider-options'
+import type {
+  BedrockDocumentFormat,
+  BedrockDocumentMetadata,
+  BedrockImageFormat,
+  BedrockImageMetadata,
+  BedrockMessageMetadataByModality,
+  BedrockVideoFormat,
+  BedrockVideoMetadata,
+} from './message-types'
+
+/**
+ * AWS credentials returned by a credential provider
+ */
+export interface BedrockCredentials {
+  accessKeyId: string
+  secretAccessKey: string
+  sessionToken?: string
+}
+
+interface BedrockTextBlock {
+  text: string
+}
+
+interface BedrockImageBlock {
+  image: {
+    format: BedrockImageFormat
+    source: { bytes: string } | { s3Location: { uri: string; bucketOwner?: string } }
+  }
+}
+
+interface BedrockVideoBlock {
+  video: {
+    format: BedrockVideoFormat
+    source: { bytes: string } | { s3Location: { uri: string; bucketOwner?: string } }
+  }
+}
+
+interface BedrockDocumentBlock {
+  document: {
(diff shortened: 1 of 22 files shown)


packages/typescript/ai-bedrock/src/model-meta.ts
```diff
@@ -0,0 +1,1830 @@
...
+  BedrockAnthropicOptions,
+  BedrockAnthropicReasoningOptions,
+  BedrockProviderOptions,
+  BedrockReasoningEffortOptions,
+} from './text/text-provider-options'
+
+interface ModelMeta<TProviderOptions = unknown> {
+  name: string
+  id: string
+  supports: {
```
AlemTuzlak, line 32: I think this array should reflect the table found here:
https://docs.aws.amazon.com/bedrock/latest/userguide/conversation-inference-supported-models-features.html

packages/typescript/ai-bedrock/src/model-meta.ts
```diff
@@ -0,0 +1,1830 @@
...
+    input: ['text'],
+    output: ['text'],
+    tools: true,
+    streaming: true,
+  },
+} as const satisfies ModelMeta<BedrockProviderOptions>
+
+const US_LLAMA_4_SCOUT = {
+  name: 'Llama 4 Scout 17B Instruct (US)',
+  id: 'us.meta.llama4-scout-17b-instruct-v1:0',
```
AlemTuzlak: the id here seems tied to the region?

chunterb: Yes some of the newer models have inference profiles, including Anthropic ones. https://docs.aws.amazon.com/bedrock/latest/userguide/inference-profiles-support.html

Maybe should have something typed for regions since it's consistent across all models that require it. I have not added any of the EU models, etc yet.

packages/typescript/ai-bedrock/src/model-meta.ts
chunterb: @AlemTuzlak I've added this, but curious if this adapter should diverge from how we determine the modalities that a model has, when compared to other adapters (type map)?

I've left the type map for now.

packages/typescript/ai-bedrock/src/model-meta.ts
chunterb: To simplify, I've left only the ones where the inference profile is required for now.

packages/typescript/ai-bedrock/src/model-meta.ts
AlemTuzlak: sorry, we didn't understand each other, yes what you added is what I wanted to see, but I meant in addition to input/output. For example in some model-meta i've done supports.capabilities.[all of the capabilities] and then you can do supports.input as well

packages/typescript/ai-bedrock/src/model-meta.ts
```diff
@@ -0,0 +1,1710 @@
...
+    streamingToolUse: true,
+    guardrails: true,
+    s3Links: true,
+    reasoning: true,
+  },
+} as const satisfies ModelMeta<BedrockProviderOptions & BedrockReasoningEffortOptions>
+
+const US_NOVA_2_LITE = {
+  name: 'Nova 2 Lite',
+  id: 'us.amazon.nova-2-lite-v1:0',
```
AlemTuzlak: i think we should remove the regions from ID's because they are resolved via provided config, we can just also map what models support what regions and append the region based on that, otherwise use the normal id

packages/typescript/ai-bedrock/src/bedrock-adapter.ts
```diff
@@ -0,0 +1,1116 @@
...
+  /**
+   * AWS region for Bedrock service.
+   * Used to construct the endpoint URL.
+   *
+   * If not provided, defaults to:
+   * 1. `AWS_REGION` environment variable
+   * 2. `AWS_DEFAULT_REGION` environment variable
+   * 3. `'us-east-1'` as final fallback
+   */
+  region?: string
```
AlemTuzlak, line 127: can we extract the model region from here? so for example if it's eu/us/jp/sa and the model supports eu we can tell the user the model and the region don't match at runtime by throwing or default to a regionless model if it's supported?
eg:
model is anthropic, supported regions are eu and us, user specifies ca, it throws
model is anthropic, supported regions are eu and us, user specifices us, we append it to the model id before sending the request
model is anthropic, regionless, we send the request whatever region user specifies

chunterb: https://github.com/TanStack/ai/pull/116/commits/d6d66fbd93dac97000a83346fd072750ed933a76 

How's this? Added to the model config and check against that at runtime. 

```
inferenceProfile: { regions: ['us', 'eu', 'apac'], required: true },
```

Where `required` is optional because some models require the inference profile and others do not.

tombeckenham:
Hi @chunterb , this pr has been superceded by #665

PR #147, opened 2025-12-13
fix: unresponsive chat after client tool execution

harlley:
## 🎯 Changes

The conversation becomes unresponsive after a client-side tool execution, as the LLM fails to receive the tool result effectively.
- Updated `StreamProcessor.areAllToolsComplete` to check for `tool-result` parts (server tools) in addition to `output` fields (client tools).
- Updated `ChatClient.streamResponse` to check if the stream ended with a `tool-result` part. If so, it automatically triggers `continueFlow()` to send the result back to the LLM. This handles cases where `addToolResult` was called while the stream was still active (and thus blocked from continuing immediately).

## ✅ Checklist

## 🚀 Release Impact

## Summary by CodeRabbit

* **Bug Fixes**
  * Better detection of tool execution completion so server-side tool results mark calls as finished, reducing stalled actions.
  * Streaming responses now, after a successful stream, automatically continue the conversation when the last streamed item is a tool result and auto-send is enabled; continuation errors are logged but do not interrupt the flow.

* **Chores**
  * Patch version metadata updated for two packages.

### DIFF
--- a/.changeset/fix-silent-chat-continuation.md
+++ b/.changeset/fix-silent-chat-continuation.md
@@ -1,0 +1,6 @@
+---
+"@tanstack/ai": patch
+"@tanstack/ai-client": patch
+---
+
+fix: Continue conversation after client tool execution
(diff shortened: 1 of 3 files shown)


AlemTuzlak:
@harlley I think this is severly outdated and potentially not a problem anymore, if it is please let us know!

PR #158, opened 2025-12-17
Feat/move model into adapter fix types and cleanup

tannerlinsley:
## ✅ Checklist

## 🚀 Release Impact


packages/typescript/ai-gemini/src/adapters/text.ts
```diff
@@ -47,40 +48,61 @@ export interface GeminiTextConfig extends GeminiClientConfig {}
...
+    : GeminiTextProviderOptions
+
+/**
+ * Resolve input modalities for a specific model.
+ * If the model has explicit modalities in the map, use those; otherwise use all modalities.
+ */
+type ResolveInputModalities<TModel extends string> =
+  TModel extends keyof GeminiModelInputModalitiesByName
+    ? GeminiModelInputModalitiesByName[TModel]
+    : readonly ['text', 'image', 'audio', 'video', 'document']
```
AlemTuzlak, line 71: i think we should use text and image only here as teh default or even never or something that will cause an issue that we  know we didn't add in the models modalities properly

PR #234, opened 2026-01-18
fix: add missing Anthropic models type and update the anthropic docs

fuadnafiz98:
### WHAT

Added missing `ANTHROPIC_MODELS` type export to the public API of `@tanstack/ai-anthropic` and updated the documentation to correctly reference it.

The `ANTHROPIC_MODELS` is a const tuple that contains all supported Anthropic model identifiers:

```typescript
type ANTHROPIC_MODELS = readonly [
  'claude-opus-4-5',
  'claude-sonnet-4-5',
  'claude-haiku-4-5',
  'claude-opus-4-1',
  'claude-sonnet-4',
  'claude-sonnet-3-7',
  'claude-opus-4',
  'claude-haiku-3-5',
  'claude-haiku-3',
]
```

### WHY

When I tried to use the `createAnthropicChat` according to the docs it gave me type errors and I find out that the first parameter it takes is the model name, not the API KEY as the docs suggest. 

Then I find out that there is actually a type `ANTHROPIC_MODELS` which is not exported that I can use to construct the adapter. 

### HOW - Consumers Should Update

Now you can import and use `ANTHROPIC_MODELS` for proper type safety when creating adapter instances:

```typescript
import { createAnthropicChat, ANTHROPIC_MODELS } from '@tanstack/ai-anthropic'

const adapter = (model: ANTHROPIC_MODELS) =>
  createAnthropicChat(model, process.env.ANTHROPIC_API_KEY!, {
    // ... your config options
  })

const stream = chat({
  adapter: adapter('claude-sonnet-4-5'), // Type-checked model selection!
  messages: [{ role: 'user', content: 'Hello!' }],
})
```

## ✅ Checklist

## 🚀 Release Impact

## Summary by CodeRabbit

* **Breaking Changes**
  * The Anthropic adapter factory now requires the model identifier as the first argument when creating an instance.

* **New Features**
  * Public, type-safe export for Anthropic model identifiers.
  * Adapter usage updated to support runtime model selection with improved type safety.

### DIFF
--- a/.changeset/add-anthropic-models-export.md
+++ b/.changeset/add-anthropic-models-export.md
@@ -1,43 +1,32 @@
 ---
-'@tanstack/ai-anthropic': minor
+'@tanstack/ai-anthropic': major
 ---
 
-## Add missing ANTHROPIC_MODELS type export
+## Add missing AnthropicModels type export
 
 ### WHAT
 
-Added missing `ANTHROPIC_MODELS` type export to the public API of `@tanstack/ai-anthropic` and updated the documentation to correctly reference it.
-
-The `ANTHROPIC_MODELS` is a const tuple that contains all supported Anthropic model identifiers:
+Added new `AnthropicModels` **type** export to the public API of `@tanstack/ai-anthropic`. This is a union type that represents all supported Anthropic model identifiers, derived from the internal `ANTHROPIC_MODELS` const tuple.
 
 ```typescript
-type ANTHROPIC_MODELS = readonly [
-  'claude-opus-4-5',
-  'claude-sonnet-4-5',
-  'claude-haiku-4-5',
-  'claude-opus-4-1',
-  'claude-sonnet-4',
-  'claude-sonnet-3-7',
-  'claude-opus-4',
-  'claude-haiku-3-5',
-  'claude-haiku-3',
-]
+export type AnthropicModels = (typeof ANTHROPIC_MODELS)[number]
+// Equivalent to: 'claude-opus-4-5' | 'claude-sonnet-4-5' | 'claude-haiku-4-5' | ... (and more)
 ```
+
+**Note:** The `ANTHROPIC_MODELS` const tuple itself remains internal and is not exported. Only the derived type is part of the public API.
 
 ### WHY
 
-The `ANTHROPIC_MODELS` type was already used internally and in the model metadata but was not exported from the main package entry point. This prevented consumers from using it for type-safe model selection when creating custom adapter instances, resulting in incomplete TypeScript support and requiring workarounds like manual type assertions.
-
-By exporting this type, consumers now have full type safety when working with Anthropic models and can write more maintainable code.
+Consumers previously had no easy way to get the type-safe union of model names for use in function signatures and variable declarations.
 
 ### HOW - Consumers Should Update
 
-Now you can import and use `ANTHROPIC_MODELS` for proper type safety when creating adapter instances:
+Now you can import and use `AnthropicModels` for proper type safety when creating adapter instances:
 
 ```typescript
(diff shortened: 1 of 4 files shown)


docs/adapters/anthropic.md
```diff
@@ -28,11 +28,12 @@ const stream = chat({
 
 ```typescript
 import { chat } from "@tanstack/ai";
-import { createAnthropicChat } from "@tanstack/ai-anthropic";
+import { createAnthropicChat, AnthropicModels } from "@tanstack/ai-anthropic";
 
-const adapter = createAnthropicChat(process.env.ANTHROPIC_API_KEY!, {
-  // ... your config options
-});
+const adapter = (model: AnthropicModels) =>
```
AlemTuzlak, line 33: revert this change please

.changeset/add-anthropic-models-export.md
```diff
@@ -0,0 +1,64 @@
...
+
+const stream = chat({
+  adapter: adapter('claude-sonnet-4-5'), // Type-checked model selection!
+  messages: [{ role: 'user', content: 'Hello!' }],
+})
+```
+
+The type ensures only valid Anthropic model identifiers can be passed, preventing runtime errors.
+
+### Breaking Change
```
AlemTuzlak, line 42: remove the breaking change section as it's not a breaking change

.changeset/add-anthropic-models-export.md
```diff
@@ -0,0 +1,64 @@
+---
+'@tanstack/ai-anthropic': major
```
AlemTuzlak, line 2: this is not a major change, this hsould be set to patch

docs/adapters/anthropic.md
fuadnafiz98: Hi, 
Thanks for your review. 

Correct me if I am wrong, but when I was trying to use `createAnthropicChat` it gives me errors as of the previous docs example. When I see the function definition [here](https://github.com/TanStack/ai/blob/65a679663d11eef5a0b3adac3b6b94faa5d39c10/packages/typescript/ai-anthropic/src/adapters/text.ts#L628) I can see that it requires the model in the first parameter.

fuadnafiz98: If that is the case then shouldn't the docs be updated?

AlemTuzlak: that is true but it should be:
```ts
const adapter = createAnthropicChat("model", process.env.ANTHROPIC_API_KEY!, {...})
```

fuadnafiz98: Hi, 
About the example, we can either put the example like this: 

```typescript
import { createAnthropicChat } from '@tanstack/ai-anthropic'

const adapter = createAnthropicChat(
  'claude-sonnet-4-5',
  process.env.ANTHROPIC_API_KEY!,
  {
    // ... your config options
  },
)

const stream = chat({
  adapter: adapter,
  messages: [{ role: 'user', content: 'Hello!' }],
})
```

here the adapter is connected with a single model and we pass it to the `chat`, but in the other examples in the docs I can see we can change the model names in the `chat` so its more flexible to try out different model. 

So if we put the example like this: 

```typescript 
const adapter = (model: AnthropicModels) =>
  createAnthropicChat(model, process.env.ANTHROPIC_API_KEY!, {
    // ... your config options
  })

const stream = chat({
  adapter: adapter('claude-3-5-haiku'),
  messages: [{ role: 'user', content: 'Hello!' }],
})

```
We can easily swap out different models in the chat like this:

image


Let me know what do you think and if I am thinking in a wrong way :pray:

fuadnafiz98: Hello @AlemTuzlak can you please have a look 
Thanks.

AlemTuzlak:
I've added this on main already

PR #242, opened 2026-01-22
feat: improve usage

AlemTuzlak:
## 🎯 Changes

Add usage details that were missing

## ✅ Checklist

## 🚀 Release Impact

## Release Notes

* **New Features**
  * Enhanced token usage reporting with detailed breakdowns across all AI providers, including cached tokens, reasoning tokens, and request duration.
  * Added granular token tracking by modality (text, image, audio, video) for supported providers.
  * Extended dev tools UI to display comprehensive per-message usage metrics.

* **Documentation**
  * Updated protocol documentation with new token usage schema and provider-specific usage examples.


packages/typescript/ai/src/types.ts
```diff
@@ -692,14 +692,75 @@ export interface ToolResultStreamChunk extends BaseStreamChunk {
   content: string
 }
 
+/**
+ * Detailed breakdown of prompt/input token usage.
+ * Fields are populated based on provider support.
+ */
+export interface PromptTokensDetails {
+  /** Tokens read from cache (OpenAI, OpenRouter, Gemini) */
```
jherr: We don't want to be updating these lists of adapters as we grow out the numbers.

packages/typescript/ai-anthropic/src/adapters/text.ts
```diff
@@ -44,6 +45,66 @@ import type {
   AnthropicTextMetadata,
 } from '../message-types'
 import type { AnthropicClientConfig } from '../utils'
+import type { AnthropicProviderUsageDetails } from '../usage-types'
+
+/**
+ * Build normalized TokenUsage from Anthropic's usage object
+ */
+function buildAnthropicUsage(
```
jherr: Every one of the providers now has one or more of these functions. It would be good to see if we could extract common functionality. Also, let's consider creating a `usage.ts` file in each adapter that has the calculation functions. That keeps them in a known spot with a known function, and keeps the line count down in the action handlers.

jherr (changes requested):

packages/typescript/ai/src/types.ts
AlemTuzlak: you're absolutely right! jokes aside I wanted to extract some common denominators out of all the adapters and some ended up being duplicates, i've cleaned up the types now, thank you for catching this!

packages/typescript/ai-anthropic/src/adapters/text.ts
AlemTuzlak: done, also added tests for each adapter for every usage case

nikas-belogolov:
I've noticed that there are also no `aiEventClient.emit(...)` in upstream for the following events: `video:usage`, `transcription:usage`, `speech:usage` and `summarize:usage`. Should probably be also included in this pull request with the improvements of usage

pemontto:
We were about to open a PR for this exact thing, glad to see it already in flight. The nested `promptTokensDetails` / `completionTokensDetails` shape is nicer than the flat fields we were going to propose, and `providerUsageDetails` is a good escape hatch. In the meantime I've patched our local copy so our cost tracking isn't overcounting by 50-75% on cached multi-turn sessions. Anything we can do to help land this? Happy to rebase the conflicts, pick up the missing `aiEventClient.emit` wiring for `video:usage` / `transcription:usage` / `speech:usage` / `summarize:usage` that @nikas-belogolov flagged, or split the PR up if that'd make re-review easier.

tombeckenham:
Hey @AlemTuzlak — heads up that #654 just landed and introduces a shared `UsageTotals` interface (in `packages/ai/src/types.ts`) on the same `RunFinishedEvent.usage` surface this PR extends. The two PRs are fully complementary at the field level — no collisions:

- #654 added `cost?: number` and `costDetails?: UsageCostBreakdown` (cost only — provider-reported per-request cost, currently OpenRouter-only).
- This PR adds token granularity (`promptTokensDetails`, `completionTokensDetails`, `durationSeconds`, etc.).

The only overlap is the wrapper-interface name: #654 picked `UsageTotals`, this PR has `TokenUsage`. To unblock the merge, the cleanest path is probably:

1. **Rebase onto the new monorepo layout.** Paths flattened a while back: `packages/typescript/ai-*` → `packages/ai-*`. That's where most of this PR's conflicts are coming from.
2. **Drop the `TokenUsage` interface; add this PR's fields directly to `UsageTotals`** instead. Same three core fields (`promptTokens` / `completionTokens` / `totalTokens`) on both sides, just one wrapper name to keep. `PromptTokensDetails` and `CompletionTokensDetails` stay as-is.
3. **Widenings.** This PR also widens `TextCompletionChunk.usage`, `SummarizationResult.usage`, and adds `TranscriptionResult.usage`. #654 did not touch those — happy to keep those widenings here since they're tightly tied to the token-details change.

One question worth raising while you're rebasing: `providerUsageDetails?: Record<string, unknown>` on `TokenUsage`. During #654 review we deliberately moved away from a similar `Record<string, …>` shape for cost-details in favor of canonical named fields (the rationale was provider-portability — consumer code shouldn't have to know which adapter populated the data to read it). The named fields you already have (`cachedTokens`, `reasoningTokens`, the modality breakdowns) cover most of what providers stuff into an escape-hatch field today. Worth considering whether the residual is thin enough to defer to follow-up PRs that add named fields as new use cases come in, rather than ship the open record now. Not a blocker — your call.

Happy to help with the rebase if useful.

tombeckenham (approved):
I think this is good now. It would be good to sense check though to make sure the merge was correct and we like the name of the type. I'm not sure UsageTotals was the right name in retrospecet

tombeckenham (approved):
I'm made a couple of small changes. Approving

tombeckenham (approved):

packages/typescript/ai/src/types.ts
tombeckenham: Addressed — the provider-name lists were removed from the JSDoc (fields now read generically, e.g. `/** Tokens read from cache */`), so there's nothing per-provider to maintain as adapters grow. The usage-detail types (`PromptTokensDetails`/`CompletionTokensDetails`/`TokenUsage`) also now live in `@tanstack/ai-event-client` as the single source of truth, re-exported by `@tanstack/ai`.

packages/typescript/ai-anthropic/src/adapters/text.ts
tombeckenham: Done — each adapter now has a dedicated `src/usage.ts` holding its usage-calculation function (`ai-anthropic`, `ai-gemini`, `ai-ollama`, `ai-openrouter`, plus `openai-base` shared across OpenAI/Grok/Groq), so the action handlers stay lean and the calc lives in a known spot. Common functionality is extracted into `@tanstack/ai`'s `buildBaseUsage`, which every adapter builder calls.

jherr (approved):
LGTM

PR #303, opened 2026-02-20
fix(ai-client): prevent continuation stall when multiple client tools complete in the same round

### LINKED ISSUE #302: Client tool continuation stalls when multiple tools complete in the same round
## Disclosure

This issue was written by an AI coding agent (Claude Code) on behalf of a developer who encountered this bug. The analysis below is based on source-level tracing of the `@tanstack/ai-client` ChatClient code and observed runtime behavior.

## Versions

- `@tanstack/ai`: 0.5.1
- `@tanstack/ai-client`: 0.4.5
- `@tanstack/ai-vue`: 0.5.4
- Framework: Vue 3.5 with Composition API

## Description

When an LLM response triggers **multiple client-side tool calls** in the same round (e.g., the model calls `get_expenses` and `execute_sql` simultaneously), the chat gets stuck after all tools complete. The model never produces its follow-up text response. The `isLoading` flag goes `false` with the last message part being a `tool-result`, but no continuation request is sent to the server.

### Observed behavior

1. User sends a message like *"How much did I spend this year on food?"*
2. Round 1: Model calls `get_expenses` → client executes → result returned → continuation fires correctly ✓
3. Round 2: Model calls `get_expenses` (drill-down) **and** `execute_sql` in the same response → both tools execute → both results appear in the UI → **chat gets permanently stuck**
4. The "Stop generating" button stays visible indefinitely
5. No text response is ever produced by the model
6. On page reload, the last tool's result shows as "loading" forever (it was never sent back to the server in a continuation request)

### Expected behavior

After all client tools in a round complete, a continuation request should be sent to the server so the model can produce its text response.

## Root Cause Analysis

The issue appears to be in `ChatClient`'s interaction between `drainPostStreamActions()`, `checkForContinuation()`, and the shared `postStreamActions` queue.

### The flow

When multiple tools complete in the same round, each `addToolResult()` call queues a `checkForContinuation()` action onto `postStreamActions`:

```
postStreamActions = [checkForContinuation_A, checkForContinuation_B]
```

After the stream finalizes, `streamResponse()`'s `finally` block calls `drainPostStreamActions()`:

```typescript
// streamResponse() finally block:
this.setIsLoading(false)
await this.drainPostStreamActions()  // processes the queue
```

### The problem: nested drain steals actions

`drainPostStreamActions()` runs `checkForContinuation_A`, which calls `streamResponse()` for the continuation (Round 3). Inside Round 3's `streamResponse()` `finally` block, `drainPostStreamActions()` is called **again** (nested). This **inner drain** finds `checkForContinuation_B` still in the shared `postStreamActions` array and processes it:

```
Outer drain: shift checkForContinuation_A → await it
  checkForContinuation_A: continuationPending = true → await streamResponse() (Round 3)
    Round 3 finally: drainPostStreamActions() (INNER DRAIN)
      Inner drain: shift checkForContinuation_B → await it
        checkForContinuation_B: sees continuationPending = true → RETURNS immediately (no-op)
      Inner drain complete
    Round 3 streamResponse() returns
  checkForContinuation_A: continuationPending = false
Outer drain: postStreamActions is now empty → exits
```

If Round 3 itself involves tool calls (model calls another tool), the continuation from Round 3 is blocked because:
1. `checkForContinuation_B` (stolen by inner drain) sees `continuationPending = true` and returns without doing anything
2. The explicit continuation check at the end of `streamResponse()` also sees `continuationPending = true`
3. After `continuationPending` is finally set to `false`, no one calls `checkForContinuation()` again
4. The chat is stuck with tool results but no model continuation

### Secondary issue: `areAllToolsComplete()` returns `true` for no-tool messages

```typescript
areAllToolsComplete(): boolean {
  // ...
  if (toolParts.length === 0) return true  // ← always true for text-only messages
  // ...
}
```

This means `shouldAutoSend()` returns `true` even after a text-only response, which can cause spurious continuation rounds when combined with the duplicate action problem.

## Minimal reproduction scenario

1. Configure `useChat` with 2+ client tools (using `toolDefinition().client()`)
2. Set up a prompt that causes the LLM to call both tools in the same response
3. Both tools execute and complete successfully
4. Observe: `isLoading` goes `false`, last part is `tool-result`, but no continuation fires

## Suggested fix

A few possible approaches:

### Option A: Deduplicate continuation actions
Instead of queuing a `checkForContinuation` per tool result, use a single flag:

```typescript
async addToolResult(result) {
  this.processor.addToolResult(...)
  
  if (this.isLoading) {
    // Only queue ONE continuation check, not one per tool
    if (!this.postStreamActions.some(a => a === this.checkForContinuation)) {
      this.queuePostStreamAction(() => this.checkForContinuation())
    }
    return
  }
  await this.checkForContinuation()
}
```

### Option B: Don't nest drains
Prevent re-entrancy in `drainPostStreamActions()`:

```typescript
private draining = false

private async drainPostStreamActions(): Promise<void> {
  if (this.draining) return  // skip nested drain
  this.draining = true
  try {
    while (this.postStreamActions.length > 0) {
      const action = this.postStreamActions.shift()!
      await action()
    }
  } finally {
    this.draining = false
  }
}
```

### Option C: Post-continuation recheck
After `continuationPending` is cleared, recheck if continuation is still needed:

```typescript
private async checkForContinuation(): Promise<void> {
  if (this.continuationPending || this.isLoading) return

  if (this.shouldAutoSend()) {
    this.continuationPending = true
    try {
      await this.streamResponse()
    } finally {
      this.continuationPending = false
      // Recheck after clearing the flag
      if (this.shouldAutoSend() && !this.isLoading) {
        await this.checkForContinuation()
      }
    }
  }
}
```

Option B seems cleanest as it prevents the root cause (nested drain stealing actions) without changing the continuation semantics.

willwillems:
## 🎯 Changes

Fixes #302 — when an LLM response triggers multiple client-side tool calls in a single round, the chat permanently stalls after all tools complete because the continuation request never fires.

**Two changes in `ChatClient` (`packages/typescript/ai-client/src/chat-client.ts`):**

1. **Re-entrancy guard on `drainPostStreamActions`**: Adds a `draining` flag that prevents nested calls. When a continuation stream fires inside the drain and its `finally` block tries to drain again, the nested call is skipped. The outer drain's `while` loop naturally picks up any actions queued during the nested stream.

2. **`tool-result` type guard in `checkForContinuation`**: Adds an early return when the last message part is not a `tool-result`. This prevents duplicate queued checks from the same round from firing spurious continuation requests after the model has already produced its text response.

Both changes are needed together — the drain guard fixes the structural cause (nested drain stealing actions), and the lastPart guard fixes the semantic cause (duplicate checks firing after continuation already succeeded).

## ✅ Checklist

## 🚀 Release Impact

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a stall that occurred when multiple client-side tools completed simultaneously in the same processing round.

### DIFF
--- a/.changeset/fix-client-tool-continuation-stall.md
+++ b/.changeset/fix-client-tool-continuation-stall.md
@@ -1,0 +1,7 @@
+---
+'@tanstack/ai-client': patch
+---
+
+fix: prevent client tool continuation stall when multiple tools complete in the same round
+
+When an LLM response triggers multiple client-side tool calls simultaneously, the chat would permanently stall after all tools completed. This was caused by nested `drainPostStreamActions` calls stealing queued continuation checks from the outer drain. Added a re-entrancy guard on `drainPostStreamActions` and a `tool-result` type check in `checkForContinuation` to prevent both the structural and semantic causes of the stall.
--- a/packages/typescript/ai-client/src/chat-client.ts
+++ b/packages/typescript/ai-client/src/chat-client.ts
@@ -40,6 +40,8 @@ export class ChatClient {
   private pendingToolExecutions: Map<string, Promise<void>> = new Map()
   // Flag to deduplicate continuation checks during action draining
   private continuationPending = false
+  // Re-entrancy guard for drainPostStreamActions
+  private draining = false
 
   private callbacksRef: {
     current: {
@@ -619,9 +621,15 @@ export class ChatClient {
    * Drain and execute all queued post-stream actions
    */
   private async drainPostStreamActions(): Promise<void> {
-    while (this.postStreamActions.length > 0) {
-      const action = this.postStreamActions.shift()!
-      await action()
+    if (this.draining) return
+    this.draining = true
+    try {
+      while (this.postStreamActions.length > 0) {
+        const action = this.postStreamActions.shift()!
+        await action()
+      }
+    } finally {
+      this.draining = false
     }
   }
 
@@ -631,6 +639,12 @@ export class ChatClient {
   private async checkForContinuation(): Promise<void> {
     // Prevent duplicate continuation attempts
     if (this.continuationPending || this.isLoading) {
+      return
+    }
+
+    const messages = this.processor.getMessages()
+    const lastPart = messages.at(-1)?.parts.at(-1)
+    if (lastPart?.type !== 'tool-result') {
       return
     }
 


derekalia:
also running into this issue. not sure if this is the fix.

tombeckenham:
Thanks for finding this. We fixed this in #429 . Sorry it's taken so long to get back to you!

PR #395, opened 2026-03-23
fix(ai): produce new object references in tool-call message updaters

braden-w:
I ran into this while building a Svelte 5 chat UI with tool-call approvals. Svelte 5 uses proxies for reactivity, and `$derived(part.state === 'approval-requested')` never fired because the part object identity didn't change.

The root cause is four functions in `message-updaters.ts` that mutate tool-call parts in-place after `[...msg.parts]`. The shallow copy creates a new array but the elements are the same references, so `toolCallPart.state = 'approval-requested'` changes the original object in both arrays. This affects any framework using proxy-based reactivity (Svelte 5, Vue 3).

The other updaters (`updateToolCallPart`, `updateTextPart`, `updateThinkingPart`) already produce new objects via spread. This PR aligns the four remaining functions to the same pattern:

```typescript
const index = parts.indexOf(toolCallPart)
parts[index] = { ...toolCallPart, state: 'approval-requested', approval: { ... } }
```

Functions fixed: `updateToolCallApproval`, `updateToolCallState`, `updateToolCallWithOutput`, `updateToolCallApprovalResponse`. Four immutability tests added.

## Checklist

- [x] Changeset included.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed tool-call message updater behavior to ensure proper state isolation and consistency when processing multiple updates, improving reliability of tool interactions.

### DIFF
--- a/.changeset/immutable-tool-call-updates.md
+++ b/.changeset/immutable-tool-call-updates.md
@@ -1,0 +1,18 @@
+---
+'@tanstack/ai': patch
+---
+
+fix(ai): produce new object references in tool-call message updaters
+
+`updateToolCallApproval`, `updateToolCallState`, `updateToolCallWithOutput`,
+and `updateToolCallApprovalResponse` previously mutated the found tool-call
+part in-place (`toolCallPart.state = ...`) after spreading the parts array.
+The shallow `[...msg.parts]` copy created a new array but preserved the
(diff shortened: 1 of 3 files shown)


tombeckenham (approved):
Thanks for finding this

PR #398, opened 2026-03-24
feat(ai): add `clientOutput` to tool definitions for server-side result filtering

imsherrill:
## Summary

- Add optional `clientOutput` transform to tool definitions that filters tool results before they reach the client
- The full result is always sent to the LLM (for reasoning), while only the filtered result is streamed to the browser
- Designed for tools that return sensitive data (PII, internal scores, credentials) that the model needs but the UI should not display
- Zero breaking changes — omitting `clientOutput` preserves existing behavior exactly

## Motivation

Many AI applications have tools that query databases, user records, or internal systems. These tools often return data the model needs for reasoning (e.g., a user's full profile with SSN, credit score, internal flags) but that should **never** leave the server and reach the client UI.

Today, tool results follow a single path: the same data goes to both the LLM and the client. The only workaround is middleware (`onChunk` or `onAfterToolCall`), which is per-chat-call configuration rather than a property of the tool itself. This means every call site that uses the tool must remember to add the right filtering logic.

`clientOutput` makes the filtering **declarative and co-located with the tool definition**, so it can't be forgotten:

```typescript
const userLookup = toolDefinition({
  name: 'lookup_user',
  description: 'Look up user details',
  outputSchema: z.object({
    id: z.string(),
    name: z.string(),
    ssn: z.string(),
    internalScore: z.number(),
  }),
  clientOutput: (result) => ({
    id: result.id,
    name: result.name,
    // ssn and internalScore never leave the server
  }),
}).server(async (args) => db.users.find(args.query))
```

## Changes

**Core (`@tanstack/ai`):**
- `clientOutput?: (result: any) => any` on `Tool` interface (`types.ts`)
- `clientOutput?: (result: InferSchemaType<TOutput>) => unknown` on `ToolDefinitionConfig` and `ClientTool` (strongly typed from `outputSchema`)
- `buildToolResultChunks()` in `TextEngine` splits results: filtered for `TOOL_CALL_END` chunk, full for `tool` role message
- `ToolCallManager.executeTools()` applies the same split for the standalone code path
- `clientOutput` propagates through `.server()` and `.client()` via config spread (no additional code needed)
- Errors bypass the filter (error results are never transformed)

**Tests:**
- 7 type-level tests (`client-output-types.test.ts`): parameter inference from `outputSchema`, `@ts-expect-error` on bad property access, propagation through `.server()`/`.client()`, `any` fallback without schema, optional by default
- 2 unit tests in `tool-definition.test.ts`: preserves `clientOutput` across definition/server/client, defaults to undefined
- 4 unit tests in `tool-call-manager.test.ts`: ToolCallManager filters `TOOL_CALL_END` but keeps full LLM result, `executeToolCalls` filtering, errors bypass filter, no filter when undefined

## Test plan

- [x] All 48 tests pass (13 new + 35 existing)
- [x] `tsc --noEmit` clean on core package
- [x] `tsc --noEmit` clean on `ts-react-chat` example (after build)
- [x] Full chat + middleware test suites pass (94 tests)
- [x] Manually verified in `ts-react-chat` example: browser receives filtered data, LLM receives full data and reasons over it correctly

## Summary by CodeRabbit

* **New Features**
  * Tools can now define an optional `clientOutput` transformation to filter tool results sent to clients, while preserving the full result for the AI model's consumption.

### DIFF
--- a/packages/typescript/ai/src/activities/chat/index.ts
+++ b/packages/typescript/ai/src/activities/chat/index.ts
@@ -1084,7 +1084,14 @@ class TextEngine<
     const chunks: Array<StreamChunk> = []
 
     for (const result of results) {
-      const content = JSON.stringify(result.result)
+      const fullContent = JSON.stringify(result.result)
+
+      // Apply clientOutput filter if the tool defines one
+      const tool = this.tools.find((t) => t.name === result.toolName)
+      const clientContent =
+        tool?.clientOutput && result.state !== 'output-error'
+          ? JSON.stringify(tool.clientOutput(result.result))
+          : fullContent
 
       chunks.push({
         type: 'TOOL_CALL_END',
@@ -1092,14 +1099,14 @@ class TextEngine<
         model: finishEvent.model,
         toolCallId: result.toolCallId,
         toolName: result.toolName,
-        result: content,
+        result: clientContent,
       })
 
       this.messages = [
         ...this.messages,
         {
           role: 'tool',
-          content,
+          content: fullContent,
           toolCallId: result.toolCallId,
         },
       ]
--- a/packages/typescript/ai/src/activities/chat/tools/tool-calls.ts
+++ b/packages/typescript/ai/src/activities/chat/tools/tool-calls.ts
@@ -162,6 +162,7 @@ export class ToolCallManager {
       const tool = this.tools.find((t) => t.name === toolCall.function.name)
 
       let toolResultContent: string
+      let clientResultContent: string | undefined
       if (tool?.execute) {
         try {
           // Parse arguments (normalize "null" to "{}" for empty tool_use blocks)
@@ -215,6 +216,13 @@ export class ToolCallManager {
 
           toolResultContent =
             typeof result === 'string' ? result : JSON.stringify(result)
+
+          // Apply clientOutput filter if the tool defines one
+          if (tool.clientOutput) {
+            const parsed =
+              typeof result === 'string' ? JSON.parse(result) : result
+            clientResultContent = JSON.stringify(tool.clientOutput(parsed))
+          }
         } catch (error: unknown) {
           // If tool execution fails, add error message
           const message =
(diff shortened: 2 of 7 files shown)


AlemTuzlak:
This can be achieved by middleware I believe

imsherrill:
> This can be achieved by middleware I believe

Agreed, 

This was a while ago but at the time I was thinking we would want it to be coupled on the tool definition in case you want to use the tool in multiple agents?

Philosophical question I guess

imsherrill:
@AlemTuzlak 

I just remembered my thinking here: if the logic lived on the tool, we could reuse it when rehydrating the client UI from persisted state. I still think the library is missing a way to rehydrate UI with the same rules as the streamed chat path, without requiring the chat stream itself (e.g. after a page reload). This was an experiment in that direction, since the rehydration path could reuse the tool definition’s clientOutput function and stay consistent with the streaming path.

tombeckenham:
@imsherrill 

I've had a look at this. It definately got me thinking. Thank you for this.

I get what you're trying to do - a tool is returning sensitive data that you don't want repeated to the client. I can see why you'd want this control on the tool definition itself. The problem is... the LLM could do just about anything with that data. It also could be stored in logs, and so on. It the data really is sensitive, the tool should not be returning the data at all but some sort of reference to it. e.g. if you need a SSN, return some identifier, then when you needed to use that SSN (e.g. send to some other API), then call via a tool that understands your identifier. I don't see that adding a clientOutput is a safe way to achieve what you want.

Having said all that, if you're willing to let the LLM go wild with sensitive data, would the chat stream that is processing that data be the same stream that is going back to the user? Perhaps you want to have something secondary which is observing the task then streaming something else back to the user. e.g. sending messages such as "processing application" etc

There's one interesting thing you might want explore, instead of a clientOutput field, I'm wondering if we could put middleware directly onto tool definitions? We don't support that yet, but if you'd like to work it through, and raise a PR, I'd be happy to look at it.

In the meantime, I'm going to close this. Main has moved a long way since this was raised. I'm sorry it took us so long to look at this.

PR #409, opened 2026-03-30
feat: extract @tanstack/openai-base and @tanstack/ai-utils packages

AlemTuzlak:
## Summary

Extracts shared logic into two new packages to eliminate code duplication across all adapter packages:

- **`@tanstack/ai-utils`** — Zero-dependency, provider-agnostic utilities: `generateId`, `getApiKeyFromEnv`, `transformNullsToUndefined`, `ModelMeta` types + `defineModelMeta`. Used by all 9 adapter packages.
- **`@tanstack/openai-base`** — OpenAI-compatible adapter base classes and utilities: Chat Completions text adapter, Responses API text adapter, image/summarize/transcription/TTS/video adapters, schema converter (`makeStructuredOutputCompatible`), 15 tool converters, shared types. Used by `ai-openai` and `ai-grok` (full adapter inheritance) and `ai-groq`/`ai-openrouter`/`ai-ollama` (selective utility imports).

### Key results
- `ai-openai`: ~1,800 lines removed, extends `OpenAICompatibleResponsesTextAdapter`
- `ai-grok`: ~500 lines removed, extends `OpenAICompatibleChatCompletionsTextAdapter`
- `ai-groq`, `ai-openrouter`, `ai-ollama`: utility dedup via imports
- `ai-anthropic`, `ai-gemini`, `ai-fal`, `ai-elevenlabs`: `generateId`/`getApiKeyFromEnv` dedup via `ai-utils`
- **Zero breaking changes** — all public APIs remain identical
- 103 files changed, +6,469 / -3,154 lines

### New package architecture
```
@tanstack/ai (core)
    ↑
@tanstack/ai-utils (new, zero deps)
    ↑
@tanstack/openai-base (new, deps: ai-utils + openai SDK)
    ↑              ↑
ai-openai      ai-grok          ← extend adapter classes
```

## Test plan

- [x] `pnpm build` — 28/28 projects pass
- [x] `pnpm test:lib` — all unit tests pass across all packages
- [x] `pnpm test:types` — all type checks pass
- [x] `pnpm test:build` — publint validates all packages
- [x] `pnpm format` — all files formatted
- [x] New packages have dedicated test suites (21 tests in ai-utils, 48 in openai-base)
- [x] Existing tests in all migrated packages pass unchanged

## Summary by CodeRabbit

* **New Features**
  * New shared utilities package with API-key lookup, ID generation, null→undefined transform, and model-metadata types/validation
  * New OpenAI‑compatible base package providing unified adapters and tool converters for text, images, summarization, transcription, TTS, and video

* **Refactor**
  * Provider packages delegate shared logic to the new shared utilities and OpenAI‑base while preserving public APIs

* **Tests**
  * Added comprehensive tests covering utilities, schema converters, and OpenAI‑compatible adapters


packages/typescript/ai-utils/src/env.ts
```diff
@@ -0,0 +1,18 @@
+export function getApiKeyFromEnv(envVarName: string): string {
+  const env =
+    typeof globalThis !== 'undefined' && (globalThis as any).window?.env
+      ? (globalThis as any).window.env
+      : typeof process !== 'undefined'
```
tombeckenham, line 5: Install types/node as a dev dependency to avoid the lint

packages/typescript/openai-base/src/adapters/image.ts
```diff
@@ -0,0 +1,150 @@
...
+    this.client = createOpenAICompatibleClient(config)
+  }
+
+  async generateImages(
+    options: ImageGenerationOptions<TProviderOptions>,
+  ): Promise<ImageGenerationResult> {
+    const { model, prompt, numberOfImages, size } = options
+
+    // Validate inputs
+    this.validatePrompt({ prompt, model })
```
tombeckenham, line 58: These were inside the try previously and there was a logger.request

packages/typescript/openai-base/src/adapters/image.ts
```diff
@@ -0,0 +1,150 @@
...
+    // Default: no size validation — subclasses can override
+  }
+
+  protected validateNumberOfImages(
+    _model: string,
+    numberOfImages: number | undefined,
+  ): void {
+    if (numberOfImages === undefined) return
+
+    if (numberOfImages < 1 || numberOfImages > 10) {
```
tombeckenham: Why bother having a max here? Won't the adapters decide?

packages/typescript/ai-openai/src/adapters/summarize.ts
```diff
@@ -32,125 +28,24 @@ export interface OpenAISummarizeProviderOptions {
...
-    }
-
-    return prompt
+    // The text adapter accepts richer provider options than the summarize adapter needs,
+    // but we only pass basic options (model, messages, systemPrompts, etc.) at call time.
+    super(
+      new OpenAITextAdapter(
+        config,
+        model,
+      ) as unknown as ChatStreamCapable<OpenAISummarizeProviderOptions>,
```
tombeckenham: This cast is unnecessary and show be removed

packages/typescript/openai-base/src/adapters/summarize.ts
```diff
@@ -0,0 +1,145 @@
...
+    name: string = 'openai-compatible',
+  ) {
+    super({}, model)
+    this.name = name
+    this.textAdapter = textAdapter
+  }
+
+  async summarize(options: SummarizationOptions): Promise<SummarizationResult> {
+    const systemPrompt = this.buildSummarizationPrompt(options)
+
```
tombeckenham, line 52: missing logger.request.

packages/typescript/openai-base/src/adapters/summarize.ts
```diff
@@ -0,0 +1,145 @@
...
+    let usage = { promptTokens: 0, completionTokens: 0, totalTokens: 0 }
+
+    for await (const chunk of this.textAdapter.chatStream({
+      model: options.model as TModel,
+      messages: [{ role: 'user', content: options.text }],
+      systemPrompts: [systemPrompt],
+      maxTokens: options.maxLength,
+      temperature: 0.3,
+      logger: options.logger,
+    } as TextOptions<TProviderOptions>)) {
```
tombeckenham: use satisfies instead of as

packages/typescript/openai-base/src/adapters/summarize.ts
```diff
@@ -0,0 +1,145 @@
...
+    this.name = name
+    this.textAdapter = textAdapter
+  }
+
+  async summarize(options: SummarizationOptions): Promise<SummarizationResult> {
+    const systemPrompt = this.buildSummarizationPrompt(options)
+
+    let summary = ''
+    const id = generateId(this.name)
+    let model = options.model
```
tombeckenham, line 55: I've never understood why the model is passed both when constructing, AND in options. I thought the whole idea is to use this.model throughout the code so that we could avoid casting.

packages/typescript/openai-base/src/adapters/summarize.ts
```diff
@@ -0,0 +1,145 @@
...
+      }
+      if (chunk.type === 'RUN_FINISHED') {
+        if (chunk.usage) {
+          usage = chunk.usage
+        }
+      }
+      // Surface failures: the underlying chatStream emits RUN_ERROR instead of
+      // throwing, so without this branch summarize() would return an empty
+      // summary and pretend a failed run succeeded.
+      if (chunk.type === 'RUN_ERROR') {
```
tombeckenham: There is no longer a try / catch around summarize. Was this intentional? Logger errors are not captured. It also looks like there is some additional error handling

packages/typescript/openai-base/src/adapters/summarize.ts
```diff
@@ -0,0 +1,145 @@
...
+
+    return { id, model, summary, usage }
+  }
+
+  async *summarizeStream(
+    options: SummarizationOptions,
+  ): AsyncIterable<StreamChunk> {
+    const systemPrompt = this.buildSummarizationPrompt(options)
+
+    yield* this.textAdapter.chatStream({
```
tombeckenham: missing logging

packages/typescript/openai-base/src/adapters/responses-text.ts
```diff
@@ -0,0 +1,1119 @@
...
+    // AG-UI lifecycle tracking
+    const aguiState = {
+      runId: generateId(this.name),
+      messageId: generateId(this.name),
+      timestamp,
+      hasEmittedRunStarted: false,
+    }
+
+    try {
+      const response = await this.client.responses.create(
```
tombeckenham, line 108: logger.request missing. Was this intentional?

packages/typescript/openai-base/src/adapters/responses-text.ts
```diff
@@ -0,0 +1,1119 @@
...
+    const jsonSchema = this.makeStructuredOutputCompatible(
+      outputSchema,
+      outputSchema.required || [],
+    )
+
+    try {
+      // Strip streaming-only fields a subclass override of mapOptionsToRequest
+      // might have returned (parallel to chat-completions's structuredOutput
+      // cleanup) — sending stream_options to a non-streaming call is a 4xx.
+      const {
```
tombeckenham, line 185: missing logger.request

packages/typescript/openai-base/src/adapters/responses-text.ts
```diff
@@ -0,0 +1,1119 @@
...
+        },
+        {
+          headers: chatOptions.request?.headers,
+          signal: chatOptions.request?.signal,
+        },
+      )
+
+      // Extract text content from the response
+      const rawText = this.extractTextFromResponse(
+        response as OpenAI_SDK.Responses.Response,
```
tombeckenham: use satisfies instead of as

packages/typescript/openai-base/src/adapters/responses-text.ts
```diff
@@ -0,0 +1,1119 @@
...
+    let hasEmittedStepStarted = false
+    // Track whether we've emitted a terminal RUN_FINISHED so the
+    // end-of-stream fallback below knows to synthesise one when the upstream
+    // cuts off without a response.completed event.
+    let runFinishedEmitted = false
+
+    try {
+      for await (const chunk of stream) {
+        // Emit RUN_STARTED on first chunk
+        if (!aguiState.hasEmittedRunStarted) {
```
tombeckenham, line 358: You're missing the logger.provider line here. Is this intentional? You haven't mentioned these changes in the PR

packages/typescript/openai-base/src/adapters/responses-text.ts
```diff
@@ -0,0 +1,1119 @@
...
+          yield asChunk({
+            type: 'RUN_STARTED',
+            runId: aguiState.runId,
+            model: model || options.model,
+            timestamp,
+          })
+        }
+
+        const handleContentPart = (contentPart: {
+          type: string
```
tombeckenham, line 369: This type is now a lot wider. It doesn't seem to be a problem, but flagging anyway

packages/typescript/ai-openai/src/adapters/text.ts
```diff
@@ -872,207 +125,35 @@ export class OpenAITextAdapter<
...
       OpenAI_SDK.Responses.ResponseCreateParams,
       'stream'
     > = {
-      model: options.model,
-      temperature: options.temperature,
-      max_output_tokens: options.maxTokens,
-      top_p: options.topP,
-      metadata: options.metadata,
-      instructions: options.systemPrompts?.join('\n'),
       ...modelOptions,
```
tombeckenham, line 132: I'm wondering if this spread could be a utitlity function as its repeated in the OpenAICompatibleResponsesTextAdapter

packages/typescript/openai-base/src/adapters/responses-text.ts
```diff
@@ -0,0 +1,1119 @@
...
+          | { detail?: 'auto' | 'low' | 'high' }
+          | undefined
+        if (part.source.type === 'url') {
+          return {
+            type: 'input_image',
+            image_url: part.source.value,
+            detail: imageMetadata?.detail || 'auto',
+          }
+        }
+        // For base64 data, construct a data URI using the mimeType from
```
tombeckenham, line 1072: I don't love the base64 data uri thing... I've got another pr that prevents that. It's ok to keep as is in this pr as it mirrors what was there

packages/typescript/openai-base/src/adapters/transcription.ts
```diff
@@ -0,0 +1,216 @@
...
+  }
+
+  async transcribe(
+    options: TranscriptionOptions<TProviderOptions>,
+  ): Promise<TranscriptionResult> {
+    const { model, audio, language, prompt, responseFormat, modelOptions } =
+      options
+
+    // Convert audio input to File object
+    const file = this.prepareAudioFile(audio)
```
tombeckenham, line 50: Missing logger.request

packages/typescript/openai-base/src/adapters/transcription.ts
```diff
@@ -0,0 +1,216 @@
...
+      )
+    }
+  }
+
+  /**
+   * Decodes a base64 string to an ArrayBuffer.
+   * Uses `atob` when available, falling back to `Buffer.from` in Node.js.
+   */
+  private decodeBase64(base64: string): ArrayBuffer {
+    if (typeof atob === 'function') {
```
tombeckenham: This function should definately move to ai-utils.

I'd also suggest including the new Uint8Array.fromBase64 function as it is a lot more performant.

```suggestion
    if (typeof (Uint8Array as any).fromBase64 === 'function') {
      return (Uint8Array as any).fromBase64(base64).buffer
    }
    if (typeof atob === 'function') {
```

As a side note, I've run into a tonne of issues on workers moving audio, image, and video files between formats - especially to / from Buffers. They run out of memory very easily if these functions aren't bounded. It's usually a good idea to read the first few bytes to get an idea of the size but kinda complicated when you don't know the format.

packages/typescript/openai-base/src/adapters/tts.ts
```diff
@@ -0,0 +1,137 @@
...
+  if (typeof Buffer !== 'undefined' && typeof Buffer.from === 'function') {
+    return Buffer.from(buffer).toString('base64')
+  }
+  const view = new Uint8Array(buffer)
+  let binary = ''
+  for (let i = 0; i < view.byteLength; i += 1) {
+    binary += String.fromCharCode(view[i]!)
+  }
+  return btoa(binary)
+}
```
tombeckenham: Like my comment in transcription, this should be moved to utils. There's also a new function for toBase64 which should be used if available. Its faster and more memory efficient. If you don't like the as any thing then declare toBase64 as an optional function in a global module 

```suggestion
function arrayBufferToBase64(buffer: ArrayBuffer): string {
  const bytes = new Uint8Array(buffer)

  if (typeof (bytes as any).toBase64 === 'function') {
    return (bytes as any).toBase64()
  }

  if (typeof Buffer !== 'undefined' && typeof Buffer.from === 'function') {
    return Buffer.from(buffer).toString('base64')
  }

  if (typeof btoa === 'function') {
    let binary = ''
    const chunkSize = 0x8000 // 32KB — safe for String.fromCharCode.apply
    for (let i = 0; i < bytes.length; i += chunkSize) {
      const chunk = bytes.subarray(i, i + chunkSize)
      binary += String.fromCharCode.apply(null, chunk as unknown as number[])
    }
    return btoa(binary)
  }

  throw new Error('No base64 encoder available in this environment.')
}

```

packages/typescript/openai-base/src/adapters/video.ts
```diff
@@ -0,0 +1,329 @@
...
+    try {
+      const client = this.client as any
+      const response = await client.videos.create(request)
+
+      return {
+        jobId: response.id,
+        model,
+      }
+    } catch (error: any) {
+      if (error?.message?.includes('videos') || error?.code === 'invalid_api') {
```
tombeckenham, line 121: Missing the logger here

packages/typescript/openai-base/src/adapters/video.ts
```diff
@@ -0,0 +1,329 @@
...
+   *
+   * @experimental Video generation is an experimental feature and may change.
+   */
+  async createVideoJob(
+    options: VideoGenerationOptions<TProviderOptions>,
+  ): Promise<VideoJobResult> {
+    const { model, size, duration, modelOptions } = options
+
+    // Validate inputs
+    this.validateVideoSize(model, size)
```
tombeckenham, line 93: Missing logger.request

packages/typescript/openai-base/src/adapters/video.ts
```diff
@@ -0,0 +1,329 @@
...
+    const seconds = duration ?? (modelOptions as any)?.seconds
+    this.validateVideoSeconds(model, seconds)
```
tombeckenham, line 95: We really need to change the duration validation. Many models support strings like "6s" instead of the number 6. Some take the string "auto". The duration thing is one of the hardest parts of doing video... it's as hard as the sizes in images. I'll add an issue for it. I think duration needs to be a template type, but we might also want to think about creating a snapDuration method. I'll think about it later

packages/typescript/openai-base/src/adapters/video.ts
```diff
@@ -0,0 +1,329 @@
...
+  }
+
+  /**
+   * Get the URL to download/view the generated video.
+   *
+   * @experimental Video generation is an experimental feature and may change.
+   */
+  async getVideoUrl(jobId: string): Promise<VideoUrlResult> {
+    try {
+      const client = this.client as any
```
tombeckenham: Why do we cast to any here?

packages/typescript/openai-base/src/adapters/video.ts
```diff
@@ -0,0 +1,329 @@
...
+          expiresAt: videoInfo.expires_at
+            ? new Date(videoInfo.expires_at)
+            : undefined,
+        }
+      }
+
+      // SDK download fall-through: try the various possible method names in
+      // decreasing order of modernity.
+      if (typeof client.videos?.downloadContent === 'function') {
+        const contentResponse = await client.videos.downloadContent(jobId)
```
tombeckenham: This will almost certainly blow up workers. It's pretty common for videos to be bigger than 100mb.

packages/typescript/openai-base/src/adapters/video.ts
```diff
@@ -0,0 +1,329 @@
...
+          url: `data:${mimeType};base64,${base64}`,
+          expiresAt: undefined,
+        }
+      }
+
+      // The remaining SDK fall-throughs all return a binary payload
+      // (Blob/Response/ArrayBuffer-shaped), NOT an `{ url, expires_at }`
+      // object the way the bottom return assumed. Convert to a data URL
+      // here so the caller actually receives a usable URL.
+      let response: any
```
tombeckenham, line 232: For this PR... ok, it mirrors what we had before but this would definately blow up most servers. Downloading the video and base64'ing it will cause most backends to run out of memory. I'm wondering if we support these older methods at all. At the very least we should add a warning with logger and look at whether buffers are over e.g. 10mb. It's quite hard to figure out where memory issues come from as they don't necessarily occur at the point this blob would be created. I'm wondering if we might want to add a media uploader function to the the constructor in a future PR

packages/typescript/openai-base/src/tools/code-interpreter-tool.ts
```diff
@@ -0,0 +1,39 @@
...
+/**
+ * Creates a standard Tool from CodeInterpreterTool parameters.
+ *
+ * Base (non-branded) factory. Providers that need branded return types should
+ * re-wrap this in their own package.
+ */
+export function codeInterpreterTool(
+  container: CodeInterpreterToolConfig,
+): Tool {
+  return {
```
tombeckenham, line 31: None of these tools are called from the open ai adapter. Shouldn't they call these then cast using ProviderTool?

packages/typescript/ai-grok/src/adapters/image.ts
```diff
@@ -37,7 +31,7 @@ export interface GrokImageConfig extends GrokClientConfig {}
  */
 export class GrokImageAdapter<
   TModel extends GrokImageModel,
```
tombeckenham, line 33: This image model is obsolete btw. I'll raise another pr

packages/typescript/ai-grok/src/utils/client.ts
```diff
@@ -1,46 +1,34 @@
...
  */
-export function generateId(prefix: string): string {
-  return `${prefix}-${Date.now()}-${Math.random().toString(36).substring(7)}`
+export function toCompatibleConfig(
+  config: GrokClientConfig,
+): OpenAICompatibleClientConfig {
+  return {
+    ...config,
+    baseURL: config.baseURL || 'https://api.x.ai/v1',
+  } as unknown as OpenAICompatibleClientConfig
```
tombeckenham: We're converting to as uknown for all configs. It makes OpenAICompatibleClientConfig pointless.

packages/typescript/ai-grok/src/utils/client.ts
```diff
@@ -1,46 +1,34 @@
-import OpenAI_SDK from 'openai'
+import { getApiKeyFromEnv } from '@tanstack/ai-utils'
+import type { OpenAICompatibleClientConfig } from '@tanstack/openai-base'
 import type { ClientOptions } from 'openai'
 
 export interface GrokClientConfig extends ClientOptions {
```
tombeckenham: This should extend OpenAICompatibleClientConfig

packages/typescript/ai-openai/src/utils/client.ts
```diff
@@ -1,42 +1,27 @@
-import OpenAI_SDK from 'openai'
+import { getApiKeyFromEnv } from '@tanstack/ai-utils'
+import type { OpenAICompatibleClientConfig } from '@tanstack/openai-base'
 import type { ClientOptions } from 'openai'
 
 export interface OpenAIClientConfig extends ClientOptions {
```
tombeckenham: This should extend OpenAICompatibleClientConfig

packages/typescript/ai-openai/src/utils/client.ts
```diff
@@ -1,42 +1,27 @@
...
+ * This bridges the type gap between the local config type (which extends
+ * the local copy of ClientOptions) and the base package's config type
+ * (which extends its own copy of ClientOptions).
  */
-export function generateId(prefix: string): string {
-  return `${prefix}-${Date.now()}-${Math.random().toString(36).substring(7)}`
+export function toCompatibleConfig(
+  config: OpenAIClientConfig,
+): OpenAICompatibleClientConfig {
+  return config as unknown as OpenAICompatibleClientConfig
```
tombeckenham: Should not be casting at all here

packages/typescript/openai-base/src/adapters/chat-completions-text.ts
```diff
@@ -0,0 +1,808 @@
...
+                input: parsedInput,
+              })
+              emittedAnyToolCallEnd = true
+            }
+            // Clear tool-call state after emission so a subsequent
+            // `finish_reason: 'stop'` chunk (or the post-loop synthetic
+            // block) doesn't see lingering entries and misreport the finish.
+            toolCallsInProgress.clear()
+          }
+          void emittedAnyToolCallEnd
```
tombeckenham: This emittedAnyToolCallEnd variable seems to have no use at all

packages/typescript/openai-base/src/adapters/chat-completions-text.ts
```diff
@@ -0,0 +1,808 @@
...
+        // run terminated — surface it instead. Use `tool_calls` only when
+        // the upstream actually said so OR when we just closed pending tool
+        // calls (truncated stream); a clean `stop` with no started tool
+        // calls must NOT be remapped to `tool_calls`.
+        const finishReason: string =
+          pendingFinishReason === 'tool_calls' || pendingToolCount > 0
+            ? 'tool_calls'
+            : (pendingFinishReason ?? 'stop')
+
+        yield asChunk({
```
tombeckenham, line 543: Claude found this - tool_calls finishReason can fire with no TOOL_CALL_END events
  — chat-completions-text.ts:399-456 + 529-532. If every toolCallsInProgress entry has started === false, no end events emit but RUN_FINISHED { finishReason: 'tool_calls' } still fires.

packages/typescript/openai-base/src/adapters/responses-text.ts
```diff
@@ -0,0 +1,1119 @@
...
+        }
+
+        // Handle function call arguments delta (streaming). Drop the
+        // previously-emitted `args` field — it had inverted polarity
+        // (populated only when metadata was MISSING, i.e. when the
+        // matching TOOL_CALL_START hadn't fired) and the chat-completions
+        // adapter never emitted it, so it leaked partial deltas as
+        // pseudo-args only on the orphan path. Consumers should accumulate
+        // `delta` themselves.
+        if (
```
tombeckenham, line 722: Claude also found this - function_call_arguments.delta emits TOOL_CALL_ARGS without
  checking metadata?.started — responses-text.ts:706-717. The .done
   handler at line 728 correctly guards; .delta does not.

packages/typescript/openai-base/src/types/message-metadata.ts
```diff
@@ -0,0 +1,19 @@
...
+
+export interface OpenAICompatibleAudioMetadata {
+  format?: 'mp3' | 'wav' | 'flac' | 'ogg' | 'webm' | 'aac'
+}
+
+export interface OpenAICompatibleVideoMetadata {}
+export interface OpenAICompatibleDocumentMetadata {}
+export interface OpenAICompatibleTextMetadata {}
+
+export interface OpenAICompatibleMessageMetadataByModality {
```
tombeckenham: What consumes this? Nothing imports this file

packages/typescript/openai-base/src/types/provider-options.ts
```diff
@@ -0,0 +1,41 @@
+export interface OpenAICompatibleBaseOptions {
```
tombeckenham: What consumes this? Nothing imports this file

packages/typescript/ai-utils/tests/model-meta.test.ts
```diff
@@ -0,0 +1,106 @@
+import { describe, it, expect } from 'vitest'
+import { defineModelMeta } from '../src/model-meta/define'
+import type { ModelMeta, Modality } from '../src/model-meta/types'
+
+describe('defineModelMeta', () => {
```
tombeckenham: What is the point of these defineModelMeta tests? You're creating a structure in the test, then testing that strucutre. It doesn't do anything... no where else is defineModelMeta used

packages/typescript/openai-base/src/adapters/responses-text.ts
```diff
@@ -0,0 +1,1119 @@
...
+   * - response.completed
+   * - error
+   */
+  protected async *processStreamChunks(
+    stream: AsyncIterable<OpenAI_SDK.Responses.ResponseStreamEvent>,
+    toolCallMetadata: Map<
+      string,
+      { index: number; name: string; started: boolean }
+    >,
+    options: TextOptions,
```
tombeckenham: This should be TextOptions<TProviderOptions>

```suggestion
    options: TextOptions<TProviderOptions>,
```

packages/typescript/openai-base/src/adapters/responses-text.ts
```diff
@@ -0,0 +1,1119 @@
...
+      })
+    }
+  }
+
+  /**
+   * Maps common TextOptions to Responses API request format.
+   * Override this in subclasses to add provider-specific options.
+   */
+  protected mapOptionsToRequest(
+    options: TextOptions,
```
tombeckenham: Should be TextOptions<TProviderOptions>


```suggestion
    options: TextOptions<TProviderOptions>,
```

packages/typescript/ai-openai/src/adapters/text.ts
```diff
@@ -116,738 +88,19 @@ export class OpenAITextAdapter<
...
-  }
-
-  private async *processOpenAIStreamChunks(
-    stream: AsyncIterable<OpenAI_SDK.Responses.ResponseStreamEvent>,
-    toolCallMetadata: Map<
-      string,
-      { index: number; name: string; started: boolean }
-    >,
+  protected override mapOptionsToRequest(
     options: TextOptions,
```
tombeckenham: Should be TextOptions<InternalTextProviderOptions>,


```suggestion
    options: TextOptions<InternalTextProviderOptions>,
```

packages/typescript/ai-openai/src/adapters/text.ts
```diff
@@ -116,738 +88,19 @@ export class OpenAITextAdapter<
...
-    }
-  }
-
-  /**
-   * Maps common options to OpenAI-specific format
-   * Handles translation of normalized options to OpenAI's API format
-   */
-  private mapTextOptionsToOpenAI(options: TextOptions) {
+  ): Omit<OpenAI_SDK.Responses.ResponseCreateParams, 'stream'> {
     const modelOptions = options.modelOptions as
```
tombeckenham, line 107: once you fix the options type you'll no longer need to cast

packages/typescript/openai-base/tests/chat-completions-text.test.ts
```diff
@@ -0,0 +1,919 @@
+import { describe, it, expect, vi, afterEach, beforeEach } from 'vitest'
```
tombeckenham, line 1: There's no tests for any of the media adapters in base

tombeckenham (changes requested):
This is a good move moving common code to base. In future, it would be useful if you'd re-organised in one PR, then made changes in another PR. It took quite a long time to review as I was comparing the base to the original and there were changes as well as reorganisation. It was a good exercise to go through though as it gave me a chance to properly go through all of the code.

Please take a look through the suggestions. There are some changes you should make

tombeckenham:
@alem — flagging a question about migration scope, specifically around openrouter.

On this branch, only `ai-openai` and `ai-grok` extend the new bases. `ai-groq`, `ai-openrouter`, and `ai-ollama` still extend `BaseTextAdapter` directly — together that's ~2k lines of duplicated text-adapter code the bases were designed to absorb.

For groq the migration looks mechanical (Chat Completions wire format, OpenAI SDK works fine pointed at Groq's endpoint). **openrouter is the interesting one** — wanted to flag the design question before assuming it slots in.

### Wire format: identical. SDK shape: not.

OpenRouter's HTTP API is OpenAI-compatible Chat Completions, so on the wire the base's SSE accumulator, partial-JSON, tool-call buffers, and RUN_ERROR taxonomy all apply unchanged. But the **TypeScript SDK** (`@openrouter/sdk`, Speakeasy-generated) has a different shape from the OpenAI SDK the base assumes:

| | OpenAI SDK (base) | `@openrouter/sdk` |
|---|---|---|
| Call shape | `client.chat.completions.create(params)` | `client.chat.send({ chatRequest: params })` |
| Field naming | snake_case (`max_completion_tokens`, `top_p`) | camelCase (`maxCompletionTokens`, `topP`) |
| Abort error | `APIUserAbortError` | `RequestAbortedError` from `@openrouter/sdk/models/errors` |

Grok gets away with a 60-line subclass because xAI's "OpenAI-compatible" endpoint accepts the OpenAI SDK verbatim — `toCompatibleConfig` just rewrites the `baseURL`. OpenRouter's SDK is not a drop-in.

### OpenRouter-specific surface

These are real user-facing features, not incidental:

1. **Provider routing** — `provider: { order, fallbacks, allow_fallbacks, sort, ignore }`, `models[]` fallback chain, `route`, `transforms: ['middle-out']`. OpenRouter's distinguishing feature.
2. **Response metadata** — which provider served the request, which underlying model was actually used, generation ID for cost lookup.
3. **App attribution headers** — `HTTP-Referer`, `X-Title`.

### Considered and ruled out: dropping `@openrouter/sdk`

The cleanest-looking option would be replacing `@openrouter/sdk` with the OpenAI SDK pointed at `openrouter.ai/api/v1` (i.e., the grok approach). That deletes ~700 lines and the base works unchanged.

I ruled it out: it forces provider routing through untyped `extra_body` pass-through and loses the typed response metadata. That's a type-safety regression on OpenRouter's headline feature, and the SDK is purpose-built around exactly that surface — keeping it is the point.

### Proposal: parameterize the base with two protected hooks

Add to `OpenAICompatibleChatCompletionsTextAdapter`:

```ts
protected callChatCompletion(params, signal): Promise<ChatCompletion>
protected callChatCompletionStream(params, signal): AsyncIterable<ChatCompletionChunk>
```

Default impl calls the OpenAI SDK (covers grok and any future "OpenAI-compatible-endpoint" providers like deepseek/together/fireworks). openrouter overrides to call `@openrouter/sdk`.

Everything else — tool/message converters, SSE accumulator, partial-JSON, RUN_ERROR taxonomy, structured output, lifecycle — stays in the base. openrouter subclass collapses to ~150 lines (SDK shim + provider-routing typing + response metadata extraction) vs. 807 today.

### Why this matters for #409 specifically

PR #527 (streaming structured output) currently duplicates ~290 lines of `structuredOutput` + `chatStreamStructured` across openai/grok/groq. The plan there is to lift that into the bases so each wire format has one implementation. **That lift only covers openrouter if openrouter is on the Chat Completions base.** If #409 ships with openrouter unmigrated, #527 has to either keep duplicating into openrouter or fold the openrouter migration into a feature PR.

### Questions

1. Was leaving groq/openrouter/ollama on `BaseTextAdapter` deliberate (scoping #409 to "introduce bases, migrate later") or time-boxed?
2. Open to expanding #409 to migrate groq + openrouter, or would you prefer that as a follow-up?
3. Any concerns with the two-hook approach to the base, or alternative shapes you'd prefer?

Happy to take a swing at the hook signatures on a side branch if it'd help the conversation.

packages/typescript/openai-base/src/adapters/summarize.ts
tombeckenham: Won't be addressed in this PR

tombeckenham:
Moved the migration-scope discussion to its own issue so it doesn't get lost regardless of how this PR lands: #543

tombeckenham (approved):
Good work on this. Approving. There is still the open questiion about the openrouter base which is important before we rebase the streaming output pr.

PR #459, opened 2026-04-16
fix(ai-gemini): read/write thoughtSignature at Part level for Gemini 3.x

### LINKED ISSUE #403: `providerMetadata` lost in client-side UIMessage pipeline — breaks Gemini 3 thoughtSignature on multi-turn tool calls
### Describe the bug

PR #401 added `providerMetadata` to `ToolCall` and `ToolCallStartEvent` types, and correctly wired it through the **server-side** `ToolCallManager` (used by `chat()`) and the **Gemini adapter** (`processStreamChunks` → `formatMessages`).

However, the **client-side** pipeline (`StreamProcessor` → UIMessage → ModelMessage) does not carry `providerMetadata` at all. This means:

1. **First request works:** `chat()` on the server uses `ToolCallManager`, which preserves `providerMetadata` on `ToolCall` objects. The Gemini adapter reads `toolCall.providerMetadata?.thoughtSignature` in `formatMessages()` — all good.

2. **Second request fails:** The client sends the conversation history back to the server. The `StreamProcessor` processed the first response's `TOOL_CALL_START` events, but dropped `providerMetadata` when creating `ToolCallPart` UIMessage parts. When converting back to `ModelMessage.toolCalls` for the second request, `providerMetadata` is missing. The Gemini adapter's `formatMessages()` finds no `thoughtSignature`, and Gemini returns HTTP 400:

```
Function call is missing a thought_signature in functionCall parts.
This is required for tools to work correctly, and missing thought_signature
may lead to degraded model performance.
```

### Root cause — 5 locations in `@tanstack/ai` core

| # | File | Function/Type | Issue |
|---|------|---------------|-------|
| 1 | `types.ts` | `ToolCallPart` | Missing `providerMetadata` field |
| 2 | `stream/types.ts` | `InternalToolCallState` | Missing `providerMetadata` field |
| 3 | `stream/message-updaters.ts` | `updateToolCallPart()` | Doesn't accept or store `providerMetadata` |
| 4 | `stream/processor.ts` | `handleToolCallStartEvent()` | Doesn't pass `chunk.providerMetadata` to `updateToolCallPart()` or `InternalToolCallState` |
| 5 | `messages.ts` | `buildAssistantMessages()` | Doesn't include `providerMetadata` on `toolCalls` when converting `ToolCallPart` → `ModelMessage.toolCalls` |

Additionally, `modelMessageToUIMessage()` and `getCompletedToolCalls()` also drop `providerMetadata`.

### Why this wasn't caught

The test in PR #401 (`preserves thoughtSignature in functionCall parts when sending history back to Gemini`) only tests the **server-side** `chat()` flow where `ToolCallManager` correctly handles `providerMetadata`. The client-side `StreamProcessor` → `UIMessage` → `ModelMessage` path is not tested.

The server-side `chat()` loop works fine for single-session server-only scenarios. The bug manifests in client-server architectures where:
- The client uses `ChatClient` / `StreamProcessor` to process SSE events into UIMessages
- On the next user message, UIMessages are converted to ModelMessages via `uiMessageToModelMessages()` and sent to the server
- The server passes these messages to `chat()` → adapter → Gemini API

### Steps to reproduce

1. Use `@tanstack/ai-gemini` with `gemini-3-flash-preview` and thinking enabled
2. Send a message that triggers a tool call
3. First response succeeds (tool call + response)
4. Send a follow-up message
5. HTTP 400: "Function call is missing a thought_signature in functionCall parts"

### Environment

- `@tanstack/ai@0.9.1`
- `@tanstack/ai-gemini@0.8.4`
- `@tanstack/ai-client@0.7.4`
- Model: `gemini-3-flash-preview` with `thinkingConfig: { includeThoughts: true }`
- Architecture: Client (`ChatClient`/`StreamProcessor`) ↔ Server (`chat()` + Gemini adapter)

### Related

- Fixes the remaining gap from #216
- PR #401 fixed the server-side adapter but not the client-side pipeline

pemontto:
## Summary

Gemini 3.x models emit `thoughtSignature` as a **Part-level sibling** of `functionCall` (per the [`@google/genai` `Part` type definition](https://github.com/googleapis/js-genai/blob/main/src/types.ts)), not nested inside `functionCall`. The `FunctionCall` interface has no `thoughtSignature` property at all.

The adapter was:
- **Reading** from `functionCall.thoughtSignature` (wrong location, doesn't exist in SDK types)
- **Writing** it back nested inside `functionCall` (wrong location, API ignores it there)

This causes Gemini 3.x to reject subsequent tool-call turns with:
```
400 INVALID_ARGUMENT: "Function call is missing a thought_signature"
```

### The `@google/genai` Part type (for reference)
```typescript
export declare interface Part {
    functionCall?: FunctionCall;
    thoughtSignature?: string;  // <-- Part-level sibling
    // ...
}

export declare interface FunctionCall {
    id?: string;
    args?: Record<string, unknown>;
    name?: string;
    // no thoughtSignature here
}
```

## Changes

- **Read side** (`processStreamChunks`): reads `part.thoughtSignature` first, falls back to `functionCall.thoughtSignature` for Gemini 2.x compatibility
- **Write side** (`formatMessages`): emits `thoughtSignature` as a Part-level sibling of `functionCall` instead of nesting it inside

## Test plan

- [x] Existing tests pass (66/66)
- [x] Added test: reads Part-level `thoughtSignature` from Gemini 3.x streaming response and round-trips it at the Part level
- [x] Added test: falls back to `functionCall.thoughtSignature` for Gemini 2.x wire format
- [x] Verified fix against live `gemini-3.1-pro-preview` and `gemini-3.1-flash-lite-preview` sessions (multi-turn tool calling with thinking enabled)

Closes #403
Related: #218, #401, #404

## Summary by CodeRabbit

* **Bug Fixes**
  * Read and emit tool-call thought signatures at the correct Part level and preserve signatures from streamed chunks to avoid validation errors.

* **Tests**
  * Updated and added tests to verify Part-level signature handling and legacy nested-signature fallback and non-propagation cases.

* **Chores / Breaking Changes**
  * Tool-call metadata renamed and typed: providerMetadata → metadata (now generic/typed) and is preserved through the message pipeline.

### DIFF
--- a/packages/typescript/ai-gemini/src/adapters/text.ts
+++ b/packages/typescript/ai-gemini/src/adapters/text.ts
@@ -33,7 +33,10 @@ import type {
   TextOptions,
 } from '@tanstack/ai'
 import type { ExternalTextProviderOptions } from '../text/text-provider-options'
-import type { GeminiMessageMetadataByModality } from '../message-types'
+import type {
+  GeminiMessageMetadataByModality,
+  GeminiToolCallMetadata,
+} from '../message-types'
 import type { GeminiClientConfig } from '../utils'
 
 /** Cast an event object to StreamChunk. Adapters construct events with string
@@ -104,7 +107,8 @@ export class GeminiTextAdapter<
   TProviderOptions,
   TInputModalities,
   GeminiMessageMetadataByModality,
-  TToolCapabilities
+  TToolCapabilities,
+  GeminiToolCallMetadata
 > {
   readonly kind = 'text' as const
   readonly name = 'gemini' as const
@@ -435,9 +439,9 @@ export class GeminiTextAdapter<
                 timestamp,
                 index: toolCallData.index,
                 ...(toolCallData.thoughtSignature && {
-                  providerMetadata: {
+                  metadata: {
                     thoughtSignature: toolCallData.thoughtSignature,
-                  },
+                  } satisfies GeminiToolCallMetadata,
                 }),
               })
             }
@@ -714,8 +718,9 @@ export class GeminiTextAdapter<
             >
           }
 
-          const thoughtSignature = toolCall.providerMetadata
-            ?.thoughtSignature as string | undefined
+          const thoughtSignature = (
+            toolCall.metadata as GeminiToolCallMetadata | undefined
+          )?.thoughtSignature
           // Gemini requires thoughtSignature at the Part level (sibling of
           // functionCall), not nested inside functionCall. Nesting it causes
           // the API to reject the next turn with
--- a/packages/typescript/ai-gemini/src/adapters/text.ts
+++ b/packages/typescript/ai-gemini/src/adapters/text.ts
@@ -289,6 +289,15 @@ export class GeminiTextAdapter<
               `${functionCall.name}_${Date.now()}_${nextToolIndex}`
             const functionArgs = functionCall.args || {}
 
+            // Gemini 3.x emits thoughtSignature as a Part-level sibling of
+            // functionCall (see @google/genai Part type), not nested inside
+            // functionCall. Read from the Part first, fall back to
+            // functionCall for older Gemini 2.x models.
+            const partThoughtSignature =
+              (part as any).thoughtSignature ||
+              (functionCall as any).thoughtSignature ||
+              undefined
+
             let toolCallData = toolCallMap.get(toolCallId)
             if (!toolCallData) {
               toolCallData = {
@@ -299,11 +308,13 @@ export class GeminiTextAdapter<
                     : JSON.stringify(functionArgs),
                 index: nextToolIndex++,
                 started: false,
-                thoughtSignature:
-                  (functionCall as any).thoughtSignature || undefined,
+                thoughtSignature: partThoughtSignature,
               }
               toolCallMap.set(toolCallId, toolCallData)
             } else {
+              if (!toolCallData.thoughtSignature && partThoughtSignature) {
+                toolCallData.thoughtSignature = partThoughtSignature
+              }
               try {
                 const existingArgs = JSON.parse(toolCallData.args)
                 const newArgs =
@@ -585,14 +596,18 @@ export class GeminiTextAdapter<
 
           const thoughtSignature = toolCall.providerMetadata
             ?.thoughtSignature as string | undefined
+          // Gemini 3.x requires thoughtSignature at the Part level (sibling
+          // of functionCall), not nested inside functionCall. Nesting it
+          // causes the API to reject the next turn with
+          // "Function call is missing a thought_signature".
           parts.push({
             functionCall: {
               id: toolCall.id,
               name: toolCall.function.name,
               args: parsedArgs,
-              ...(thoughtSignature && { thoughtSignature }),
-            } as any,
-          })
+            },
+            ...(thoughtSignature && { thoughtSignature }),
+          } as Part)
         }
       }
 
--- a/packages/typescript/ai-gemini/src/adapters/text.ts
+++ b/packages/typescript/ai-gemini/src/adapters/text.ts
@@ -385,14 +385,10 @@ export class GeminiTextAdapter<
               `${functionCall.name}_${Date.now()}_${nextToolIndex}`
             const functionArgs = functionCall.args || {}
 
-            // Gemini 3.x emits thoughtSignature as a Part-level sibling of
-            // functionCall (see @google/genai Part type), not nested inside
-            // functionCall. Read from the Part first, fall back to
-            // functionCall for older Gemini 2.x models.
-            const partThoughtSignature =
-              (part as any).thoughtSignature ||
-              (functionCall as any).thoughtSignature ||
-              undefined
+            // Gemini emits thoughtSignature as a Part-level sibling of
+            // functionCall (per @google/genai Part type), not nested inside
+            // functionCall itself.
+            const partThoughtSignature = part.thoughtSignature || undefined
 
             let toolCallData = toolCallMap.get(toolCallId)
             if (!toolCallData) {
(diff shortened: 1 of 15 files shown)


pemontto:
Re: the pre-merge check about #403 scope.

This PR fixes the **server-side adapter** (the `GeminiTextAdapter` read/write paths), which is the direct cause of the `400 INVALID_ARGUMENT: "Function call is missing a thought_signature"` errors on Gemini 3.x models.

The client-side `providerMetadata` pipeline issue (threading through `ToolCallPart`, `InternalToolCallState`, `StreamProcessor`, etc.) is a separate concern tracked in #404. That PR addresses whether `providerMetadata` survives the client-side UIMessage round-trip. Both fixes are needed for a complete solution, but they're independent: this adapter fix stops the API rejections, and #404 ensures the metadata persists through the client layer.

mattsoltani:
Can confirm this also fixes the issue we were experiencing missing thought signature

packages/typescript/ai-gemini/src/adapters/text.ts
```diff
@@ -351,6 +351,15 @@ export class GeminiTextAdapter<
               `${functionCall.name}_${Date.now()}_${nextToolIndex}`
             const functionArgs = functionCall.args || {}
 
+            // Gemini 3.x emits thoughtSignature as a Part-level sibling of
+            // functionCall (see @google/genai Part type), not nested inside
+            // functionCall. Read from the Part first, fall back to
+            // functionCall for older Gemini 2.x models.
+            const partThoughtSignature =
+              (part as any).thoughtSignature ||
+              (functionCall as any).thoughtSignature ||
```
AlemTuzlak: why are these cast to any? can we update the sdk or whatever else is required to make this typesafe (I'm aware it was cast before as well below)

pemontto:
Good call, addressed in ab05e4ce. The `@google/genai` SDK actually types `thoughtSignature` on `Part` properly, so both `as any` casts in this area were unnecessary:

- **Read side:** removed `(part as any).thoughtSignature` → `part.thoughtSignature` (typed). Also removed the `(functionCall as any).thoughtSignature` fallback entirely. The SDK has never had `thoughtSignature` on `FunctionCall`, so that fallback was based on a misconception from the original PR #401 and was never a real code path.
- **Write side:** replaced the `} as Part)` spread-literal cast with an explicit typed `const part: Part = {...}` construction and conditional assignment.

Only remaining `as any` in the Gemini adapter is the pre-existing `functionResponse` cast (line ~714), which is unrelated to this fix.

All 66 tests still pass, typecheck clean.

mattsoltani:
@pemontto After some further testing, it seems there are a few more places in the tool call life cycle where the provider meta data should be added to avoid more errors. See the following file locations:
1. [ai/packages/typescript/ai/src/activities/chat/messages.ts#L206](https://github.com/TanStack/ai/blob/main/packages/typescript/ai/src/activities/chat/messages.ts#L206)
2. [ai/packages/typescript/ai/src/activities/chat/messages.ts#L342](https://github.com/TanStack/ai/blob/main/packages/typescript/ai/src/activities/chat/messages.ts#L342)
3. [ai/packages/typescript/ai/src/activities/chat/stream/message-updaters.ts#L78](https://github.com/TanStack/ai/blob/main/packages/typescript/ai/src/activities/chat/stream/message-updaters.ts#L78)
4. [ai/packages/typescript/ai/src/activities/chat/stream/processor.ts#L922](https://github.com/TanStack/ai/blob/main/packages/typescript/ai/src/activities/chat/stream/processor.ts#L922)

pemontto:
Good catch @mattsoltani, those locations are exactly what [#404](https://github.com/TanStack/ai/pull/404) (by @houmark, opened 2026-03-28) addresses. It threads `providerMetadata` through `ToolCallPart`, `InternalToolCallState`, `updateToolCallPart`, `handleToolCallStart`, `buildAssistantMessages`, and `modelMessageToUIMessage`, the full client-side pipeline.

That PR has been open for nearly a month without a maintainer review. Could we get it reviewed alongside this one? Without both, the adapter write-side fix here never actually has a signature to write because it's dropped upstream in the pipeline.

If preferred, I'm happy to pull those changes into this PR (with attribution to @houmark) so it's a single comprehensive fix. Just let me know which you'd like.

mattsoltani:
I see @pemontto and agree with your points. Don't know about the preference here, but would be good to get both fixes in so we can remove the local patches. @AlemTuzlak what do you think is the best path forward?

AlemTuzlak:
Would you mind pulling it in? Also adding e2e tests if possible?

AlemTuzlak:
also #404 should use modelOptions instead of providerOptions

pemontto:
@AlemTuzlak before pulling #404 in: it threads `providerMetadata?: Record<string, unknown>` through `ToolCallPart`/`InternalToolCallState`, matching the existing `ToolCall.providerMetadata` shape from #401. Want that as-is, or refactored into a per-adapter typed metadata generic (similar to `TMessageMetadataByModality`)?

AlemTuzlak:
@pemontto I think that's outdated, we moved away from providerMetadata and providerOptions to modelOptions and modelMetadata

pemontto:
@AlemTuzlak I might be being obtuse, I can see `modelOptions` but no trace of `modelMetadata` in the codebase. Closest typed pattern I can find is `messageMetadataByModality` / `metadata?: TMetadata` on content parts. Is the intent to apply that same shape to `ToolCall` (typed `metadata?` via a per-adapter generic), replacing the generic `providerMetadata: Record<string, unknown>`?

AlemTuzlak:
Yeah then it should be metadata, i hallucinated providerOptions 💀

pemontto:
Rebased on current main and pulled in #404's plumbing under the new naming.

- `ToolCall<TMetadata>.metadata?: TMetadata`, threaded via a new `TToolCallMetadata` generic on `BaseTextAdapter` (same shape as the `MessageMetadataByModality` pattern on content parts).
- `ToolCallPart` likewise generic with typed `metadata?`.
- `ToolCallStartEvent.metadata` stays `Record<string, unknown>` because making it generic breaks the AGUIEvent discriminated-union narrowing.
- New `GeminiToolCallMetadata { thoughtSignature?: string }`. Gemini adapter binds the generic so `toolCall.metadata?.thoughtSignature` is typed end-to-end.
- `providerMetadata` renamed to `metadata` everywhere (breaking, hence `@tanstack/ai` minor).
- 711 ai tests + 92 ai-gemini tests pass, 31 projects typecheck clean.

Co-authored attribution to @houmark for #404.

houmark:
Looking forward to this one making it in so we can drop our temporary custom adaptors.

mattsoltani:
@AlemTuzlak Is there anything else needed to merge this?

AlemTuzlak (approved):

PR #469, opened 2026-04-18
feat(ai-openrouter): capture per-request cost from chat responses

### LINKED ISSUE #468: feat(ai-openrouter): surface per-request cost on RUN_FINISHED
## Problem

OpenRouter returns the authoritative per-request USD cost **inline in every chat completion response** (`usage.cost`, plus `usage.cost_details` with `upstream_inference_cost` and `cache_discount`) — see https://openrouter.ai/docs/use-cases/usage-accounting. TanStack AI's `@tanstack/ai-openrouter` adapter currently discards this data, so apps that rely on OpenRouter for prod traffic have no first-class way to observe spend from the adapter's event stream.

Two things make this tricky to solve at the call site:

1. **The `@openrouter/sdk` Zod parser strips the fields.** The SDK's response schema doesn't declare `cost` / `cost_details`, so by the time a chunk reaches the adapter, the fields are gone. A separate `/generation?id=…` lookup is possible but costs an extra HTTP round-trip per request.
2. **Cost can't be reconstructed locally from tokens × price.** OpenRouter routes the same model id to different upstream providers (primary, fallback, BYOK), each with different prices, and applies cache discounts and BYOK upstream costs the SDK can't see. A static price table would silently drift and produce wrong numbers.

## Proposal

Capture cost **out-of-band, from the same response**, via the SDK's public `HTTPClient.addHook('response', …)` API:

- The hook calls `Response.clone()` (tees the body via `ReadableStream.tee()`) and parses the clone to pull `usage.cost` / `usage.cost_details` before the Zod parser sees them.
- The SDK's stream consumer reads the other branch untouched — **no extra HTTP request, no added end-of-stream latency.** Cost arrives in the trailing SSE chunk, which the adapter was already waiting on for final token usage.
- Captured values are attached to `RUN_FINISHED` under `usage.cost` (USD) and `usage.costDetails.{upstreamInferenceCost, cacheDiscount}`.

### Type surface (additive, backwards-compatible)

Extend `RunFinishedEvent.usage` in `@tanstack/ai` with optional `cost` and `costDetails`. The middleware `UsageInfo` (`onUsage`) and `FinishInfo.usage` (`onFinish`) reuse the same shape so middleware authors can read cost without casts. Adapters that don't populate cost are unaffected.

```ts
interface UsageTotals {
  promptTokens: number
  completionTokens: number
  totalTokens: number
  cost?: number
  costDetails?: {
    upstreamInferenceCost?: number
    cacheDiscount?: number
  }
}
```

### Consumer example

```ts
  if (chunk.type === 'RUN_FINISHED') {
    console.log('USD:', chunk.usage?.cost)
    console.log('Cache discount:', chunk.usage?.costDetails?.cacheDiscount)
  }
}
```

## Scope

- `@tanstack/ai-openrouter` — adapter attaches the cost-capture hook, defers `RUN_FINISHED` until the stream fully drains so trailing usage-only chunks are included.
- `@tanstack/ai` — `UsageTotals` type + additive fields on `RunFinishedEvent.usage`, `UsageInfo`, `FinishInfo.usage`.
- No change to other adapters. Other providers can opt in later via the same type surface.

## Alternatives considered

- **`/generation?id=…` lookup after the stream ends** — adds an extra HTTP round-trip per request; also tends to race OpenRouter's internal cost aggregation and can return `null` if called too quickly.
- **Local tokens × price table** — unreliable (see above).
- **New event type** (`COST_REPORTED`) instead of extending `usage` — more intrusive on consumers; `usage` is the natural home for cost and keeps the agent loop and middleware unaware of the new field if they don't want it.

Happy to open a PR — branch is ready and full PR suite (lint, types, build, tests, docs, knip, sherif) is green across all 40 projects.

season179:
## Changes

Capture OpenRouter's provider-reported per-request cost from chat completion responses and expose it on `RUN_FINISHED.usage`.

OpenRouter returns `usage.cost` and `usage.cost_details` inline in chat responses ([docs](https://openrouter.ai/docs/use-cases/usage-accounting)), but the `@openrouter/sdk` chat-completion parser strips fields that are not declared in its schema. This PR recovers those exact provider-returned fields from the streamed response and forwards them through TanStack AI.

Closes #468

## Why This Uses OpenRouter's Exact Cost

This does not compute cost locally from token counts.

OpenRouter can route the same model id to different upstream providers, fallback providers, BYOK providers, and cached-token paths. Those paths can have different effective prices and provider-specific breakdowns. A local token × price-table calculation would drift and could silently report the wrong amount.

This PR only forwards the cost amount and cost details OpenRouter actually returned.

## Implementation

- Adds a cost-capture response hook using the SDK's public `HTTPClient`.
- The hook only runs for streaming `/chat/completions` responses:
  - checks `text/event-stream` case-insensitively
  - matches against `URL.pathname`, not the full URL string
- The hook calls `Response.clone()` and parses the cloned SSE body out-of-band, so the SDK's normal stream consumer remains untouched.
- Captures `usage.cost` as the authoritative total.
- Captures numeric/null fields from `usage.cost_details`, camelCased under `usage.costDetails`.
- Drops orphan `cost_details` when `usage.cost` is absent, because a breakdown without the total cannot be reconciled.
- Stores captured cost by upstream response id so overlapping streams do not block each other.
- Clones caller-provided `httpClient` instances before appending the hook, preserving custom fetchers, retries, tracing, and existing hooks without mutating the caller's client.
- Defers `RUN_FINISHED` until the upstream stream fully drains so trailing usage-only chunks are included.

If cost is absent or parsing fails, the stream still completes normally; `RUN_FINISHED.usage` simply omits cost fields.

## API Surface

Adds shared `UsageTotals` fields:

- `usage.cost?: number`
- `usage.costDetails?: Record<string, number | null | undefined>`

`RunFinishedEvent.usage`, middleware `UsageInfo`, and `FinishInfo.usage` now share that shape so cost can be read consistently from runtime events and middleware callbacks.

`costDetails` is intentionally generic. OpenRouter's current fields, such as `upstreamInferenceCost`, `upstreamInferenceInputCost`, `upstreamInferenceOutputCost`, and `cacheDiscount`, are preserved at runtime without hard-coding OpenRouter-specific keys into the shared `@tanstack/ai` type.

## Docs + Release

- Adds a Cost Tracking section to `docs/adapters/openrouter.md`.
- Adds README mention for OpenRouter cost tracking.
- Adds a minor changeset for `@tanstack/ai-openrouter` and `@tanstack/ai`.

## Validation

```bash
pnpm exec eslint packages/typescript/ai-openrouter/src/adapters/cost-capture.ts packages/typescript/ai-openrouter/src/adapters/text.ts packages/typescript/ai-openrouter/tests/cost-capture.test.ts packages/typescript/ai-openrouter/tests/openrouter-adapter.test.ts packages/typescript/ai/src/activities/chat/middleware/types.ts packages/typescript/ai/src/types.ts

pnpm --filter @tanstack/ai-openrouter test:lib -- cost-capture.test.ts openrouter-adapter.test.ts

git diff --check
```

`@tanstack/ai-openrouter` unit tests pass: 4 files, 69 tests.

Known unrelated typecheck failures remain on latest `main`:

- `packages/typescript/ai-openrouter/src/adapters/image.ts`
- existing audio/speech/transcription event-name errors in `@tanstack/ai`

## Checklist


season179:
Hi @AlemTuzlak, any chance you could take a look at this? Happy to address any feedback.

AlemTuzlak:
@season179 can you check if this is still an issue on the latest version of tanstack ai, I've fixed a bug in our chunk processor that indeed did stop usage from going through properly, this fix might not be needed at all

season179:
Thanks @AlemTuzlak, I checked this against the latest TanStack AI changes.

The recent chunk processor fix does help token usage flow through correctly, but this PR is still needed for a different reason: it exposes OpenRouter’s provider-reported `usage.cost` and `usage.cost_details` directly from the chat response.

The goal here is not to compute cost from token usage. OpenRouter’s actual cost can vary based on routing, fallback providers, BYOK, caching, and provider-specific pricing details, so token × price-table math can be wrong. This PR preserves the exact cost amount and cost breakdown returned by OpenRouter itself and attaches it to `RUN_FINISHED.usage`.

I’ve updated the PR to make that scope clearer.

AaronTan11:
i agree with this PR, @AlemTuzlak would love to have this merged 🙌

AlemTuzlak:
@season179 hey man, just talked to the opemrouter team directly, they will fix it on their end and then I'll update the sdk and resolve this without any need for workarounds!

ping-Toven:
Looking into it!

tombeckenham:
Hey @season179 the openrouter team have now included cost in their usage response. The feature is still missing in our adapter. Do you want to open a new pr? Otherwise we can update it

season179:
Thanks for informing, @tombeckenham. I think I'll open a new PR. But let me close this one and first study what has changed.

Issue #309, opened 2026-02-21
Streaming responses incompatible with Expo server runtime


GoestaHuppenbauer:
### TanStack AI version

0.5.1

### Framework/Library version

Expo SDK 52, Expo Router v3, @expo/server

### Description
Expo's server runtime does not support streaming ReadableStream responses in API routes. The response is never sent to the client and isLoading immediately returns to false.

Both toServerSentEventsResponse and toHttpResponse return ReadableStream-based responses which crash in Expo's server runtime with:

```
TypeError: Cannot read properties of undefined (reading 'statusText')
at respond (@expo/server/src/vendor/http.ts:99:31)
```

### Steps to reproduce:

1. Create an Expo API route
2. Use chat() to create a stream
3. Return _toServerSentEventsResponse(stream)_ or t_oHttpResponse(stream)_
4. Call the route from a React Native client

### Workaround: 
Collect all chunks and return as a single JSON response:
```
const chunks = [];
for await (const chunk of stream) {
  chunks.push(chunk);
}
return new Response(JSON.stringify(chunks), {
  headers: { 'Content-Type': 'application/json' },
});

```

### Your Minimal, Reproducible Example - (Sandbox Highly Recommended)

Snack

### Do you intend to try to help solve this bug with your own PR?

Yes, I think I know how to fix it and will discuss it in the comments of this issue

### Terms & Code of Conduct

- [x] I agree to follow this project's Code of Conduct
- [x] I understand that if my bug cannot be reliable reproduced in a debuggable environment, it will probably not be fixed and this issue may even be closed.

AlemTuzlak:
this is fixed with latest main!

Issue #308, opened 2026-02-21
@tanstack/ai packages cannot be resolved by Metro bundler due to ESM-only exports


GoestaHuppenbauer:
### TanStack AI version

0.5.1

### Framework/Library version

 Expo SDK 52, Metro bundler, React Native

### Describe the bug and the steps to reproduce it

All @tanstack/ai packages use ESM-only exports (only import condition in exports map). Metro bundler in React Native/Expo cannot resolve them, including subpath exports like ./adapters. This affects both client-side bundling and API route bundling even with unstable_enablePackageExports: true and unstable_conditionNames configured.

`Cannot resolve @tanstack/ai/adapters`

Workaround: Manually add default condition to all exports in node_modules/@tanstack/ai/package.json:

```
json"./adapters": {
  "types": "./dist/esm/activities/index.d.ts",
  "default": "./dist/esm/activities/index.js",
  "import": "./dist/esm/activities/index.js"
}
```

Fix required: Add require or default condition to all package exports.

### Your Minimal, Reproducible Example - (Sandbox Highly Recommended)

Snack

### Do you intend to try to help solve this bug with your own PR?

Yes, I think I know how to fix it and will discuss it in the comments of this issue

### Terms & Code of Conduct

- [x] I agree to follow this project's Code of Conduct
- [x] I understand that if my bug cannot be reliable reproduced in a debuggable environment, it will probably not be fixed and this issue may even be closed.

AlemTuzlak:
we have an example app using metro + expo on latest main, use it as a reference and see if this is still an issue, i managed to create a working expo app without cjs! if i'm wrong please lmk and feel free to re-open

PR #478, opened 2026-04-20
feat: dual ESM+CJS builds + toJSONResponse/fetchJSON for non-streaming runtimes

AlemTuzlak:
## Summary

Two related fixes for Expo / Metro / non-streaming runtimes.

### #308 — dual ESM + CJS output

\`@tanstack/ai\`, \`@tanstack/ai-client\`, and \`@tanstack/ai-event-client\` were ESM-only (\`import\` condition only, no \`require\` / \`default\`). Metro can't resolve that configuration, even with \`unstable_enablePackageExports: true\`, so consumers saw \`Cannot resolve @tanstack/ai/adapters\` etc.

Changes:
- \`vite.config.ts\` for all three packages flipped to \`cjs: true\` — emits \`dist/esm/**/*.js\` + \`.d.ts\` **and** \`dist/cjs/**/*.cjs\` + \`.d.cts\`.
- \`package.json\` \`exports\` for \`.\`, \`./adapters\`, \`./middlewares\` now use the nested \`import\`/\`require\` shape with type-aware conditions.
- Added \`main\` pointing at the CJS entry so legacy resolvers still work.
- \`publint --strict\` passes.

### #309 — \`toJSONResponse\` + \`fetchJSON\`

Expo's \`@expo/server\` can't emit \`ReadableStream\` responses, so \`toServerSentEventsResponse\` / \`toHttpResponse\` crash with \`Cannot read properties of undefined (reading 'statusText')\`.

Changes:
- **\`@tanstack/ai\`: \`toJSONResponse(stream, init?)\`** — drains the stream, returns \`new Response(JSON.stringify(chunks), { Content-Type: application/json })\`. Honours caller-provided headers / status; aborts a supplied \`abortController\` if the upstream throws.
- **\`@tanstack/ai-client\`: \`fetchJSON(url, options?)\`** — matching connection adapter. POSTs \`{ messages, data }\`, parses the response as a \`StreamChunk[]\`, yields each chunk into the normal \`ChatClient\` pipeline.

Trade-off: you lose incremental rendering — the UI sees everything at once when the request resolves. Docs in both JSDoc blocks call this out and tell users to prefer SSE / HTTP-stream when the runtime supports them.

## Test plan

- [x] Unit tests for \`toJSONResponse\`: defaults, header passthrough, custom \`Content-Type\`, abort-on-upstream-error
- [x] \`pnpm test\` across all 41 projects
- [x] \`test:build\` (publint strict) green on all three packages — dual exports map validated

## Summary by CodeRabbit

* **New Features**
  * Dual ESM + CommonJS distributions for packages to improve CJS compatibility
  * Added toJSONResponse — server helper returning a chat stream as a single JSON-array Response
  * Added fetchJSON — client adapter that fetches JSON-array chat responses and replays them into the chat pipeline

* **Documentation**
  * Guides for non-streaming runtimes (React Native / Expo) and JSON-adapter usage

* **Tests**
  * Tests covering fetchJSON behavior, error cases, headers, and abort handling


packages/typescript/ai-client/src/connection-adapters.ts
```diff
@@ -424,6 +424,81 @@ export function fetchHttpStream(
...
+      }
+
+      const payload = (await response.json()) as unknown
+      if (!Array.isArray(payload)) {
+        throw new Error(
+          'fetchJSON: expected response body to be a JSON array of StreamChunks. Did you forget to use `toJSONResponse(stream)` on the server?',
+        )
+      }
+      for (const chunk of payload) {
+        yield chunk as StreamChunk
```
tombeckenham, line 523: You never check for the abort signal in this yield loop despite adding it in toJSONResponse

packages/typescript/ai-client/src/connection-adapters.ts
```diff
@@ -424,6 +424,81 @@ export function fetchHttpStream(
...
+        signal: abortSignal || resolvedOptions.signal,
+      })
+
+      if (!response.ok) {
+        throw new Error(
+          `HTTP error! status: ${response.status} ${response.statusText}`,
+        )
+      }
+
+      const payload = (await response.json()) as unknown
```
tombeckenham: You need to wrap this in try catch otherwise the user will get an annoying "Unexpected token < in JSON at position 0" error instead of what probably happened e.g. a gateway error or something that returned html


```suggestion
  let payload: unknown
  try {
    payload = await response.json()
  } catch (err) {
    const cause = err instanceof Error ? err.message : String(err)
    throw new Error(
      `fetchJSON: failed to parse response body as JSON from ${resolvedUrl} (status ${response.status}):
  ${cause}`,
      { cause: err },
    )
  }
```

packages/typescript/ai-client/src/connection-adapters.ts
```diff
@@ -424,6 +424,81 @@ export function fetchHttpStream(
...
+      const fetchClient = resolvedOptions.fetchClient ?? fetch
+      const response = await fetchClient(resolvedUrl, {
+        method: 'POST',
+        headers: requestHeaders,
+        body: JSON.stringify(requestBody),
+        credentials: resolvedOptions.credentials || 'same-origin',
+        signal: abortSignal || resolvedOptions.signal,
+      })
+
+      if (!response.ok) {
```
tombeckenham, line 483: servers usually put the actual diagnostic information in the body. e.g. OpenAI/Anthropic upstream rate limit: {"error":{"type":"rate_limit_error","message":"Rate limit exceeded for...","retryAfter":42}}


```suggestion
   if (!response.ok) {
    let bodySnippet = ''
    try {
      const text = await response.text()
      bodySnippet = text.length > 500 ? `${text.slice(0, 500)}…` : text
    } catch {
      // body unreadable, fall through with status only
    }
    throw new Error(
      `HTTP error! status: ${response.status} ${response.statusText}${
        bodySnippet ? ` — ${bodySnippet}` : ''
      }`,
    )
  }
```

packages/typescript/ai-client/src/connection-adapters.ts
```diff
@@ -424,6 +424,81 @@ export function fetchHttpStream(
...
+ *
+ * @example
+ * ```typescript
+ * // Expo / RN client that hits an Expo API route returning toJSONResponse(stream)
+ * const connection = fetchJSON('/api/chat')
+ *
+ * const client = new ChatClient({ connection })
+ * ```
+ */
+export function fetchJSON(
```
tombeckenham, line 451: You should add an e2e test for the roundtrip toJSONResponse → fetchJSON

tombeckenham (changes requested):
A couple of small changes to make would make this more solid

PR #499, opened 2026-04-23
chore(adapters): move common options to model options

harry-whorlow:
## Summary by CodeRabbit

* **New Features**
  * Added temperature and topP sampling control options for Gemini model configuration.

* **Refactor**
  * Restructured how temperature, max tokens, and top_p parameters are handled across AI providers. These settings now flow through model-scoped provider options rather than top-level configuration fields.

* **Tests**
  * Updated e2e smoke test routing configuration.


AlemTuzlak:
Thanks for kicking this off, @harry-whorlow! 🙏 This branch was built on the old `packages/typescript/` layout and couldn't be cleanly rebased onto the restructured `packages/` tree, so I've reimplemented the idea on current `main` and carried it across all 7 providers (with a codemod, docs, and tests) in #660. Closing in favor of that — your original approach is credited there.

PR #502, opened 2026-04-24
feat(ai-gemini): add geminiTextInteractions() adapter for stateful Interactions API

### LINKED ISSUE #501: feat(ai-gemini): expose stateful Interactions API (previous_interaction_id)
## Problem

The `@google/genai` SDK (v1.43+) exposes a preview **Interactions API** that provides server-side conversation state — the Gemini equivalent of OpenAI's Responses API (`previous_response_id` / `conversation`). This lets callers skip storing and replaying message history on their side.

```ts
// @google/genai
const interaction1 = await ai.interactions.create({
  model: 'gemini-2.5-flash',
  input: 'Hi, my name is Amir.',
})

const interaction2 = await ai.interactions.create({
  model: 'gemini-2.5-flash',
  input: 'What is my name?',
  previous_interaction_id: interaction1.id, // server-side state
})
```

Today the `ai-gemini` text adapter routes through the stateless `client.models.generateContent` / `generateContentStream` path (`packages/typescript/ai-gemini/src/adapters/text.ts:131,177`), so there's no way to take advantage of this — even via `providerOptions`.

For comparison, `ai-openai` already supports the equivalent pattern: it uses the Responses API (`client.responses.create`) and exposes `previous_response_id` / `conversation` via `providerOptions` (`packages/typescript/ai-openai/src/text/text-provider-options.ts:30,49`).

## Proposal

Add first-class support for Gemini's Interactions API. Two reasonable shapes:

1. **New tree-shakeable adapter** — e.g. `geminiInteractions()` alongside `geminiText()`, calling `client.interactions.create` under the hood. Matches the existing tree-shakeable adapter convention.
2. **Opt-in via `providerOptions`** — e.g. `providerOptions.previous_interaction_id` on the existing text adapter, with the adapter routing to `client.interactions.create` when that option (or a dedicated flag) is present.

Option 1 is probably cleaner given the Interactions API has a different request/response shape than `generateContent`, and it keeps the stateless path simple.

## Notes / caveats

- The Interactions API is currently marked **preview/experimental** in `@google/genai` — worth flagging in docs and JSDoc.
- Response-id capture: callers need a way to read the interaction id off the returned response to pass back next turn. Needs a small addition to the chat result shape, or a provider-specific metadata field.
- Tool calling, streaming, and multimodal input parity with the existing text adapter should be confirmed against the Interactions API surface before shipping.

## Context

Came up in discussion — users want to avoid plumbing conversation history through their own storage when the provider already offers it server-side. OpenAI side is already covered; Gemini side is the gap.

tombeckenham:
Closes #501 

## 🎯 Changes

Adds an experimental `geminiTextInteractions()` adapter that routes through Gemini's stateful Interactions API (Google Beta) instead of `client.models.generateContent`, so callers can chain turns server-side via `previous_interaction_id` rather than replaying the full message history every turn.

**What's in scope:**

- Text/chat output with function tools.
- Built-in Gemini tools `google_search`, `code_execution`, `url_context`, `file_search`, and `computer_use`. Activity is surfaced as AG-UI `CUSTOM` events named `gemini.googleSearchCall` / `gemini.googleSearchResult` (and the matching `codeExecutionCall`/`Result`, `urlContextCall`/`Result`, `fileSearchCall`/`Result` variants), carrying the raw Interactions delta payload. Function-tool `TOOL_CALL_*` events are unchanged, and `finishReason` stays `stop` when only built-in tools ran.
- Stateful chaining via `modelOptions.previous_interaction_id`. The server-assigned interaction id is surfaced as an AG-UI `CUSTOM` event with `name: 'gemini.interactionId'` emitted just before `RUN_FINISHED`.
- Provider options derived from `Pick<>` over `@google/genai`'s own `CreateModelInteractionParamsStreaming`, so field shapes (allowed `response_modalities`, `generation_config`, etc.) stay in sync with the SDK automatically. Unsupported media mime types are rejected at runtime against the SDK's allowed sets, pinned via `satisfies`.

**Not supported on this adapter** (each throws a targeted error pointing to the alternative): `google_search_retrieval`, `google_maps`, `mcp_server`, and image/audio output via Interactions. Use `geminiText()`, `geminiImage`, or `geminiSpeech` for those.

**Import path:**

```ts
import { geminiTextInteractions } from '@tanstack/ai-gemini/experimental'
```

The adapter is exported from a dedicated `@tanstack/ai-gemini/experimental` subpath so the experimental status is load-bearing in callers' editors and bundles. Sources live at `packages/typescript/ai-gemini/src/experimental/text-interactions/`.

### Diverges from the original draft

The PR has evolved during review; calling out the key shifts vs. the initial summary:

- **interactionId surfacing.** The first draft added a `providerMetadata?: Record<string, unknown>` field to `RunFinishedEvent` and exposed the id as `providerMetadata.interactionId`. That core-type change was reverted; the id now flows through an AG-UI `CUSTOM` event (`name: 'gemini.interactionId'`) so the core types stay untouched and the surface is consistent with how other adapters surface provider-specific signals.
- **Built-in tools.** Originally these all threw "use `geminiText()`" errors. `google_search`, `code_execution`, `url_context`, `file_search`, and `computer_use` now work end-to-end — the Interactions API uses snake_case tool shapes that needed translation.
- **Experimental subpath.** New: the adapter is gated behind `@tanstack/ai-gemini/experimental`, not the package root. Cleanest signal that callers are opting into a Beta API.
- **Changesets.** The follow-up "built-in tools" changeset was consolidated into the main changeset so the published changelog describes the unreleased adapter as one feature instead of an "added it" / "actually it does support built-ins" flip-flop.
- **Tests.** Adapter unit tests now route through the public `chat()` core function rather than calling the adapter directly, matching project convention. Coverage grew from ~74 to **105 passing**.
- **Error shape.** `RUN_ERROR` now emits the spec-compliant flat `{ message, code }` shape rather than nesting under `error: {...}`.
- **Example UX.** `ts-react-chat` model selector labels the Interactions models with `(experimental)` to mirror the experimental subpath.

## ✅ Checklist

- [x] Verified locally per package: `pnpm --filter @tanstack/ai-gemini test:lib` (106/106), `test:types`, `test:eslint`, `test:build` (publint --strict clean).
- [x] E2E: spec at `testing/e2e/tests/stateful-interactions.spec.ts` is live — covers the two-turn `previous_interaction_id` chain via aimock 1.24.1 (`@copilotkit/aimock` shipped Gemini Interactions handler in 1.18 / [PR #139](https://github.com/CopilotKit/aimock/pull/139)).

## 🔭 Follow-ups

- #570 — native `structuredOutputStream` for `geminiText` and `geminiTextInteractions`. Both adapters currently fall through to the buffered `fallbackStructuredOutputStream` helper; truly incremental JSON streaming is tracked there as a follow-up.

## 🚀 Release Impact

- [x] This change affects published code; a single consolidated [changeset](.changeset/gemini-text-interactions.md) is included (`@tanstack/ai-gemini` — minor).

## Summary by CodeRabbit

* **New Features**
  * Experimental geminiTextInteractions adapter for stateful multi-turn conversations; emits and reuses server interaction IDs.

* **Documentation**
  * Usage guide for the stateful adapter, interaction ID capture/reuse, provider options, and caveats.

* **Tests**
  * Added unit and E2E tests for streaming, interaction ID emission/chaining, tool-call flows, structured JSON output, and provider feature gating.

* **Examples**
  * Updated chat example UI and model selector to support the new stateful Gemini provider and interaction ID handling.

---

### Incidental fix included: AudioGenUI route wiring (regression from #527)

While verifying E2E green on this branch I tripped over 4 failing tests (`elevenlabs -- audio-gen` × 2, `elevenlabs -- sound-effects` × 2) and traced them to a regression in [#527](https://github.com/TanStack/ai/pull/527). That PR ("streaming structured output …") silently dropped the AudioGenUI scaffold from `testing/e2e/src/routes/$provider/$feature.tsx`:

- the `AudioGenUI` import
- the `'audio-gen'` / `'sound-effects'` entries in `MEDIA_FEATURES`
- the `case 'audio-gen': case 'sound-effects':` block in `MediaFeature`

The removal isn't mentioned in #527's title, body, or any of its ~30 commit messages, and is orthogonal to its stated scope. Everything else for audio-gen (matrix, types, features, provider/media-provider factories, fixtures, spec files, `AudioGenUI.tsx` itself) was left intact on main, leaving the spec files orphaned — `providersFor('audio-gen')` returns `[]` on main so the loops generate zero tests and nothing complains.

The matrix changes on this branch re-activate those loops, which is why the regression surfaced here first. Restored the dropped lines in `04081dfb` so this branch is green; happy to spin the AudioGenUI restoration out into a separate fix-PR against main if that's preferred.


docs/adapters/gemini.md
```diff
@@ -110,6 +110,192 @@ const stream = chat({
...
+
+// Turn 1: introduce yourself, capture the interaction id.
+let interactionId: string | undefined;
+
+for await (const chunk of chat({
+  adapter: geminiTextInteractions("gemini-2.5-flash"),
+  messages: [{ role: "user", content: "Hi, my name is Amir." }],
+})) {
+  if (chunk.type === "CUSTOM" && chunk.name === "gemini.interactionId") {
+    interactionId = (chunk.value as { interactionId?: string }).interactionId;
```
AlemTuzlak: errr we can't have this in the docs!

tombeckenham: Fixed

AlemTuzlak (approved):

PR #527, opened 2026-05-05
feat: streaming structured output across openai/openrouter/grok/groq + summarize fix

### LINKED ISSUE #526: feat(ai-openrouter): support streaming structured output (response_format json_schema with stream: true)
## Summary

OpenRouter natively supports combining `response_format: { type: 'json_schema', ... }` with `stream: true`, but our adapter currently performs structured output as a **separate non-streaming call**. This means a logical structured-output run produces two distinct HTTP requests (one streaming free-text, one non-streaming JSON), and callers can't consume the JSON incrementally.

OpenRouter docs: https://openrouter.ai/docs/guides/features/structured-outputs#streaming-with-structured-outputs

Prior discussion: #370 (also related: #195)

## Current behavior

`packages/typescript/ai-openrouter/src/adapters/text.ts`

- `chatStream()` (line 113) — `stream: true`, no `responseFormat`. Free-text only.
- `structuredOutput()` (line 266) — `stream: false`, sends `responseFormat: { type: 'json_schema', jsonSchema: { name, schema, strict: true } }`. Returns a `Promise`, no streaming.

The `TextAdapter` base interface (`packages/typescript/ai/src/activities/chat/adapter.ts:97`) hardcodes `structuredOutput` as `Promise<StructuredOutputResult<unknown>>`, so today there is no surface for incremental structured output even where the provider supports it.

## Proposal

Add a streaming structured-output path. Two shapes worth considering:

1. **New adapter method `structuredOutputStream(options)` returning `AsyncIterable<StreamChunk>`**, with chunks carrying partial JSON deltas (we already depend on `partial-json` for the same purpose elsewhere). Adapters that don't natively support it can fall back to the existing non-streaming `structuredOutput`.
2. **Extend `chatStream` to accept an `outputSchema` option** and have OpenRouter emit `responseFormat` alongside `stream: true`. Simpler wire change, but couples structured-output semantics into the general chat path and makes the typed result harder to expose.

(1) is probably cleaner and parallels the `chatStream` / `structuredOutput` split we already have.

## Acceptance criteria

- `@tanstack/ai-openrouter` exposes streaming structured output that emits a single HTTP request with `stream: true` + `response_format: json_schema` (strict).
- Partial JSON deltas are surfaced as they arrive (parsed via `partial-json`).
- Final result validates against the supplied schema (re-using `convertSchemaToJsonSchema(..., { forStructuredOutput: true })` for the strict transform).
- E2E coverage in `testing/e2e` (mandatory per `CLAUDE.md`) — fixture + spec under the structured-output scenarios, with OpenRouter listed in `feature-support.ts` / `test-matrix.ts`.
- Other adapters: at minimum, default to the existing non-streaming path so the new method is callable everywhere without breaking changes. Native streaming JSON for OpenAI/Anthropic/Gemini can land separately.

## Notes / risks

- Tool calls + structured output in the same request: OpenRouter's docs allow it but behavior varies by upstream provider — worth a smoke test, not a blocker.
- Strict schema transform already lives in `ai`; reuse it, don't fork.
- The base `TextAdapter` interface change is a breaking surface for downstream adapters — mitigated if we add `structuredOutputStream` as optional with a default that wraps `structuredOutput`.

tombeckenham:
## 🎯 Changes

Started as openrouter-only (#526) and grew into a multi-package effort: typed streaming structured output across the four OpenAI-compatible providers, a Chat Completions sibling for OpenAI, a fix for streaming summarize, the decoupling of `@tanstack/ai-openrouter` from the shared OpenAI base, and a hand-testable example to exercise the whole surface.

### Core — `@tanstack/ai`

- New `chat({ outputSchema, stream: true })` overload returning `StructuredOutputStream<InferSchemaType<TSchema>>`. Yields raw JSON deltas plus a terminal `CUSTOM { name: 'structured-output.complete', value: { object, raw, reasoning? } }` event.
- `StructuredOutputStream<T>` is a discriminated union over three tagged `CUSTOM` variants — `structured-output.complete<T>`, `approval-requested`, and `tool-input-available` (new `ApprovalRequestedEvent` / `ToolInputAvailableEvent` interfaces exported from `@tanstack/ai`). Consumers narrow with a plain `chunk.type === 'CUSTOM' && chunk.name === '<literal>'` and get a fully-typed `chunk.value` — no helper or cast required. The bare `CustomEvent` (`value: any`) is excluded from the union on purpose to keep the narrow from collapsing to `any`; user-emitted events via the `emitCustomEvent` context API still flow at runtime as a documented residual gap.
- Activity-layer hardening: always-finalise after the stream loop (no silent hangs on missing `finishReason`), typed `RUN_ERROR` on empty content, mid-stream provider errors terminate cleanly, schema-validation failures carry `runId / model / timestamp`.
- `fallbackStructuredOutputStream` in the activity layer is the single source of truth for adapters that don't implement `structuredOutputStream` natively; `BaseTextAdapter` no longer ships a default.
- `ChatStreamSummarizeAdapter.summarizeStream` accumulates summary text and emits a terminal `CUSTOM { name: 'generation:result' }` event before the final `RUN_FINISHED`. Fixes `useSummarize` never populating `result` over streaming connections (the client only sets `result` on that specific CUSTOM event).

### Base — `@tanstack/openai-base`

- Renamed from `@tanstack/ai-openai-compatible`; the old name remains published for pinned lockfiles but receives no further updates.
- Centralised `structuredOutputStream` on `OpenAIBaseChatCompletionsTextAdapter` (uses `response_format: { type: 'json_schema', strict: true }` + `stream: true`) and `OpenAIBaseResponsesTextAdapter` (uses `text.format: { type: 'json_schema', strict: true }` + `stream: true`). Both call the OpenAI SDK directly post-#543's "adopt openai SDK" refactor.
- Cross-SDK abort detection hook (`isAbortError`) so subclasses can map provider-specific abort errors to `code: 'aborted'` without owning the rest of the finalisation path.
- Per-chunk `logger.provider(...)` debug logging now fires inside `structuredOutputStream` loops (mirroring the existing `chatStream` pattern) so debug mode gives end-to-end chunk introspection for the structured-output path.

### Provider adapters

| Adapter | API | Reasoning surface |
| --- | --- | --- |
| `@tanstack/ai-openai` `openaiText` | Responses | `response.reasoning_text.delta` + `response.reasoning_summary_text.delta` (requires `reasoning.summary: 'auto'`) |
| `@tanstack/ai-openai` `openaiChatCompletions` (new) | Chat Completions | reasoning emitted silently — Chat Completions has no `reasoning.summary` opt-in |
| `@tanstack/ai-grok` `grokText` | Chat Completions | `delta.reasoning_content` (DeepSeek convention; not typed by OpenAI SDK) |
| `@tanstack/ai-groq` `groqText` | Chat Completions | `delta.reasoning` (requires `reasoning_format: 'parsed'`; not typed by groq-sdk) |
| `@tanstack/ai-openrouter` `openRouterText` | Chat Completions | `delta.reasoningDetails` (camelCase) |
| `@tanstack/ai-openrouter` `openRouterResponsesText` (beta) | Responses (beta) | `response.reasoning_text.delta` + `response.reasoning_summary_text.delta` via `normalizeStreamEvent` |

All six emit the contractual `REASONING_*` lifecycle (`REASONING_START` → `REASONING_MESSAGE_START` → `REASONING_MESSAGE_CONTENT` deltas → `REASONING_MESSAGE_END` → `REASONING_END`) and close it before `TEXT_MESSAGE_START`. Accumulated reasoning is also surfaced on `structured-output.complete.value.reasoning` for consumers that only subscribe to the terminal event. OpenRouter SDK's proprietary `RequestAbortedError` is mapped (alongside DOM `AbortError`) to `code: 'aborted'` in the two openrouter adapters.

`@tanstack/ai-openai` also exports a new `OpenAIChatCompletionsTextAdapter` / `openaiChatCompletions` / `createOpenaiChatCompletions` factory — a sibling to the existing Responses adapter for callers who want the older `/v1/chat/completions` wire format against the OpenAI SDK.

`@tanstack/ai-openrouter` is now decoupled from the shared OpenAI base and reads `@openrouter/sdk`'s camelCase types natively (no more snake_case ↔ camelCase round-trips). It extends `BaseTextAdapter` directly; `@tanstack/openai-base` and `openai` are removed from its deps. Public OpenRouter API is unchanged.

### `ts-react-chat` example

`/generations/structured-output` is now a 6-way demo of the entire feature surface:

- Provider/API matrix: OpenAI (Responses), OpenAI (Chat Completions), Grok, Groq, OpenRouter (Chat Completions), OpenRouter (Responses beta) — each with its own model dropdown.
- `Stream` toggle: off uses non-streaming `chat({ outputSchema })`, on uses `structuredOutputStream`.
- Progressive UI rendering via `parsePartialJSON` — title, summary, recommendation cards (brand → name → type → price → reason), and next steps fill in field-by-field as JSON streams in, snapping to the validated payload on the terminal event.
- Live "Thinking" strip rendering the latest reasoning sentence from `REASONING_MESSAGE_CONTENT` deltas.
- Per-provider reasoning opt-ins via `modelOptions` (gpt-5.x/o-series → `reasoning.summary: 'auto'`, groq gpt-oss/qwen3/kimi-k2 → `reasoning_format: 'parsed'`, openrouter → `reasoning.effort: 'medium'`).
- Sidebar label cleaned up: "Structured Output (OpenRouter)" → "Structured Output" (it now spans every provider).

`/generations/summarize` overhauled too:

- New model picker (gpt-4o-mini through gpt-5.2) plumbed through both API route and server-fn paths.
- Streaming and Server-Fn modes now visibly stream — wired an `onChunk` handler that accumulates `TEXT_MESSAGE_CONTENT` deltas into local state so the summary renders token-by-token, with a "streaming…" indicator next to the heading.
- Dropped the hard-coded `maxLength: 200`. On Responses-API reasoning models `maxLength` maps to `max_output_tokens` which covers reasoning + visible output combined; a 200-token cap consumed the whole budget on hidden reasoning, returning truncated responses with no summary. The `concise` / `bullet-points` / `paragraph` prompt instruction is enough to drive length.

### E2E

`testing/e2e/src/lib/feature-support.ts` — `structured-output-stream` set expanded from `['openrouter']` to all four providers. The parameterised spec in `testing/e2e/tests/structured-output-stream.spec.ts` (happy path + abort) now runs across each.

## ✅ Checklist

Verified during development: `pnpm test:lib`, `pnpm test:types`, `pnpm test:eslint`, `pnpm test:build`, `pnpm test:knip` across affected packages. Hand-tested in `examples/ts-react-chat` against each provider for both streaming and non-streaming. E2E (`pnpm --filter @tanstack/ai-e2e test:e2e -- --grep structured-output-stream`) needs to run on a host where port 4010 is free.

## 🚀 Release Impact

One consolidated changeset (`.changeset/streaming-structured-output.md`) covers the union of version bumps: minor on `@tanstack/ai`, `@tanstack/openai-base`, `@tanstack/ai-openai`, `@tanstack/ai-grok`, `@tanstack/ai-groq`, `@tanstack/ai-openrouter`; patch on `@tanstack/ai-anthropic`, `@tanstack/ai-gemini`, `@tanstack/ai-ollama`. The body retains the per-area sections from the originals (core, openai-base, provider adapters, openrouter decoupling, summarize subsystem) and adds the tagged-CustomEvent type design.


tombeckenham:
Structured streaming in action


https://github.com/user-attachments/assets/713675ff-7805-4137-a7ee-fd3c5bed5a3c

autofix-troubleshooter:
Hi! I'm the <a href="https://autofix.ci">autofix logo<strong>autofix.ci</strong></a> troubleshooter bot.

It looks like you correctly set up a CI job that uses the autofix.ci GitHub Action, but the [**autofix.ci GitHub App**](https://github.com/apps/autofix-ci/) has not been installed for this repository. This means that autofix.ci unfortunately does not have the permissions to fix this pull request. If you are the repository owner, please install the app and then restart the CI workflow! 😃

packages/typescript/ai-openrouter/src/adapters/responses-text.ts
```diff
@@ -258,6 +258,424 @@ export class OpenRouterResponsesTextAdapter<
...
+        timestamp,
+      } satisfies StreamChunk
+      yield {
+        type: EventType.STEP_STARTED,
+        stepName: stepId,
+        stepId,
+        model,
+        timestamp,
+        stepType: 'thinking',
+      } satisfies StreamChunk
```
AlemTuzlak: step started should be first i think

AlemTuzlak: okay after inspecting the previous method i don't think i know

packages/typescript/ai-openrouter/src/adapters/responses-text.ts
```diff
@@ -258,6 +258,424 @@ export class OpenRouterResponsesTextAdapter<
...
+    }.bind(this)
+
+    try {
+      chatOptions.logger.request(
+        `activity=structuredOutputStream provider=${this.name} model=${this.model} messages=${chatOptions.messages.length}`,
+        { provider: this.name, model: this.model },
+      )
+      const reqOptions = extractRequestOptions(chatOptions.request)
+      const rawStream = (await this.orClient.beta.responses.send(
```
AlemTuzlak: let's avoid casts here

packages/typescript/ai-openrouter/src/adapters/responses-text.ts
```diff
@@ -258,6 +258,424 @@ export class OpenRouterResponsesTextAdapter<
...
+          chunk.type === 'response.in_progress'
+        ) {
+          const r = chunk.response as { model?: string } | undefined
+          if (r?.model) model = r.model
+          continue
+        }
+
+        if (chunk.type === 'response.refusal.delta') {
+          const delta =
+            typeof (chunk as { delta?: unknown }).delta === 'string'
```
AlemTuzlak: lets avoid casts on chunks

packages/typescript/ai-openrouter/src/adapters/responses-text.ts
```diff
@@ -258,6 +258,424 @@ export class OpenRouterResponsesTextAdapter<
...
+          // call; TS can't see through the generator side-effect.
+          const messageId = reasoningMessageId!
+          accumulatedReasoning += reasoningDelta
+          yield {
+            type: EventType.REASONING_MESSAGE_CONTENT,
+            messageId,
+            delta: reasoningDelta,
+            model,
+            timestamp,
+          } satisfies StreamChunk
```
AlemTuzlak: satisfies is not required

packages/typescript/ai-openrouter/src/adapters/text.ts
```diff
@@ -271,6 +271,320 @@ export class OpenRouterTextAdapter<
...
+                strict: true,
+              },
+            },
+          },
+        },
+        {
+          signal: reqOptions.signal ?? undefined,
+          ...(reqOptions.headers && { headers: reqOptions.headers }),
+        },
+      )) as AsyncIterable<ChatStreamChunk>
```
AlemTuzlak: again, no casts

packages/typescript/ai/src/activities/chat/index.ts
```diff
@@ -1575,13 +1590,26 @@ class TextEngine<
...
 >(
   options: TextActivityOptions<TAdapter, TSchema, TStream>,
 ): TextActivityResult<TSchema, TStream> {
   const { outputSchema, stream } = options
 
-  // If outputSchema is provided, run agentic structured output
+  // outputSchema + stream:true is the only branch that streams structured
+  // output. Without an explicit `stream: true`, schema-bearing calls run the
+  // agent loop and resolve to a typed Promise<InferSchemaType<TSchema>>.
+  if (outputSchema && stream === true) {
```
AlemTuzlak, line 1602: just to confirm, have you confirmed that the shape of the return value changes based on stream: true/false + outputSchema properly?

packages/typescript/ai/src/activities/chat/index.ts
```diff
@@ -1741,6 +1769,302 @@ async function runAgenticStructuredOutput<TSchema extends SchemaInput>(
...
+  } satisfies StreamChunk
+
+  yield {
+    type: EventType.RUN_FINISHED,
+    runId,
+    threadId,
+    model,
+    timestamp,
+    finishReason: 'stop',
+  } satisfies StreamChunk
```
AlemTuzlak: pretty sure satisfies is not required anywhere

packages/typescript/ai/src/activities/chat/index.ts
```diff
@@ -1741,6 +1769,302 @@ async function runAgenticStructuredOutput<TSchema extends SchemaInput>(
...
+
+  if (textOptions.tools?.length) {
+    const engine = new TextEngine(
+      {
+        adapter,
+        params: { ...textOptions, model, logger } as TextOptions<
+          Record<string, unknown>,
+          Record<string, unknown>
+        >,
+        middleware,
```
AlemTuzlak: again, let's not use any casts!

AlemTuzlak (changes requested):
lets add an agent skill for this, also left some comments, great work!

packages/typescript/ai/src/activities/chat/index.ts
tombeckenham: packages/typescript/ai/tests/chat-result-types.test.ts:20 pins every combination
  with strict expectTypeOf(...).toEqualTypeOf(...)

packages/typescript/ai-openrouter/src/adapters/responses-text.ts
tombeckenham: Mirroring the order in `text.ts` (REASONING_START → REASONING_MESSAGE_START → STEP_STARTED) deliberately — the stream processor handles STEP_STARTED arriving after REASONING_START via `consumePendingThinkingStep`, so the open/close pairing is what matters, not the order within the pair.

AlemTuzlak (approved):

PR #545, opened 2026-05-11
refactor(openai-base): rename, adopt openai SDK, decouple ai-openrouter

### LINKED ISSUE #543: Migrate ai-groq, ai-openrouter, ai-ollama to openai-base + parameterize the base for SDK shape variance
Captured from [PR #409 comment](https://github.com/TanStack/ai/pull/409#issuecomment-4394567461) so the design proposal doesn't get lost regardless of how #409 lands.

## Context

After #409 merges, only `ai-openai` and `ai-grok` will extend the new `@tanstack/openai-base`. `ai-groq`, `ai-openrouter`, and `ai-ollama` still extend `BaseTextAdapter` directly — together that's ~2k lines of duplicated text-adapter code the bases were designed to absorb.

For `ai-groq` the migration looks mechanical (Chat Completions wire format, OpenAI SDK works fine pointed at Groq's endpoint). **`ai-openrouter` is the interesting one** — design question worth resolving before assuming it slots in.

## Wire format: identical. SDK shape: not.

OpenRouter's HTTP API is OpenAI-compatible Chat Completions, so on the wire the base's SSE accumulator, partial-JSON, tool-call buffers, and RUN_ERROR taxonomy all apply unchanged. But the **TypeScript SDK** (`@openrouter/sdk`, Speakeasy-generated) has a different shape from the OpenAI SDK the base assumes:

| | OpenAI SDK (base) | `@openrouter/sdk` |
|---|---|---|
| Call shape | `client.chat.completions.create(params)` | `client.chat.send({ chatRequest: params })` |
| Field naming | snake_case (`max_completion_tokens`, `top_p`) | camelCase (`maxCompletionTokens`, `topP`) |
| Abort error | `APIUserAbortError` | `RequestAbortedError` from `@openrouter/sdk/models/errors` |

Grok gets away with a ~60-line subclass because xAI's "OpenAI-compatible" endpoint accepts the OpenAI SDK verbatim — `withGrokDefaults` just rewrites the `baseURL`. OpenRouter's SDK is not a drop-in.

## OpenRouter-specific surface

These are real user-facing features, not incidental:

1. **Provider routing** — `provider: { order, fallbacks, allow_fallbacks, sort, ignore }`, `models[]` fallback chain, `route`, `transforms: ['middle-out']`. OpenRouter's distinguishing feature.
2. **Response metadata** — which provider served the request, which underlying model was actually used, generation ID for cost lookup.
3. **App attribution headers** — `HTTP-Referer`, `X-Title`.

## Considered and ruled out: dropping `@openrouter/sdk`

The cleanest-looking option would be replacing `@openrouter/sdk` with the OpenAI SDK pointed at `openrouter.ai/api/v1` (the grok approach). That deletes ~700 lines and the base works unchanged.

Ruled out: it forces provider routing through untyped `extra_body` pass-through and loses the typed response metadata. That's a type-safety regression on OpenRouter's headline feature, and the SDK is purpose-built around exactly that surface — keeping it is the point.

## Proposal: parameterize the base with two protected hooks

Add to `OpenAICompatibleChatCompletionsTextAdapter`:

```ts
protected callChatCompletion(params, signal): Promise<ChatCompletion>
protected callChatCompletionStream(params, signal): AsyncIterable<ChatCompletionChunk>
```

Default impl calls the OpenAI SDK (covers grok and any future "OpenAI-compatible-endpoint" providers like deepseek / together / fireworks). `ai-openrouter` overrides to call `@openrouter/sdk`.

Everything else — tool/message converters, SSE accumulator, partial-JSON, RUN_ERROR taxonomy, structured output, lifecycle — stays in the base. The `ai-openrouter` subclass collapses to ~150 lines (SDK shim + provider-routing typing + response metadata extraction) vs. 807 today.

## Why this matters for #527

PR #527 (streaming structured output) currently duplicates ~290 lines of `structuredOutput` + `chatStreamStructured` across openai/grok/groq. The plan there is to lift that into the bases so each wire format has one implementation. **That lift only covers openrouter if openrouter is on the Chat Completions base.** If #409 ships with openrouter unmigrated, #527 has to either keep duplicating into openrouter or fold the openrouter migration into a feature PR.

## Open questions

1. Was leaving `ai-groq` / `ai-openrouter` / `ai-ollama` on `BaseTextAdapter` deliberate (scoping #409 to "introduce bases, migrate later") or time-boxed?
2. Expand scope (here or in a follow-up PR) to migrate `ai-groq` + `ai-openrouter`?
3. Any concerns with the two-hook approach, or alternative shapes preferred?

Happy to prototype the hook signatures on a side branch if it'd help the conversation.

tombeckenham:
## 🎯 Changes

This PR consolidates the OpenAI-compatible provider stack and rewires `ai-groq` and `ai-openrouter`. The work landed in three arcs over many commits — the shape changed mid-flight, so the headline below describes the **final state** rather than the history.

### Final state

**`@tanstack/openai-base`** (renamed from the experimental `@tanstack/ai-openai-compatible` / `@tanstack/openai-compatible` names that were tried mid-PR) is now a thin shim over the `openai` SDK:

- Owns the AG-UI lifecycle pipeline for both Chat Completions and Responses APIs (~1k LOC of stream accumulation, partial-JSON tool-call buffering, RUN_ERROR taxonomy, structured-output coercion).
- Imports wire-format types from `openai/resources/*` directly — no vendored types.
- Constructor takes a pre-built `OpenAI` client; the base calls `client.chat.completions.create` / `client.responses.create` itself. **No more abstract `callChatCompletion*` / `callResponse*` hooks.**
- Subclasses (`ai-openai`, `ai-grok`, `ai-groq`) just construct the SDK with their provider-specific `baseURL` + headers and pass it to `super`.

**`ai-groq`** is migrated off the `groq-sdk` package. It now uses the OpenAI SDK pointed at `https://api.groq.com/openai/v1` (same pattern as `ai-grok` against xAI). Groq-specific quirks preserved via overridable hooks:
- `processStreamChunks` promotes `chunk.x_groq.usage` → `chunk.usage` so the base's RUN_FINISHED accounting works.
- `makeStructuredOutputCompatible` applies Groq's strict-mode schema quirks.

**`ai-openrouter`** is **fully decoupled** from `openai-base`. The earlier pass migrated it onto the base via SDK-call hooks + a snake_case ↔ camelCase round-trip; the final pass undoes that because OpenRouter's SDK shape was always native camelCase and the round-trip was pure friction:
- Two standalone classes (`OpenRouterTextAdapter`, `OpenRouterResponsesTextAdapter`) extend `BaseTextAdapter` directly.
- Stream processors duplicated locally and rewritten to read OpenRouter's camelCase types natively.
- ~300 LOC of `toOpenRouterRequest` / `toChatCompletion` / `adaptOpenRouterStreamChunks` / `toSnakeResponseResult` / etc. deleted.
- No `@tanstack/openai-base`, no `openai` SDK dep — only `@openrouter/sdk`, `@tanstack/ai`, `@tanstack/ai-utils`.

**New: OpenRouter Responses (beta) adapter** (`openRouterResponsesText`, `createOpenRouterResponsesText`). OpenRouter's `/v1/responses` endpoint fans out to Anthropic Claude, Google Gemini, etc. — the adapter exposes that surface alongside the existing chat-completions adapter.

**Other improvements**:
- Summarize adapters across providers unified on the chat-stream wrapper (`chat-stream-summarize`).
- `@tanstack/ai` normalizes abort-shaped errors (`AbortError`, `APIUserAbortError`, `RequestAbortedError`) to a stable `{ message: 'Request aborted', code: 'aborted' }` payload in `toRunErrorPayload` so consumers can discriminate user-cancellation from other failures.
- StreamChunk emissions across adapters now use `satisfies StreamChunk` instead of `asChunk` casts — drift surfaces at compile time.

### Bug fixes captured along the way

`openai-base` (the shared pipeline):
- `structuredOutput` throws a distinct `"response contained no content"` error rather than letting empty content cascade into a misleading JSON-parse error.
- Post-loop tool-args drain block now logs malformed JSON via `logger.errors`, matching the in-loop `finish_reason` path so truncated streams emitting partial tool args are debuggable instead of silently invoking the tool with `{}`.
- Stop processing chunks after a top-level error event so in-flight `TEXT_MESSAGE_CONTENT` / `TOOL_CALL_*` events don't leak past the terminal `RUN_ERROR`.
- `Responses.structuredOutput` routes through the `transformStructuredOutput` hook so subclasses that opt out of null-stripping (OpenRouter) don't have to fork the whole method.

`ai-openrouter`:
- `stream_options.include_usage` correctly camelCased to `includeUsage` so streaming `RUN_FINISHED.usage` is populated (was silently dropped by the SDK Zod schema).
- `chunk.error.code` stringified on mid-stream errors so provider error codes (401, 429, 500, …) survive the `toRunErrorPayload` narrow.
- Assistant `toolCalls[].function.arguments` stringified to match the SDK's `string` contract.
- `convertMessage` mirrors the base's fail-loud guards (throws on empty user content and unsupported content parts) instead of silently sending empty paid requests.
- Image data URIs default to `application/octet-stream` when the source has no MIME type (was producing invalid `data:undefined;base64,...` URIs).
- Audio URLs route to text fallback on chat-completions (the wire format has no URL variant for `input_audio`); inline document data on chat-completions throws (no `input_file` shape there) — both with explicit guidance to use the Responses adapter.
- Array-shaped tool message content has its text extracted rather than JSON-stringified (was feeding the literal ContentPart shape back to the model).
- Speakeasy's `{ raw, type: 'UNKNOWN', isUnknown: true }` discriminated-union fallback events pass through verbatim — Responses streams from upstreams that omit `sequence_number` / `logprobs` fields no longer get dropped.

`ai-groq`:
- `ChatCompletionNamedToolChoice` shape corrected.
- Spurious `timestamp` field removed from `processStreamChunks` override.
- `pendingMockCreate` reset between tests to prevent cross-test pollution.

## ✅ Checklist

## 🚀 Release Impact

Changeset: `.changeset/decouple-openrouter-collapse-openai-base.md` (covers `@tanstack/openai-base` minor + `ai-openai` / `ai-grok` / `ai-groq` / `ai-openrouter` patches).


tombeckenham:
### Follow-up: complete the openai-SDK decouple at the type level (commit e30a3ca8)

Section §4 of the description said `@tanstack/ai-openai-compatible` lists `openai` "only as an _optional_ peer." After re-checking with `pnpm knip` (which surfaces `Referenced optional peerDependencies`), that was still a half-measure: the runtime was decoupled but the **type surface** still leaked `openai` everywhere — abstract method signatures (`OpenAI_SDK.Chat.Completions.*` / `OpenAI_SDK.Responses.*`), all 12 tool-config aliases, and the public re-exports in `src/index.ts`. Net effect: installing `@tanstack/ai-openrouter` still pulled `openai` into the user's `node_modules` because their TypeScript needed it to resolve types coming through the upstream contract.

This commit finishes the decouple:

- Hand-wrote minimal wire-format types in `packages/typescript/ai-openai-compatible/src/types/{chat-completions,responses,tools}.ts` — only the fields the base actually reads/writes, structurally compatible with the openai SDK's richer shapes.
- Swapped every `import type ... from 'openai'` in the base adapters, tool converters, and 12 tool files to use the local types. Public re-exports in `src/index.ts` now source from `./types/*`.
- Dropped `openai` from `@tanstack/ai-openai-compatible`'s `peerDependencies`, `peerDependenciesMeta`, and `devDependencies` (no test file imported it).
- Dropped `openai` from `@tanstack/ai-openrouter`'s `devDependencies` (its sole purpose was satisfying the leaked type contract — `@openrouter/sdk` doesn't depend on it at runtime).
- Updated `@tanstack/ai-openai`'s text adapter overrides (`callResponse`, `callResponseStream`, `mapOptionsToRequest`) to use the local protocol types and cast at the SDK boundary inside the body — the only way to keep variance compat with the now-local base while still calling `this.client.responses.create()`.

#### Verified
- `pnpm test:types` across ai-openai-compatible / ai-openai / ai-groq / ai-grok / ai-openrouter — clean
- `pnpm test:lib` — 351 tests pass across the chain
- `pnpm knip` — exit 0, the previous `Referenced optional peerDependencies (1) openai` warning is gone
- `pnpm test:sherif` — no issues
- `pnpm test:build` (publint --strict) — passes for all 5 packages
- `grep "from ['\"]openai" packages/typescript/ai-openai-compatible/dist packages/typescript/ai-openrouter/dist` — zero matches in both

**Net effect:** end-users installing `@tanstack/ai-openrouter` (or any future protocol-compatible adapter that doesn't itself call the openai SDK at runtime) no longer pull `openai` into their dependency tree at all — not as a transitive dep, not as a peer warning, not as anything. The only packages that still list `openai` are `ai-openai` / `ai-groq` / `ai-grok`, which legitimately instantiate the OpenAI SDK as their HTTP client.

A follow-up changeset for this is still TODO (it's a notable behaviour improvement for `ai-openrouter` consumers worth calling out in release notes).

tombeckenham:
Heads-up: the latest push reverts the package name back to **`@tanstack/openai-base`**.

Earlier in this PR we tried two intermediate names (`@tanstack/openai-compatible`, then `@tanstack/ai-openai-compatible`) to signal the multi-vendor protocol surface. Now that `ai-openrouter` has been **decoupled entirely** (it extends `BaseTextAdapter` directly with its own duplicated stream processors), the only remaining consumers — `ai-openai`, `ai-grok`, `ai-groq` — all back onto the `openai` SDK with a different `baseURL`. "openai-base" describes that role accurately and matches the package's original name.

Imports for downstream consumers:

```diff
- import { OpenAICompatibleChatCompletionsTextAdapter } from '@tanstack/ai-openai-compatible'
+ import { OpenAIBaseChatCompletionsTextAdapter } from '@tanstack/openai-base'
```

`@tanstack/ai-openai-compatible@0.2.x` and `@tanstack/openai-compatible@*` remain published for any pinned-lockfile consumers but will receive no further updates.

packages/typescript/ai/src/types.ts
```diff
@@ -830,7 +830,13 @@ export interface RunFinishedEvent extends AGUIRunFinishedEvent {
   /** Model identifier for multi-model support */
   model?: string
   /** Why the generation stopped */
-  finishReason?: 'stop' | 'length' | 'content_filter' | 'tool_calls' | null
+  finishReason?:
+    | 'stop'
+    | 'length'
+    | 'content_filter'
+    | 'tool_calls'
+    | 'function_call'
```
AlemTuzlak: why is this added here?

packages/typescript/ai-openai/src/adapters/transcription.ts
```diff
@@ -72,32 +175,6 @@ export function createOpenaiTranscription<
...
- * // Automatically uses OPENAI_API_KEY from environment
- * const adapter = openaiTranscription('whisper-1');
- *
- * const result = await generateTranscription({
- *   adapter,
- *   audio: audioFile
- * });
- *
- * console.log(result.text)
- * ```
```
AlemTuzlak, line 99: why remove the jsdoc

packages/typescript/ai-openai/src/adapters/transcription.ts
```diff
@@ -10,58 +19,152 @@ import type { OpenAIClientConfig } from '../utils/client'
...
- * ```typescript
- * const adapter = createOpenaiTranscription('whisper-1', "sk-...");
- *
- * const result = await generateTranscription({
- *   adapter,
- *   audio: audioFile,
- *   language: 'en'
- * });
- * ```
- */
```
AlemTuzlak, line 64: bring back jsdoc

packages/typescript/ai-openrouter/src/adapters/responses-text.ts
```diff
@@ -0,0 +1,1500 @@
...
+      ...(options.metadata !== undefined && { metadata: options.metadata }),
+      ...(options.systemPrompts &&
+        options.systemPrompts.length > 0 && {
+          instructions: options.systemPrompts.join('\n'),
+        }),
+      input: input as ResponsesRequest['input'],
+      ...(tools &&
+        tools.length > 0 && {
+          tools: tools as ResponsesRequest['tools'],
+        }),
```
AlemTuzlak: there should be no casts here

packages/typescript/ai-openrouter/src/adapters/responses-text.ts
```diff
@@ -0,0 +1,1500 @@
...
+
+        // `output_item.done` is the last point at which a function_call's
+        // name is guaranteed to be on the wire.
+        if (chunk.type === 'response.output_item.done') {
+          const item = chunk.item as {
+            type: string
+            id?: string
+            name?: string
+            arguments?: string
+          }
```
AlemTuzlak: i don't like casting chunks it's a footgun waiting to happen

packages/typescript/ai-openrouter/src/adapters/responses-text.ts
```diff
@@ -0,0 +1,1500 @@
...
+          yield doneChunk
+          if (doneChunk.type === 'RUN_ERROR') {
+            runFinishedEmitted = true
+            return
+          }
+        }
+
+        // handle output_item.added to capture function call metadata (name)
+        if (chunk.type === 'response.output_item.added') {
+          const item = chunk.item as {
```
AlemTuzlak: don't cast

packages/typescript/ai-openrouter/package.json
```diff
@@ -53,6 +53,7 @@
     "zod": "^4.2.0"
   },
   "peerDependencies": {
-    "@tanstack/ai": "workspace:^"
+    "@tanstack/ai": "workspace:^",
+    "zod": "^4.0.0"
```
AlemTuzlak: why is zod a peer dep

packages/typescript/ai-openrouter/src/adapters/responses-text.ts
tombeckenham: Initial fix in 44db9257 for the `output_item.{added,done}` handlers (narrowing via `OutputItems`). Extended in f473e441 to the remaining sibling sites — `content_part.{added,done}` now narrows via `ContentPartAddedEventPart`, and `response.{created,in_progress,incomplete,failed}` reads directly off `Partial<OpenResponsesResult>`. `NormalizedStreamEvent.{item,response,part}` carry the SDK discriminated unions directly.

PR #561, opened 2026-05-15
fix(ai): make chat({ stream: false }) actually non-streaming on the wire

### LINKED ISSUE #557: chat({ stream: false }) still streams on the wire (uses runStreamingText + streamToText)
### Summary

Calling `chat({ stream: false })` does not actually send `stream: false` to the provider. Internally it calls `runStreamingText` and concatenates the SSE stream into a string via `streamToText`. The wire request still has `Accept: text/event-stream` and `"stream": true` in the body. The only code path that sends a wire-level non-streaming request is `chat({ outputSchema })` → `runAgenticStructuredOutput` → `adapter.structuredOutput`.

### Reproduction

```ts
import { openRouterText } from '@tanstack/ai-openrouter'

const adapter = openRouterText('x-ai/grok-4.3', { /* … */ })
const text = await chat({
  adapter,
  messages: [{ role: 'user', content: 'hello' }],
  stream: false,
  modelOptions: {
    responseFormat: { type: 'json_schema', jsonSchema: { /* … */ } },
  },
})
```

Wire capture: `Accept: text/event-stream`, body `"stream": true`. OpenRouter responds with SSE, SDK concatenates.

### Root cause (current `main`)

`packages/typescript/ai/src/activities/chat/index.ts:1609-1631` (dispatch):

```ts
if (outputSchema) return runAgenticStructuredOutput(options)
if (stream === false) return runNonStreamingText(options)
return runStreamingText(options)
```

`packages/typescript/ai/src/activities/chat/index.ts:1666-1675` (the offender):

```ts
function runNonStreamingText(options): Promise<string> {
  const stream = runStreamingText(options)
  return streamToText(stream)
}
```

`runStreamingText` → `TextEngine.streamModelResponse` → `adapter.chatStream(...)`. The OpenRouter adapter's `chatStream` hardcodes `stream: true` (`packages/typescript/ai-openrouter/src/adapters/text.ts:131`). Its `structuredOutput` is the only place that sends `stream: false` (`packages/typescript/ai-openrouter/src/adapters/text.ts:214`).

The adapter interface (`packages/typescript/ai/src/activities/chat/adapter.ts:59-120`) only defines `chatStream`, `structuredOutput`, and optional `structuredOutputStream` — there is no non-streaming `chat()` method on adapters today.

### Why it matters

Reasoning models under concurrent load (Grok 4.3 via OpenRouter, in our case) can take 30s+ of pure reasoning before any content emission. Observed wire behavior with the same prompt × 6 parallel calls:

- streaming, no proxy: 25–52s wall-clock each, all clean
- non-streaming (true wire-level), no proxy: 22–41s wall-clock each, all clean
- streaming through a proxy with a 30s socket idle timeout: one of the six **truncates mid-stream** because OpenRouter sends nothing for >30s during the reasoning phase

If `chat({ stream: false })` actually sent `stream: false`, that proxy-idle-timeout class of bug would not apply and fixtures/replay paths would be a single JSON body.

### Proposed fix

1. Add a non-streaming method to the adapter interface alongside `chatStream` / `structuredOutput`:
   ```ts
   chat(options: ChatStreamOptions): Promise<{
     content: string
     reasoning?: string
     toolCalls?: …
     usage?: …
   }>
   ```
2. OpenRouter implements it as a single `this.client.chat.send({ chatRequest: { …, stream: false } })` returning `result.choices[0].message.content` — mirroring `structuredOutput` minus the schema enforcement.
3. Rewire `runNonStreamingText` to call `adapter.chat(...)` directly instead of `runStreamingText` + `streamToText`.
4. For adapters that don't implement `chat()`, fall back to current behavior and emit a one-time warning so users know they're not getting wire-level non-streaming.

Happy to send a PR if the shape sounds right.

### Environment

- `@tanstack/ai`: 0.14.0
- `@tanstack/ai-openrouter`: 0.8.2
- Node 24.x (Bun 1.x)
- Provider: `x-ai/grok-4.3` via OpenRouter

tombeckenham:
Closes #557.

## Summary

- Adds optional `TextAdapter.chatNonStreaming()` returning `Promise<NonStreamingChatResult>` (`{ content, reasoning?, toolCalls?, finishReason?, usage? }`). When implemented, `chat({ stream: false })` routes through it for a true wire-level non-streaming request; when omitted, the legacy stream-then-concatenate fallback applies — out-of-tree adapters keep working unchanged.
- Implements `chatNonStreaming()` on all in-tree adapters: openai-base chat-completions + responses (covers openai, grok, groq), openrouter (chat-completions + responses), anthropic, gemini, ollama.
- Rewires `runNonStreamingText` as a wire-level non-streaming agent loop: call adapter → execute returned tool calls server-side → append results → repeat until `finishReason !== 'tool_calls'` or `agentLoopStrategy` halts. Approval-required and pure client-side tools throw an explicit error directing callers to streaming, since they can't round-trip through `Promise<string>`.

## Why

Previously `chat({ stream: false })` invoked `runStreamingText()` and concatenated the SSE response via `streamToText`. The wire request still carried `Accept: text/event-stream` and `"stream": true` in the body — only the SDK return type changed (`Promise<string>` instead of `AsyncIterable<StreamChunk>`). On reasoning models with long pre-content thinking phases (Grok 4.3 via OpenRouter was the original repro), proxies with sub-30s socket-idle timeouts could cut off the stream mid-flight even though a real non-streaming POST would complete fine.

## Test plan

- [x] Unit: 5 new tests covering wire-level dispatch, multi-turn tool loop, `agentLoopStrategy` cap, approval-tool error, and out-of-tree fallback. Existing 757 tests still pass (762 total).
- [x] E2E: rewired `api.chat.ts` so `feature === 'one-shot-text'` actually calls `chat({ stream: false })`. The existing `one-shot-text.spec.ts` now runs end-to-end against the new path across all 8 providers (openai, anthropic, gemini, ollama, groq, grok, openrouter, openrouter-responses). Updated the aimock fixture to include `systemFingerprint: null` so OpenRouter SDK's strict zod validation accepts the response. Full e2e suite (190 tests) passes.
- [x] `pnpm test:lib`, `pnpm test:types`, `pnpm test:eslint`, `pnpm build`: all clean.

## Notes

- Marked draft per author's request — wants more review before merging.
- Middleware is intentionally not invoked on the non-streaming path, mirroring the silence of the existing `adapter.structuredOutput` call inside `runAgenticStructuredOutput`.
- The `Promise<string>` return value remains content-only — reasoning is captured on the internal `NonStreamingChatResult` for future surfaces but is not part of the public contract.

### DIFF
--- a/.changeset/stream-false-actually-non-streaming.md
+++ b/.changeset/stream-false-actually-non-streaming.md
@@ -1,0 +1,27 @@
+---
+'@tanstack/ai': minor
+'@tanstack/openai-base': minor
+'@tanstack/ai-openrouter': minor
+'@tanstack/ai-anthropic': minor
+'@tanstack/ai-gemini': minor
+'@tanstack/ai-ollama': minor
+'@tanstack/ai-openai': patch
+'@tanstack/ai-grok': patch
+'@tanstack/ai-groq': patch
+---
+
+Make `chat({ stream: false })` actually non-streaming on the wire.
+
+Previously, `chat({ stream: false })` ran `runStreamingText()` and concatenated the SSE response via `streamToText`. The wire request still carried `Accept: text/event-stream` and `"stream": true` in the body — the only practical effect of `stream: false` was that the SDK returned `Promise<string>` instead of `AsyncIterable<StreamChunk>`. Reasoning models with long pre-content thinking phases (Grok 4.3 via OpenRouter was the original repro) could be cut off by sub-30s socket-idle timeouts in proxies along the path.
+
+Now `chat({ stream: false })` calls a new optional adapter method `chatNonStreaming()` that sends `stream: false` directly to the provider and returns a single JSON response. When tools are involved, the dispatch layer drives a wire-level non-streaming agent loop: call `chatNonStreaming`, execute any returned tool calls, append the results, repeat until the model stops requesting tools or the agent-loop strategy halts.
+
+**Adapter API addition.** `TextAdapter.chatNonStreaming(options)` is added as an optional method returning `Promise<NonStreamingChatResult>` (`{ content, reasoning?, toolCalls?, finishReason?, usage? }`). All in-tree adapters implement it:
+
+- `@tanstack/openai-base` — `OpenAIBaseChatCompletionsTextAdapter` and `OpenAIBaseResponsesTextAdapter`. Covers `@tanstack/ai-openai`, `@tanstack/ai-grok`, `@tanstack/ai-groq` automatically.
+- `@tanstack/ai-openrouter` — both `OpenRouterTextAdapter` and the Responses variant.
+- `@tanstack/ai-anthropic`, `@tanstack/ai-gemini`, `@tanstack/ai-ollama`.
+
+Adapters that don't override `chatNonStreaming` keep the legacy stream-then-concatenate behaviour — out-of-tree adapters continue working unchanged but won't get the wire-level fix until they opt in by implementing the method.
+
+**Behaviour notes:** approval-required tools and pure client-side tools cannot round-trip through a `Promise<string>`; the loop throws an explicit error directing callers to `stream: true` for those flows. Middleware is intentionally not invoked on the non-streaming path, matching the silence of the existing `adapter.structuredOutput` call inside `runAgenticStructuredOutput`. The `Promise<string>` return value remains content-only — reasoning is captured on the internal result for future surfaces but is not part of the public return type.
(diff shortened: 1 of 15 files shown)


tombeckenham:
I've looked into this a lot more. There are a tonne of restrictions when running non-streaming ai calls - particularly with anthropic. Given the amount of duplication we'd have to introduce to properly make non-streaming on the wire work, I don't think this is worth it @AlemTuzlak and @jherr . I'm closing this one as a not planned

PR #572, opened 2026-05-18
chore: monorepo setup audit (build, tsconfig, types, docs)

AlemTuzlak:
## Summary

End-to-end setup audit of the monorepo, focused on making the repo easier to contribute to without changing user-facing runtime behaviour (one intentional removal, called out below).

Seven commits, grouped by theme:

1. **`chore(monorepo)`** — `testing/**` excluded from build/CI scripts (was only excluding `examples/**`); defensive `ignore` list of every in-workspace private package in `.changeset/config.json`; vite bumped `^7.2.7 → ^7.3.3` across every declaring package.json to resolve a transitive version skew; `engines: { node: ">=24", pnpm: ">=11.0.0" }` on the root.
2. **`chore(tsconfig)`** — new `tsconfig.base.json` at the root holding the shared `compilerOptions`; every package tsconfig migrated to extend it with consistent `include: ["src", "tests"]` and only package-unique overrides. Tests are now included in `tsc` everywhere (previously several packages silently excluded them).
3. **`refactor(ai-gemini)`** — `GeminiThinkingOptions` and `GeminiThinkingAdvancedOptions` merged into a single interface with all fields optional. The old shape's intersection couldn't cleanly express the runtime contract.
4. **`refactor(ai-anthropic)`** — `'system'` removed from `validKeys`; `modelOptions.system` is no longer accepted. System prompts now flow through the `systemPrompts` public API as designed. See the note below.
5. **`fix(ai)`** — re-export `Logger` from the `@tanstack/ai/adapter-internals` subpath (was an `export interface` not re-exported on the public path).
6. **`test`** — fix the ~200 pre-existing type errors that the typecheck standardization surfaced (stale model names, `EventType` enum drift, missing required fields on result mocks, etc.) and clean up unsafe casts (10 `as any` + 8 `as unknown as` removed, 5 remaining at trust boundaries with justification comments).
7. **`docs(contributing)`** — `CONTRIBUTING.md`, `.editorconfig`, `.vscode/extensions.json`. Fixes the PR template's broken link to a previously-missing CONTRIBUTING.md.

## Behaviour change to flag (commit 4)

`modelOptions.system` is no longer accepted by the Anthropic adapter. Anyone constructing system prompts that way (via a cast — the public type never exposed it) needs to move to `systemPrompts`.

**Regression to track**: `systemPrompts: Array<string>` doesn't currently accept `cache_control` for prompt-caching. A follow-up will extend `systemPrompts` to `Array<string | { content, metadata }>` with provider-typed metadata. Design doc drafted separately.

## Changesets

Not added — maintainers may want different bump levels. My read:
- `@tanstack/ai-anthropic` — at least a minor (silent removal of an internal-but-accessible-via-cast escape hatch). Could argue major.
- `@tanstack/ai-gemini` — patch (the public type loosens; nothing tightens).
- `@tanstack/ai` — patch (additive `Logger` re-export).

## Test plan

- [ ] `pnpm test:ci` is green on this branch (verified locally: 33 projects × 8 targets all pass).
- [ ] CI on this PR matches the local result.
- [ ] Review the Anthropic system-prompt change — is removing the cast escape hatch what we want, or do you prefer a different shape for the follow-up?
- [ ] Spot-check `CONTRIBUTING.md` for accuracy on commands and conventions.
- [ ] Decide on changeset bump levels and add them.

## Summary by CodeRabbit

* **Documentation**
  * Added a comprehensive CONTRIBUTING guide
  * Removed outdated Node.js "Requirements" from READMEs

* **Bug Fixes**
  * Clarified Anthropic system-prompt handling (use systemPrompts; unknown modelOptions keys are now logged)
  * Simplified Gemini thinking-related options and typing

* **Chores**
  * Bumped dev tooling versions across packages/examples
  * Consolidated TypeScript configs to a shared base
  * Added EditorConfig and VS Code extension recommendations
  * Added package manager engine constraint and updated changeset ignore list

[Review Change Stack](https://app.coderabbit.ai/change-stack/TanStack/ai/pull/572?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


crutchcorn (approved):
Only reviewed `.changeset/config` and `scripts/` changes for security needs, nothing else. From that perspective LGTM

PR #575, opened 2026-05-18
feat(ai): systemPrompts accept { content, metadata } for per-provider metadata

AlemTuzlak:
> **Base:** stacked on top of #572. Merge that first.

## Summary

Restores the ability to attach Anthropic `cache_control` to system prompts — the regression introduced by removing the `modelOptions.system` escape hatch in #572 — via a typed, cross-provider mechanism.

`chat({ systemPrompts })` now accepts either a plain string (existing shape — fully backward compatible) or `{ content, metadata }`. The structured form lets providers attach typed metadata per prompt:

```ts
import { anthropicText, type AnthropicSystemPromptMetadata } from '@tanstack/ai-anthropic'

chat({
  adapter: anthropicText(),
  model: 'claude-sonnet-4-6',
  systemPrompts: [
    {
      content: 'Stable instructions — cache me.',
      metadata: { cache_control: { type: 'ephemeral' } } satisfies AnthropicSystemPromptMetadata,
    },
    'Volatile per-request instruction.',
  ],
})
```

## Design

- `@tanstack/ai` exports `SystemPrompt<TMetadata = unknown>`, `NormalizedSystemPrompt`, and `normalizeSystemPrompts()`. The chat layer carries the wide shape through middleware; each adapter calls the helper at the top of its option-mapping pipeline so the rest of its code only sees `{ content, metadata }`.
- The object form is **portable across providers**. Adapters that don't recognise a metadata field silently drop it; OpenAI / Gemini / Ollama / OpenRouter / openai-base all just join `.content` for their respective `instructions` / `system` / `systemInstruction` fields.
- The Anthropic adapter (the only one that consumes metadata today) reads `metadata.cache_control` and attaches it to the corresponding `TextBlockParam`. The new `AnthropicSystemPromptMetadata` interface is exported from `@tanstack/ai-anthropic`.

## Plumbing

- `TextOptions.systemPrompts` widened from `Array<string>` to `Array<SystemPrompt>`.
- `ChatMiddlewareContext.systemPrompts` / `ChatMiddlewareConfig.systemPrompts` widened so middleware can read/mutate the wide shape.
- OpenTelemetry middleware extracts `.content` for span events; per-prompt metadata is dropped from spans (not useful for observability).
- `@tanstack/ai-event-client` mirrors the `SystemPrompt` shape locally (it intentionally avoids importing `@tanstack/ai` to prevent a circular dep) and projects metadata away on the devtools wire — devtools UI continues to receive `Array<string>`.

## Tests

- `ai-anthropic`: new test verifies `cache_control` flows from `systemPrompts[i].metadata` onto the outbound `TextBlockParam`, and plain-string entries still produce metadata-less blocks.
- `ai-openai`: new test verifies mixed string + object-form input produces the expected joined `instructions` and that provider-foreign metadata is silently dropped.

## Test plan

- [ ] Verified locally: `pnpm test:ci` is green on this branch (33 projects × 8 targets).
- [ ] CI on this PR matches.
- [ ] Spot-check the API ergonomics: `systemPrompts: ['x']` still works for every provider; `systemPrompts: [{ content: 'x' }]` works for every provider; Anthropic `metadata.cache_control` reaches the wire payload.

## Notes

- Changeset added (`@tanstack/ai` + `@tanstack/ai-anthropic` minor; downstream adapters patch).
- The full design document with open questions/tradeoffs is in `C:\Users\AlemTuzlak\.claude\projects\F--projects-tanstack-ai\design\system-prompts-metadata-2026-05-18.md` (outside the repo, not committed).

## Release Notes

* **New Features**
  * System prompts now accept structured objects with metadata in addition to plain strings. Pass `{ content: string, metadata }` for provider-specific configurations.
  * Anthropic adapter: per-prompt `cache_control` metadata is now supported.

* **Tests**
  * Added tests verifying system prompt metadata handling across all supported providers.

[Review Change Stack](https://app.coderabbit.ai/change-stack/TanStack/ai/pull/575?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


tombeckenham (commented):
# PR Review — `feat/system-prompts-metadata`

**Scope:** 25 files, +410/-46, reviewed against `chore/monorepo-setup-audit` (stacked PR). Type plumbing verified, all unit tests pass, `Array<string>` backward compatibility confirmed.

## Critical (1)

**C1 — `normalizeSystemPrompts` does zero runtime validation**
`packages/typescript/ai/src/system-prompts.ts:64-68`. The cast `p as NormalizedSystemPrompt<TMetadata>` trusts the TS boundary. `{ metadata: {…} }` (missing `content`) or `{ content: undefined as any }` passes through; every adapter then does `prompts.map((p) => p.content).join('\n')` and produces a literal `"undefined"` in the system prompt — no log, no throw, no telemetry. **Fix:** when the entry is object-form, validate `typeof p.content === 'string'`; throw with the offending index. This is a public API boundary.

## Important (6)

**I1 — No E2E coverage; CLAUDE.md mandates it**
`testing/e2e/` has zero diff. The repo policy requires E2E for behavior changes. `testing/e2e/src/routes/api.chat.ts` already wires `systemPrompts`; one Anthropic fixture passing `metadata.cache_control` would cover the wire path.

**I2 — Middleware-context shape widened but only types changed, not behavior**
`ChatMiddlewareContext.systemPrompts` is now `Array<SystemPrompt>` but every runtime assertion in `packages/typescript/ai/tests/middleware.test.ts` still uses plain strings. A middleware doing `config.systemPrompts.join('\n')` would now produce `'[object Object]'` and no test catches it.

**I3 — Per-adapter mapping has no direct coverage for Gemini, Ollama, OpenRouter**
Anthropic+OpenAI tests cover the helper indirectly, but per-adapter glue diverges: Gemini joins into `systemInstruction` (string), Ollama uses spread-or-omit, OpenRouter `text.ts:1117-1122` pushes a positional `{ role: 'system' }` message (different shape entirely), OpenRouter `responses-text.ts` uses yet another SDK. None of these have a direct test.

**I4 — OTel middleware drops `cache_control` from observability**
`packages/typescript/ai/src/middlewares/otel.ts:374-403` flattens to strings; the fact that `cache_control` was attached is invisible. For Anthropic prompt-caching users this is the one attribute that explains cache hit/miss. **Fix:** attach metadata as a `tanstack.ai.system_prompt.metadata` JSON span attribute when `captureContent: true`.

**I5 — Doc/changeset wording on `metadata: never` is misleading**
Changeset and JSDoc in `system-prompts.ts:9-12`, `types.ts:699-700` claim adapters "silently ignore" foreign metadata. With `TSystemPromptMetadata = never` default they *reject* the field at compile time (the adjacent `chat/index.ts:112-113` doc correctly says "reject the field entirely"). Also: `{ content: 'x', metadata: undefined }` *does* typecheck — the "passing it errors at compile time" claim should be tightened to "cannot carry a meaningful value." The `@example` showing `satisfies AnthropicSystemPromptMetadata` is contradicted by the test comment saying no `satisfies` is needed.

**I6 — `DevtoolsSystemPrompt` mirror has no structural-equality guard**
`packages/typescript/ai-event-client/src/devtools-middleware.ts:15-20` is intentionally duplicated to avoid a circular import. If `SystemPrompt` gains a third variant, the projection `typeof p === 'string' ? p : p.content` silently emits `undefined`. Add a `const _check: SystemPrompt = null as unknown as DevtoolsSystemPrompt` assertion in a test to force breakage if shapes diverge.

## Suggestions (5)

- **S1** — `NormalizedSystemPrompt` is incidental cruft; structurally `Extract<SystemPrompt<T>, object>` with a new name. Consider deleting and inlining the `Extract`.
- **S2** — Add `Readonly<…>` to both `SystemPrompt` object arm and `NormalizedSystemPrompt`. Adapters shouldn't mutate caller-supplied data; cheap, prevents fan-out bugs.
- **S3** — 7 positional generics on `TextAdapter`/`BaseTextAdapter` is approaching unmanageable. Don't refactor in this PR, but the next addition should consolidate to a single `TConfig` object generic.
- **S4** — `systemPromptMetadata: undefined as never` in mocks is awkward. Introduce a tiny `phantom<T>(): T` helper so the cast lives in one place, and/or add a one-line justifying comment at `tests/test-utils.ts:175`.
- **S5** — Devtools wire could ship `hasMetadata: boolean` per prompt now to allow a future UI badge without a wire-format break.

## Strengths

- Clean type chain `~types['systemPromptMetadata']` → `SystemPrompt<TMetadata>` → call-site narrowing works as advertised.
- Backward compat verified: `systemPrompts: ['foo']` typechecks and behaves identically on every adapter.
- Anthropic test covers the regression most likely to break (`TextBlockParam` shape preservation across the rewrite).
- IIFE pattern across 6 adapters is consistent and readable; the structural mirror in `ai-event-client` is the right call given the circular-import constraint.
- Most new doc comments are genuine WHY/non-obvious (OTel rationale, devtools wire-stability rationale, circular-dep mirror).
- Snake_case `cache_control` is correct for SDK pass-through.
- Changeset bump levels are correct.

## Recommended Action

1. **Before merge:** fix C1 (runtime validation), address I1 (E2E fixture), tighten I5 (doc/changeset wording).
2. **Strongly recommended:** add direct mapping tests for Gemini/Ollama/OpenRouter (I3) and one middleware-context test with the object form (I2).
3. **Follow-up acceptable:** I4 (OTel attribute), I6 (mirror guard), all suggestions.

tombeckenham (approved):
Good to go

PR #578, opened 2026-05-18
feat: per-turn schemas for structured-output chat (Phase 2)

tombeckenham:
## Summary

Phase 2 of the structured-output redesign. Stacks on #577 — opening against \`feat/use-chat-structured-output-parts\` so the diff is just the Phase 2 delta. Will rebase onto main once #577 lands.

\`useChat\` can now multiplex multiple structured-output schemas in one conversation, picking which the assistant should respond against on a per-turn basis. Each assistant message's structured-output part carries the chosen \`schemaName\` as a runtime discriminator; \`getStructuredPart(message, name)\` gives a type-safe view of the part with \`data\` and \`partial\` narrowed against the matching schema.

## What changed

- **New \`outputSchemas\` option** on \`useChat\`: \`Record<string, SchemaInput>\` map of named schemas.
- **Per-call schema picker:** \`sendMessage(content, { schema: 'recipe' })\`. Typed as \`keyof TSchemas\` when \`outputSchemas\` is set; uninvocable otherwise.
- **New \`getStructuredPart(message, name)\` helper** on the hook return — typed view of a turn's structured part where \`data\`/\`partial\` are narrowed against \`InferSchemaType<TSchemas[name]>\`. Only available when \`outputSchemas\` is configured.
- **Server-side \`chat({ schemaName })\` option** — propagates into the synthetic \`structured-output.start\` and the terminal \`structured-output.complete\` events. Pass it from your server route after looking up the schema by name from \`forwardedProps.schemaName\`.
- **Client processor** records \`schemaName\` from \`structured-output.start\` and persists it through delta updates and on \`complete\`, even when only the start event carries it.
- **New example \`/generations/structured-multi\`** in ts-react-chat: a two-schema conversation (recipe + weekly plan) with a per-message picker, demonstrating type-safe render branching on \`part.schemaName\`.

Single-schema \`outputSchema\` flows from PR #577 continue to work unchanged.

## Test plan

- [x] \`pnpm --filter @tanstack/ai test:lib\` — **810 passing** (2 new processor tests: schemaName persistence from start event, fallback when only start carries it)
- [x] \`pnpm --filter @tanstack/ai-client test:lib\` — 207 passing
- [x] \`pnpm --filter @tanstack/ai-react test:lib\` — **114 passing** (1 new: two-schema flow with getStructuredPart narrowing)
- [x] \`pnpm --filter @tanstack/ai test:eslint\` / \`ai-client\` / \`ai-react\` — clean
- [x] Types green across ai / ai-client / ai-react
- [x] Example builds clean
- [ ] Manual: \`pnpm --filter ts-react-chat dev\`, open \`/generations/structured-multi\`, pick \"Recipe\" and send \"weeknight pasta for two\", then switch to \"Weekly plan\" and send \"focused week for learning TypeScript\". Verify each turn renders against the correct schema. Needs \`OPENAI_API_KEY\`.

## Reviewer notes

- **Stacked on #577.** Diff is Phase-2-only; the base branch carries the foundational MessagePart design.
- **Decision: schema-by-name vs schema-by-value.** Went with the named-map approach because (a) it lets the typed view (\`getStructuredPart\`) narrow against the schema without runtime serialization, and (b) the server has to be in on the schema list anyway for validation. Per-call ad-hoc schemas would require serializing JSON Schema over the wire and was deemed not worth the complexity for this iteration.
- **\`UIMessage<TTools, TSchemas>\` not parameterized.** I considered making the part typed at the message level so \`messages.map(...).find(p => p.type === 'structured-output' && p.schemaName === 'recipe')\` would narrow data automatically. Decided against it — too much ripple for the ergonomic gain when \`getStructuredPart\` covers the same use case with one extra call.

### DIFF
--- a/.changeset/structured-output-as-message-part.md
+++ b/.changeset/structured-output-as-message-part.md
@@ -1,0 +1,25 @@
+---
+'@tanstack/ai': minor
+'@tanstack/ai-client': minor
+'@tanstack/ai-react': minor
+---
+
+feat: structured-output as a typed MessagePart on each assistant UIMessage
+
+`useChat({ outputSchema })` previously kept a single hook-level `partial`/`final`
+slot, so multi-turn structured chats lost every prior turn's response as soon
+as a new one streamed in. Each assistant turn now carries its own typed
+`structured-output` MessagePart on the UIMessage it belongs to. History walks
+`messages` and finds the typed part on each turn; the hook-level `partial`
+and `final` are derived from the latest assistant message's part and continue
+to work as before.
+
+Server-side `chat({ outputSchema, stream: true })` emits a new
+`structured-output.start` CUSTOM event before the JSON deltas so the client
+processor can route them into the StructuredOutputPart instead of building a
+TextPart. The wire converter serializes the part's raw JSON back as assistant
+content, so multi-turn structured chats stay coherent (the LLM sees its own
+prior structured responses on follow-up turns).
+
+A new example route demonstrating this pattern is at
+`/generations/structured-chat` in the `ts-react-chat` example.
--- a/.changeset/structured-output-per-turn-schemas.md
+++ b/.changeset/structured-output-per-turn-schemas.md
@@ -1,0 +1,28 @@
+---
+'@tanstack/ai': minor
+'@tanstack/ai-client': minor
+'@tanstack/ai-react': minor
+---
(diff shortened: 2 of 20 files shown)


tombeckenham:
Not keen on this one. Closing

PR #651, opened 2026-05-27
feat: add OpenAI diarization support

8times4:
## 🎯 Changes

This change adds diarization support for OpenAI's gpt-4o-transcribe-diarize model, based on https://developers.openai.com/api/docs/guides/speech-to-text?lang=javascript

## ✅ Checklist

## 🚀 Release Impact

## Summary by CodeRabbit

* **New Features**
  * Added speaker diarization for OpenAI transcriptions with automatic speaker labeling and segment-level timestamps
  * New diarized JSON response format for structured, speaker-labeled transcripts
  * Transcription API now supports GPT-4o diarize model alongside existing models

* **Documentation**
  * Updated transcription docs, examples, and best practices with diarization usage, options, and constraints


tombeckenham:
Hi @8times4, thank you for this. Would you be able to create an e2e test for this using aimock? The tests are in the e2e test package. Ideally, adding a way to see the results on one of the ts-react-chat example pages would be great as well

tombeckenham:
### Code review

Found 3 issues:

1. No E2E test coverage added for the diarization feature/behavior change (new `gpt-4o-transcribe-diarize` model, `diarized_json` responseFormat, speaker-labeled `TranscriptionSegment`s, `chunking_strategy` + `known_speaker_*` options + validation). (CLAUDE.md says "Every feature, bug fix, or behavior change MUST include E2E test coverage." and "Add or update E2E tests — this is mandatory for any feature, bug fix, or behavior change"; see also the new-feature row in the E2E table and Pre-PR Quality Gate requiring `pnpm --filter @tanstack/ai-e2e test:e2e`. AGENTS.md and prior transcription PRs #409/#506 reviews establish the same convention: update feature-support.ts + test-matrix + fixture + spec.)

https://github.com/TanStack/ai/blob/05dfb5360e845db5abe6d3bbe926aefd78c6943c/packages/ai-openai/src/adapters/transcription.ts#L140-L150

2. `responseFormat` union literal duplicated (with added `| 'diarized_json'`) across three locations instead of extracting a shared type. (CLAUDE.md says "Always look for repeated code or if the function you are trying to implement is already in another file" and "Review code at the end to see if you can make it more concise and less repetitive".)

https://github.com/TanStack/ai/blob/05dfb5360e845db5abe6d3bbe926aefd78c6943c/packages/ai/src/types.ts#L1723-L1732

3. Validation guards in the newly added `validateDiarizationOptions` (and caller guard) are inconsistent with modelOptions conventions and incomplete: camelCase cast for `responseFormat` inside `modelOptions` (while spread + all other fields use snake_case `response_format`/`chunking_strategy`/`known_speaker_*`); `prompt` rejection and diarization-options guard only inspect top-level (not `modelOptions` paths); `chunking_strategy` diarize-only restriction does not check `modelOptions?.chunking_strategy`. This allows bypasses leading to late 400s instead of early errors. (CLAUDE.md says "Don't create fallback code. It hides problems. Just display errors to the user".)

https://github.com/TanStack/ai/blob/05dfb5360e845db5abe6d3bbe926aefd78c6943c/packages/ai-openai/src/adapters/transcription.ts#L339-L370

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

8times4:
Thanks for the review @tombeckenham, should be fixed now.

tombeckenham (changes requested):
Thanks for the thorough follow-up here — the E2E coverage across all three transports, the `openai-diarize` option on the example page with speaker-labeled segments, the shared `TranscriptionResponseFormat` extraction, and the much more complete `validateDiarizationOptions` all address my earlier review. The response-mode discriminant (`'diarized' | 'verbose' | 'plain'`) + request-plan refactor is a genuine improvement. 🙏

I'm requesting changes on one architectural point, plus the small cleanups it implies.

## Requested change — keep `diarized_json` out of the shared union

`diarized_json` was added to the **shared, cross-provider** `TranscriptionResponseFormat` (`packages/ai/src/types.ts:1712-1718`). The problem: only the OpenAI adapter reads the top-level `responseFormat`. The other transcription adapters drive diarization from a boolean instead and ignore `responseFormat` entirely:

- `packages/ai-elevenlabs/src/adapters/transcription.ts` → `modelOptions.diarize: boolean`
- `packages/ai-grok/src/adapters/transcription.ts` → `modelOptions.diarize: boolean`
- `packages/ai-fal/...` → no diarization support

So after this change, `responseFormat: 'diarized_json'` type-checks for **every** provider but is **silently ignored** by three of them. `diarized_json` isn't a portable format — it's OpenAI's wire-format literal for one model (`gpt-4o-transcribe-diarize`). The shared union should only advertise formats a generic caller can actually request.

Please **keep `diarized_json` in `OpenAITranscriptionResponseFormat` only** and revert the shared `TranscriptionResponseFormat` to the portable set (`'json' | 'text' | 'srt' | 'verbose_json' | 'vtt'`). Diarization on OpenAI is already driven through `modelOptions.response_format: 'diarized_json'` (that's what the example and E2E use), so nothing in the feature path is lost.

Two cleanups fall out of this and should land together:

1. `OpenAITranscriptionResponseFormat = TranscriptionResponseFormat | 'diarized_json'` (`packages/ai-openai/src/audio/transcription-provider-options.ts:4-6`) is currently **redundant** — `diarized_json` is already in the shared union, so it collapses to exactly `TranscriptionResponseFormat`. Once `diarized_json` moves out, this alias becomes a real extension again. ✅
2. `const topLevelResponseFormat = responseFormat as OpenAITranscriptionResponseFormat | undefined` (`packages/ai-openai/src/adapters/transcription.ts:253-255`) is a **no-op cast** today and becomes a safe widening after the change — drop the `as` and assign directly.

## Direction we want — a portable `diarize` on/off

For where this should go cross-provider: the portable concept is **diarization on/off**, not a format string. And the *output* is already normalized — `TranscriptionSegment.speaker?: string` exists in the shared type and ElevenLabs already populates it. So the clean cross-provider surface is a top-level `diarize?: boolean` on `TranscriptionOptions` that each adapter maps to its own mechanism (OpenAI → diarize model + `diarized_json`; ElevenLabs/Grok → `diarize: true`), with results unified via `segment.speaker`.

I don't want to balloon this PR's scope. Either is fine with me:
- add a top-level `diarize?: boolean` wired for **OpenAI only** in this PR (ElevenLabs/Grok wiring as a follow-up), or
- keep this PR OpenAI-scoped via `modelOptions` and I'll open a follow-up issue for the portable flag.

Let me know which you'd prefer. (Note for whoever does the portable work: Grok currently types its speaker as `number` while the shared `segment.speaker` is `string` — that needs reconciling.)

## Minor

- **Validation gap** (`packages/ai-openai/src/adapters/transcription.ts:458-467`): the matching-length check only fires when **both** `known_speaker_names` and `known_speaker_references` are present. A lone array (one provided without the other) passes local validation and defers to a late opaque 400 — the exact thing this validator exists to prevent. Please reject early when exactly one of the two is provided.
- **Dead branch** (`testing/e2e/src/lib/media-providers.ts:47`): `modelOptions?.diarize === true` is never set by any caller (the real switch is `response_format: 'diarized_json'`), and `diarize` isn't a valid OpenAI param. Please drop that clause so model selection is driven only by real fields.

Blocking: the shared-union scoping + the two cleanups it implies. The rest is quick polish. Thanks again!

tombeckenham (approved):
We got there. Thank you for this. Approving

8times4:
Thank you as well, Tom!

PR #654, opened 2026-05-29
feat(ai-openrouter): surface per-request cost on RUN_FINISHED

### LINKED ISSUE #468: feat(ai-openrouter): surface per-request cost on RUN_FINISHED

season179:
## Changes

Closes #468.

OpenRouter returns the actual per-request cost inline on the chat response. This surfaces it on the terminal `RUN_FINISHED` event as `usage.cost`, with OpenRouter's per-request breakdown under `usage.costDetails`. Both the `openRouterText` (Chat Completions) and `openRouterResponsesText` (Responses) adapters populate it.

`@tanstack/ai` gains a shared `UsageTotals` type with optional `cost` / `costDetails` fields. `RunFinishedEvent.usage`, the middleware `UsageInfo` (`onUsage`), and `FinishInfo.usage` (`onFinish`) all use it, so cost can be read without casts and on the middleware hooks. Cost is the value OpenRouter itself reports — not computed locally from token counts — so it accounts for routing, fallback providers, BYOK, and cached-token pricing. Cost-detail keys are normalized to camelCase so the SDK-parsed and raw-fallback paths stay consistent. The fields are optional and additive; adapters that don't report cost are unaffected.

This supersedes the earlier #469, which is closed. That approach cloned the SSE stream and reconstructed cost through a custom `CostStore`/HTTP-client hook. With `@openrouter/sdk` parsing `usage.cost` / `usage.cost_details` natively, this version is purely additive plumbing — no stream cloning, no provider-specific machinery.

**Test plan**
- Unit: `cost.test.ts` (the `extractUsageCost` helper — `cost === 0`, negatives, `null`, non-finite, snake/camel key normalization), cost suites in both adapter test files (including a raw/UNKNOWN `response.completed` fallback case), and `usage-cost-types.test.ts` locking the types.
- E2E: new `openrouter-cost.spec.ts` driving a deterministic OpenRouter cost mount end-to-end.
- `pnpm test:pr` green across 31 projects; full E2E suite green (224 passed) locally.

## ✅ Checklist

## Release Impact

## Summary by CodeRabbit

* **New Features**
  * OpenRouter per-request cost tracking now surfaces on run completion (cost + detailed breakdown) and is delivered to middleware/finish hooks via a shared usage shape.

* **Documentation**
  * Added cost-tracking docs and updated provider table with examples showing how to read cost and breakdowns.

* **Types**
  * Introduced shared UsageTotals and UsageCostBreakdown types to expose optional cost and breakdown fields.

* **Tests**
  * Added unit, adapter, integration, and e2e tests validating extraction, streaming emission, key casing, and end-to-end reporting.

[Review Change Stack](https://app.coderabbit.ai/change-stack/TanStack/ai/pull/654?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/packages/ai/src/types.ts
+++ b/packages/ai/src/types.ts
@@ -926,21 +926,19 @@ export interface RunStartedEvent extends AGUIRunStartedEvent {
 }
 
 /**
- * Provider-reported cost breakdown for a single request. All fields are optional;
- * an adapter only sets the keys its provider actually reports. Unknown breakdown
- * keys are dropped at extraction time so the consumer surface stays closed.
+ * Provider-reported cost breakdown for a single request, normalized onto a
+ * canonical shape so consumer code is portable across gateways. Each adapter's
+ * extractor maps its provider-specific wire keys (e.g. OpenRouter's
+ * `upstream_inference_prompt_cost`, `upstream_inference_input_cost`) onto these
+ * fields at runtime.
  */
-export interface UsageCostDetails {
+export interface UsageCostBreakdown {
   /** Total cost the gateway paid the upstream provider. */
-  upstreamInferenceCost?: number
-  /** Upstream cost for prompt tokens (Chat Completions naming). */
-  upstreamInferencePromptCost?: number
-  /** Upstream cost for completion tokens (Chat Completions naming). */
-  upstreamInferenceCompletionsCost?: number
-  /** Upstream cost for input tokens (Responses naming). */
-  upstreamInferenceInputCost?: number
-  /** Upstream cost for output tokens (Responses naming). */
-  upstreamInferenceOutputCost?: number
+  upstreamCost?: number
+  /** Upstream cost for input (prompt) tokens. */
+  upstreamInputCost?: number
+  /** Upstream cost for output (completion) tokens. */
+  upstreamOutputCost?: number
 }
 
 /**
@@ -957,7 +955,7 @@ export interface UsageTotals {
   /** Provider-reported cost for the request, when available. */
   cost?: number
   /** Provider-reported cost breakdown, when available. */
-  costDetails?: UsageCostDetails
+  costDetails?: UsageCostBreakdown
 }
 
 /**
--- a/packages/ai/src/types.ts
+++ b/packages/ai/src/types.ts
@@ -926,6 +926,27 @@ export interface RunStartedEvent extends AGUIRunStartedEvent {
 }
 
 /**
+ * Token usage totals for a run, optionally including provider-reported cost.
+ *
+ * `cost` and `costDetails` are populated only by adapters whose provider returns
+ * authoritative per-request cost (e.g. OpenRouter). They are absent for adapters
+ * that do not report cost, so consumers must treat them as optional.
+ *
+ * `costDetails` is intentionally a generic numeric map rather than a fixed shape,
+ * so provider-specific breakdown keys (e.g. `upstreamInferenceCost`, `cacheDiscount`)
+ * can be surfaced without baking provider-specific fields into this shared type.
+ */
+export interface UsageTotals {
+  promptTokens: number
+  completionTokens: number
+  totalTokens: number
+  /** Provider-reported cost for the request, when available. */
+  cost?: number
+  /** Provider-specific cost breakdown, when available. */
+  costDetails?: Record<string, number | null | undefined>
+}
+
+/**
  * Emitted when a run completes successfully.
  *
  * @ag-ui/core provides: `threadId`, `runId`, `result?`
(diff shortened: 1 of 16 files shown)


season179:
Thanks @coderabbitai 🐰 — addressed in 6ca2e898:

- **`global-setup.ts` (drain request body):** valid. Added `await drainBody(req)` in `openRouterCostMount` after the method/path guard, matching the four sibling mounts (ElevenLabs/Grok). No unread bytes left on the keep-alive socket now.

- **`api.openrouter-cost.ts` (import from `/adapters` subpath):** skipping this one — `@tanstack/ai-openrouter` doesn't expose an `/adapters` subpath (its `exports` map is only `.` and `./tools`), so `@tanstack/ai-openrouter/adapters` wouldn't resolve. `createOpenRouterText` is re-exported from the package root, and the sibling e2e routes (e.g. `api.openrouter-web-tools-wire.ts`) import it from the root too. The `/adapters` convention applies to packages like `@tanstack/ai-openai` that ship that subpath; `ai-openrouter` doesn't.

packages/ai/src/types.ts
```diff
@@ -925,6 +925,27 @@ export interface RunStartedEvent extends AGUIRunStartedEvent {
...
+ * can be surfaced without baking provider-specific fields into this shared type.
+ */
+export interface UsageTotals {
+  promptTokens: number
+  completionTokens: number
+  totalTokens: number
+  /** Provider-reported cost for the request, when available. */
+  cost?: number
+  /** Provider-specific cost breakdown, when available. */
+  costDetails?: Record<string, number | null | undefined>
```
tombeckenham: I'm going to give this the actual openrouter shape. If another provider supports cost details like this, we can look at the shape then. No one else does at this point

tombeckenham (commented):
I'm creating a type `UsageCostBreakdown` as a vendor independent cost breakdown type. I did a bit of research on how many others return cost data - no one we currently support - but thought I'd keep the shape generic and mould the openrouter inference cost data into it. it's almost the same. Does this work for you?

season179:
Yes, this works. Thank you.

tombeckenham (approved):
All good. Approving

PR #660, opened 2026-05-30
feat: move sampling options (temperature/topP/maxTokens) into modelOptions

AlemTuzlak:
## Summary

Moves sampling options — `temperature`, `topP`, `maxTokens` — **off the root** of `chat()` / `ai()` / `generate()` and into provider-native, fully-typed `modelOptions`. `modelOptions` is now the single sampling surface; there is no generic root-level mapping anymore. Each provider exposes its real SDK/API key names.

This continues and completes #499 by @harry-whorlow (which was built on the old `packages/typescript/` layout and could not be rebased onto the restructured `packages/` tree). **Supersedes #499.**

> **BREAKING** for a 0.x SDK. A provider-aware codemod migrates existing code automatically.

## Provider-native `modelOptions` keys

| Provider | temperature | topP | maxTokens |
|---|---|---|---|
| OpenAI (Responses) | `temperature` | `top_p` | `max_output_tokens` |
| Anthropic | `temperature` | `top_p` | `max_tokens` |
| Gemini | `temperature` | `topP` | `maxOutputTokens` |
| Grok | `temperature` | `top_p` | `max_tokens` |
| Groq | `temperature` | `top_p` | `max_completion_tokens` |
| Ollama | `options.temperature` | `options.top_p` | `options.num_predict` (nested) |
| OpenRouter (chat) | `temperature` | `topP` | `maxCompletionTokens` |
| OpenRouter (Responses) | `temperature` | `topP` | `maxOutputTokens` |

## What changed

- **Core (`@tanstack/ai`)**: removed `temperature`/`topP`/`maxTokens` from `TextOptions`, the public activity options, and `ChatMiddlewareConfig` (+ structured-output config). Removed all engine plumbing. Middleware now adjusts sampling via `config.modelOptions`. `metadata` is unchanged and stays at the root.
- **All 7 text providers** read sampling from typed `modelOptions` (native keys, verified against each provider's API). The provider-options types gained the sampling fields where missing (OpenAI, Anthropic, Gemini, Ollama; Grok/Groq/OpenRouter already had them).
- **Casts removed** on the sampling path (Anthropic, Ollama, OpenRouter). No new casts introduced.
- **Bugs fixed along the way**: a pre-existing Ollama double-nesting mismatch (typed nested `options` vs flat runtime spread); an OpenRouter `variant` field leaking into the request body; OTel `gen_ai.request.*` attributes now read a union of provider key spellings; `summarize({ maxLength })` now resolves the correct per-provider token key from the wrapped adapter (was about to become a silent no-op).
- **Codemod** `move-sampling-to-model-options` (jscodeshift): resolves the provider from the adapter factory and rewrites keys per provider (incl. Ollama nesting); reports + skips on conflicts or unresolvable adapters. 12 fixture cases.
- **Docs**: new migration guide (`docs/migration/sampling-options-to-model-options`), updated adapter pages, middleware `dynamic-temperature` example, API reference, and typed-options.
- **Agent skills**: `adapter-configuration`, `middleware`, `chat-experience` SKILL.md + reference files updated to the new surface.
- **Changeset**: `minor` across `@tanstack/ai` + all affected adapter packages.

## Migration

```bash
pnpm codemod:move-sampling-to-model-options "src/**/*.{ts,tsx}"
```

## Testing

- Per-adapter unit tests assert sampling reaches the wire under the correct native key (incl. Ollama no-double-nesting, Anthropic `max_tokens` default + no spurious dropped-key warning).
- Codemod fixture tests (12 cases). OTel + summarize regression tests.
- E2E suite verified green (already routes sampling via `modelOptions`).
- Full `pnpm test:pr` green: `test:sherif, test:knip, test:docs, test:eslint, test:lib, test:types, test:build, build` across 33 projects.

## Known follow-up (not blocking)

The `ts-code-mode-web` / `ts-react-chat` examples still pass a generic `maxTokens` on calls with dynamically-resolved adapters — the codemod can't infer a provider there, and they're excluded from CI (`examples/**`). They need per-example provider mapping in a follow-up.

## Summary by CodeRabbit

* **Breaking Changes**
  * Root-level sampling options (temperature, topP, maxTokens) no longer work on chat()/ai()/generate(); supply them via provider-native modelOptions (provider-specific key names; Ollama nests under modelOptions.options). metadata remains at the root.

* **New Features**
  * Added an automated codemod to migrate sampling options into modelOptions.

* **Documentation**
  * Updated adapters, middleware, migration guide, examples, and docs to reflect modelOptions usage and migration guidance.


tombeckenham (changes requested):
- pnpm vitest run move-sampling-to-model-options fails
- Ollama summarize silently drops the temperature: 0.3 default

packages/ai-code-mode/models-eval/run-eval.ts
```diff
@@ -210,6 +210,31 @@ function getTextAdapter(
...
+ * differs per provider. Ollama also keeps its larger context window via
+ * `num_ctx`.
+ */
+function maxTokensModelOptions(
+  provider: EvalProvider,
+  maxTokens: number,
+): Record<string, unknown> {
+  switch (provider) {
+    case 'ollama':
+      return { num_predict: maxTokens, num_ctx: 32768 }
```
tombeckenham: Flat `num_predict`/`num_ctx` are dropped: this result is passed as `modelOptions` (L818), but the Ollama adapter reads sampling only from `modelOptions.options`. Nest it:
```ts
return { options: { num_predict: maxTokens, num_ctx: 32768 } }
```

packages/ai/src/activities/summarize/chat-stream-summarize.ts
```diff
@@ -195,13 +271,27 @@ export class ChatStreamSummarizeAdapter<
     options: SummarizationOptions<TProviderOptions>,
     systemPrompt: string,
   ): TextOptions<TProviderOptions> {
+    // Sampling knobs now live in provider-native `modelOptions`. Apply the
+    // low-temperature default underneath any caller-supplied `modelOptions` so
+    // callers can still override it.
+    let working: Record<string, unknown> = {
+      temperature: 0.3,
+      ...(options.modelOptions as Record<string, unknown> | undefined),
+    }
```
tombeckenham, line 341: Regression for Ollama: `temperature: 0.3` stays flat, but Ollama reads sampling only from `modelOptions.options`, so the default never reaches the wire (it did before this PR). OTel reads the flat value, so telemetry will report 0.3 while the request omits it. Nest temperature under `options` for the `ollama` name.

packages/ai/src/activities/summarize/chat-stream-summarize.ts
```diff
@@ -23,6 +23,82 @@ export interface ChatStreamCapable {
   chatStream: (options: TextOptions<any>) => AsyncIterable<StreamChunk>
 }
 
+/**
+ * Provider-native max-output-tokens key per text-adapter `name`. summarize is
+ * provider-agnostic and forwards `modelOptions` opaquely to the wrapped text
+ * adapter, so `maxLength` must be written under the exact key the underlying
+ * provider reads — no adapter reads a generic `maxTokens`. A value of `null`
+ * marks a nested shape (handled specially below for Ollama).
```
tombeckenham: Inaccurate: the map is `Record<string, string>` with no `null` and no `ollama` key — Ollama is a hardcoded branch in `applyMaxLength`. Drop the `null` sentence.

packages/ai/src/activities/summarize/chat-stream-summarize.ts
```diff
@@ -23,6 +23,82 @@ export interface ChatStreamCapable {
...
+  'max_output_tokens',
+  'max_tokens',
+  'max_completion_tokens',
+  'maxOutputTokens',
+  'maxCompletionTokens',
+  'maxTokens',
+] as const
+
+/**
+ * Resolve `maxLength` to the provider-native max-output-tokens key for the
```
tombeckenham, line 103: "the text adapter's `name`" → it's *this summarize adapter's* own `name` (constructor arg, defaults to `'chat-stream-summarize'`), independent of the wrapped text adapter.

packages/ai/src/activities/summarize/chat-stream-summarize.ts
```diff
@@ -23,6 +23,82 @@ export interface ChatStreamCapable {
...
+      merged.options && typeof merged.options === 'object'
+        ? (merged.options as Record<string, unknown>)
+        : undefined
+    if (existing && typeof existing.num_predict === 'number') return merged
+    merged.options = { num_predict: maxLength, ...existing }
+    return merged
+  }
+
+  const key = MAX_TOKENS_KEY_BY_ADAPTER[adapterName]
+  if (key === undefined) return merged
```
tombeckenham, line 149: Silent no-op: any `name` not in the map (new provider, the default `'chat-stream-summarize'`, or the new `openaiCompatible` adapter's custom names) drops `maxLength` with no signal. At least `logger.warn` here; better, type adapter `name` as a literal union so the map must be exhaustive.

packages/ai/src/middlewares/otel.ts
```diff
@@ -162,6 +162,17 @@ function messageEventName(role: string): string {
   }
 }
 
+/**
+ * Return the first candidate that is a finite `number`, or `undefined`. Used to
```
tombeckenham, line 170: Says "finite" but the check is only `typeof === 'number'`, so `NaN`/`Infinity` pass. Either guard with `Number.isFinite` or drop "finite".

codemods/move-sampling-to-model-options/transform.test.ts
```diff
@@ -0,0 +1,125 @@
...
+  return s.replace(/\r\n/g, '\n').trim()
+}
+
+function expectFixture(
+  name: string,
+  ext: 'ts' | 'tsx' = 'ts',
+): { reports: Array<string> } {
+  const expected = read(`${name}.output.${ext}`)
+  const { output, reports } = runTransform(name, ext)
+  expect(normalize(output)).toBe(normalize(expected))
```
tombeckenham: This suite is red locally (9/12): recast prints `;`/no trailing commas, fixtures are Prettier-style, and `normalize()` only trims. Transform output is semantically correct — run it through Prettier in the harness (or regenerate fixtures), and wire this into `test:pr` (it currently isn't, since `codemods` has no `test:lib` target).

.changeset/sampling-options-to-model-options.md
```diff
@@ -0,0 +1,31 @@
+---
+'@tanstack/ai': minor
```
tombeckenham, line 2: Please note the removed public export `OllamaTextProviderOptions` (deleted from `ai-ollama/src/index.ts`) — it's a breaking API-surface change; mention the migration to the per-model type / SDK `ChatRequest`.

tombeckenham (commented):
Review of the sampling→`modelOptions` move. Core mapping is correct and well-tested; flagging a few real issues inline. Two blockers:

1. **Codemod suite is red** (9/12 fail) and escapes CI — `codemods` exposes a `test` script, not `test:lib`, so `pnpm test:pr` never runs it.
2. **Ollama `temperature`/token caps land flat** in two places (summarize default + the eval harness) but the adapter reads sampling only from `modelOptions.options` → silently dropped.

Also (not inline-able, examples aren't in the diff): `ts-code-mode-web` (`api.product-codemode.ts:271`, `api.judge.ts:56`, `api.banking-demo.ts:237`) and `ts-react-chat/api.structured-output.ts:409` still pass root `maxTokens` to `chat()` — now silently ignored.

packages/ai-openai/src/text/text-provider-options.ts
```diff
@@ -14,8 +14,28 @@ import type { ToolChoice } from '../tools/tool-choice'
...
+  /**
+   * Upper bound on generated tokens (visible output + reasoning tokens).
+   * https://platform.openai.com/docs/api-reference/responses/create#responses_create-max_output_tokens
+   */
+  max_output_tokens?: number
+}
+
 // Core, always-available options for Responses API
-export interface OpenAIBaseOptions {
+export interface OpenAIBaseOptions extends OpenAISamplingOptions {
```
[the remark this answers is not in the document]
AlemTuzlak: Deferring this one to a follow-up rather than folding it into this PR, for two reasons:

1. **Not a regression introduced here.** Before this PR, `temperature`/`topP` lived on the `chat()` root and were sent to *any* model, reasoning ones included — the runtime behavior (a user can send `temperature` to o3/gpt-5 and get a 400) is unchanged. This PR only relocated the fields; it didn't newly enable sending them to reasoning models at runtime.

2. **It's a sizable, orthogonal type-design change.** Every OpenAI model in `model-meta.ts` (~30 entries) composes `OpenAIBaseOptions`. Splitting sampling out of the shared base and re-applying it only to non-reasoning models touches the whole matrix and reshapes the public per-model type surface — that deserves its own PR with its own tests, not a rider on the sampling-relocation change. The docstring on `OpenAISamplingOptions` already documents that reasoning models reject these.

Tracking as a follow-up: gate `temperature`/`top_p` to non-reasoning OpenAI models at the type level (and optionally filter them in `responses-text.ts` for reasoning models).

AlemTuzlak:
Thanks for the thorough review @tombeckenham + CodeRabbit. Addressed in acb3371a.

### Blockers
- **Codemod suite red + escaping CI.** The 9/12 failures were purely recast-vs-Prettier print style (semicolons, trailing commas) — output was semantically correct. The harness now Prettier-normalizes both sides before comparing (`transform.test.ts`), so it's robust to recast's printer; **20/20 green**. And the root cause of it escaping CI: the `codemods` project had no `test:lib` target, so `nx affected --targets=…,test:lib,…` skipped it. Added `"test:lib": "vitest run"` to `codemods/package.json` — verified it now runs under `pnpm test:pr`.
- **Ollama `temperature: 0.3` dropped in summarize.** Ollama reads sampling only from `modelOptions.options`, so the flat default was dropped at the wire while OTel reported it. The default now nests under `options.temperature` for Ollama (flat for everyone else). Regression test added that asserts both `options.temperature` **and** `options.num_predict` are nested (and that nothing flat leaks).
- **Ollama caps land flat in the eval harness** (`run-eval.ts`) — now nested under `options`.

### CodeRabbit items
- migration.md "After" example: sampling moved into `modelOptions` (native keys); `metadata` stays at root.
- sampling-options-to-model-options.md: removed the leaked `</content></invoke>` tags.
- anthropic `max_tokens ?? 1024` (so an explicit `0` reaches `validateMaxTokens` instead of being coerced).
- openrouter **chat** adapter now forwards root `metadata` like the responses adapter (+ test).
- otel `firstNumber` now requires `Number.isFinite`.
- summarize: honor caller-set flat token limits in the Ollama branch; fixed the stale `null`/"text adapter's name" comments; **warn** instead of silently dropping `maxLength` for an unrecognised adapter name (new `InternalLogger.warn`, gated by the `errors` category so `debug: false` still silences it — names are legitimately open-ended, e.g. `openaiCompatible`, so a literal-union type isn't viable).
- changeset: documented the removed `OllamaTextProviderOptions` export (→ `OllamaChatModelOptionsByName` / SDK `ChatRequest`).
- codemod README: split the OpenAI row into Responses vs Chat Completions.
- codemod transform: drop the unsafe `as RootSamplingKey` cast (`Set<string>`); emit ES6 shorthand when a moved value matches its key.

### Examples (root `maxTokens` now silently ignored)
Fixed the `ts-code-mode-web` (`api.product-codemode`, `api.judge`, `api.banking-demo`, plus the `execute-prompt`/`structured-output` libs) and `ts-react-chat/api.structured-output` call sites — routed through provider-native `modelOptions` via a small `maxTokensModelOptions(adapter, …)` helper for the dynamically-resolved adapters, native keys inline where the adapter is static.

### Deferred / not changed (with reasoning)
- **OpenAI reasoning-model `temperature`/`top_p` type split** (CodeRabbit) — deferred to a follow-up; not a runtime regression (root `temperature` was always sendable pre-PR) and it's a ~30-model type-surface change orthogonal to this PR. Replied in-thread.
- **Ollama `ResolveModelOptions` fallback narrowing** (CodeRabbit, outside-diff) — a type refinement on the arbitrary-model-string fallback; pre-existing design, risks per-model assignability. Worth a follow-up, out of scope here.
- **Move `summarize-max-length.test.ts` next to source** (CodeRabbit nitpick) — skipped: `packages/ai` keeps all 43 tests in `tests/` with zero colocated tests, so moving this one would break the package's actual convention.

Full `pnpm test:pr` green across 34 projects.

tombeckenham (approved):
Made one small change to centralise a set of consts. All good to go

PR #661, opened 2026-05-30
feat(ai-client): client-side chat persistence adapter

AlemTuzlak:
## Summary

Adds an optional, opt-in **persistence adapter** to `ChatClient` so client-side conversations survive reloads — no manual `initialMessages` + `onFinish` boilerplate. Implements TanStack/ai#374.

The adapter uses the same `getItem`/`setItem`/`removeItem` shape as the rest of the SDK and may be sync or async:

```ts
interface ChatClientPersistence {
  getItem: (id) => Array<UIMessage> | null | undefined | Promise<...>
  setItem: (id, messages) => void | Promise<void>
  removeItem: (id) => void | Promise<void>
}

const client = new ChatClient({ id: 'conversation-123', connection, persistence })
```

When provided, `ChatClient`:
- **hydrates** from `getItem(id)` on construction (async hydration is race-guarded against a newly-started conversation),
- **saves** via `setItem(id, messages)` on every message change, through an ordered write queue so writes never overlap or land out of order,
- **clears** via `removeItem(id)` on `clear()`, discarding any in-flight stream so late chunks can't repopulate cleared state.

When omitted, behavior is unchanged — fully backwards compatible. Persistence is best-effort: adapter errors are swallowed so storage problems never break the chat.

## Changes
- `@tanstack/ai-client`: `ChatClientPersistence` interface + `persistence` option on `ChatClient`; hydrate / save-queue / clear integration; mid-stream clear chunk suppression. Exported from the package index.
- Framework wrappers (`react`, `preact`, `solid`, `vue`, `svelte`): thread the `persistence` option through and read hydrated messages from the client after construction.
- Docs: new `docs/chat/persistence.md` (interface, behavior, localStorage + IndexedDB examples), registered under **Chat & Streaming**.
- Changeset: minor bump for all six packages.

## Testing
- New unit tests in `ai-client` (hydration sync/async, save-on-change, clear, mid-stream clear suppression) and in every framework wrapper.
- New E2E test: persists chat across a browser reload with a `localStorage` adapter.
- Verified locally: `ai-client` 248 unit tests pass; all wrapper unit tests pass; `test:types` and `test:eslint` clean across all six packages; `ai-client` builds; `test:docs` passes; the persistence E2E test passes (full E2E suite green).

## Summary by CodeRabbit

* **New Features**
  * Chat messages can now be persisted and automatically restored across page reloads using configurable storage adapters.

* **Documentation**
  * Added a comprehensive guide showing how to enable persistence with examples for localStorage and IndexedDB; navigation updated to include the new page.

* **Tests**
  * Added e2e and extensive unit tests to verify persistence, hydration, clearing behavior, and race conditions for reliable chat state.

[Review Change Stack](https://app.coderabbit.ai/change-stack/TanStack/ai/pull/661?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


packages/ai-client/src/client-persistor.ts
```diff
@@ -0,0 +1,312 @@
...
+      this.markIgnoredChunkIds(chunk)
+      return true
+    }
+
+    if (this.isRunlessChunkFromIgnoredRun(chunk)) {
+      this.markIgnoredChunkIds(chunk)
+      return true
+    }
+
+    const toolCallId = (chunk as { toolCallId?: string }).toolCallId
```
tombeckenham: I'm going to change this. It should be 'toolCallId' in chunk ? chunk.toolCallId : undefined

packages/ai-client/src/client-persistor.ts
```diff
@@ -0,0 +1,337 @@
...
+/**
+ * Encapsulates everything persistence-related for `ChatClient` so the client
+ * itself stays focused on streaming and message state.
+ *
+ * Two responsibilities live here:
+ *
+ * 1. **Storage orchestration** — hydrate from `getItem(id)` on creation, save to
+ *    `setItem(id, messages)` on every change through an ordered write queue, and
+ *    `removeItem(id)` on clear. A generation counter discards stale writes when a
+ *    removal or a newer conversation supersedes an in-flight async operation.
```
tombeckenham, line 36: Doc nit (non-blocking): this says the generation counter discards writes when *a removal or a newer conversation* supersedes an op — but only `remove()` bumps `generation`, and a `ChatPersistor` is bound to a single `id` for its lifetime, so there is no in-instance *newer conversation* path on this counter. The superseding-hydration case is handled by the separate `messagesGeneration` counter instead. Suggest dropping `or a newer conversation` here so a maintainer isn't sent looking for a code path that doesn't exist.

packages/ai-client/src/client-persistor.ts
```diff
@@ -0,0 +1,337 @@
...
+  private skipNextPersist = false
+  private generation = 0
+  private queue: Promise<void> = Promise.resolve()
+  private queuePending = false
+  // Bumped on every message change; lets an in-flight async hydration detect
+  // that the message list moved on and avoid clobbering it.
+  private messagesGeneration = 0
+
+  // --- clear-during-stream suppression state ---
+  private readonly clearedMessageIds = new Set<string>()
```
tombeckenham, line 56: Follow-up (non-blocking): `clearedMessageIds` and `clearedToolCallIds` are only ever added to — unlike `clearedRunIds` / `ignoredActiveRunIds`, nothing prunes them. Over a long-lived client that is cleared repeatedly they grow for the session lifetime, and because membership is permanent a *new* chunk whose `messageId` / `toolCallId` collides with a previously-cleared id would be wrongly suppressed (unlikely with random ids, but real for deterministic / server-assigned ids). Worth pruning these once no cleared run is still draining. Not blocking — flagging for a follow-up.

packages/ai-client/src/connection-adapters.ts
```diff
@@ -13,6 +13,26 @@ import type {
...
+ * Content events (TEXT_MESSAGE_CONTENT, TOOL_CALL_*, …) carry no `runId` of
+ * their own, so the connect wrapper stamps the caller's run id here. Lets
+ * run-scoped consumers (e.g. clear-during-stream suppression) attribute those
+ * otherwise-runless chunks to their originating request.
+ */
+const chunkRunIds = new WeakMap<StreamChunk, string>()
+
+/**
+ * Resolve a chunk's run id, preferring the value on the chunk itself
+ * (RUN_STARTED / RUN_FINISHED / RUN_ERROR carry one) and falling back to the
```
tombeckenham, line 27: Doc nit (non-blocking): `RUN_ERROR` is grouped here with `RUN_STARTED` / `RUN_FINISHED` as carrying `runId` on the chunk itself, but per the AG-UI schema only `RUN_STARTED` / `RUN_FINISHED` define `runId` in-schema — `RUN_ERROR` carries it via the passthrough (which the chat-client comment already states correctly). Minor wording.

packages/ai-client/src/chat-client.ts
```diff
@@ -609,33 +675,27 @@ export class ChatClient<
...
-        } else if (chunk.type === 'RUN_ERROR') {
-          // RUN_ERROR without runId is a session-level error; clear all runs
-          this.activeRunIds.clear()
-        }
-        this.setSessionGenerating(this.activeRunIds.size > 0)
-        this.resolveProcessing()
-      }
+      // Run lifecycle (active-run tracking, session-generating state, and
+      // processing resolution for RUN_FINISHED / RUN_ERROR) is handled in a
+      // single place so the ignored-chunk path above and this path can't
```
tombeckenham, line 694: nit (non-blocking): *single place so the … paths can't diverge* slightly overstates it — a runId-less `RUN_ERROR` on the ignored path is drained by `drainIgnoredRunlessChunk`, which re-implements the same lifecycle steps rather than routing through `updateRunLifecycle`. So there are in fact two places that can drift. Might be worth noting the runless case is handled separately.

tombeckenham (approved):
Reviewed the client-persistence implementation in depth — the streaming queue, generation counters, and clear-during-stream suppression are well thought through, and the type/test coverage is solid. Approving. ✅

I pushed one fix + test coverage to the branch (1fb316ca):

- **Bug fix:** `takeRunlessRunId` nulled `currentRunlessRunId` instead of advancing it to the next still-ignored run (unlike its sibling `onRunSettled`). With two runs cleared mid-stream concurrently, draining one via a runId-less `RUN_ERROR` stopped suppressing the *other* run's late content, which could repopulate a cleared conversation. Now advances the pointer.
- **E2E:** added specs for `clear()`-removes-storage-across-reload and in-place thread/id switching — the two riskiest behaviors of the feature, which previously only had reload coverage.

A few non-blocking notes inline (mostly doc nits + one latent follow-up on unbounded `cleared*Ids` growth).

**One design question (not blocking):** persistence adapter errors are swallowed everywhere — which is the right call, the chat must survive a full/unavailable/corrupt store. But there's currently *zero* signal when an adapter is simply misconfigured (e.g. a dev's IndexedDB adapter always rejects → it silently never persists). Worth considering an optional `onPersistenceError(err, op)` callback, or a dev-only `console.warn`, so failures aren't invisible while keeping the chat unbreakable. Happy to add it in a follow-up if you'd like.

husayt:
@AlemTuzlak  how do we change chat instance id in vue, if neither id field in createChatClientOptions nor chatOptions are reactive? Currently there is no way to change id of chat, except for creating a new chat, unless I am missing something. that means for  switching sessions I have create a new chat with useChat every time

Thank you

AlemTuzlak:
@husayt threadId is the way, but are you saying its not reactive? It hsould be 🤔

husayt:
Neither id nor threadId are reactive for Vue.  Currently there is no other way to switch sessions other than creating new uscChat. Currently no input option for useChat is reactive (except for forwardedProps). That means one has to change any of them one has to create new chat instance.

Also, having read documentation multiple times i had an impression that I had to use id for sessions, not threadId.

Would be great to clarify their difference

husayt:
@AlemTuzlak I just double checked the documentation it mention id not threadId  for used persistence. Still neither of them are reactive. So persistence currently with vue works only for single session chats

https://tanstack.com/ai/latest/docs/chat/persistence

PR #665, opened 2026-05-30
feat(ai-bedrock): add Amazon Bedrock adapter (Converse default + OpenAI-compatible chat/responses)

AlemTuzlak:
## 🎯 Summary

Adds the first-party **`@tanstack/ai-bedrock`** package: an Amazon Bedrock adapter whose **default path uses Bedrock's Converse API** (`@aws-sdk/client-bedrock-runtime`), with Bedrock's **OpenAI-compatible** Chat Completions and Responses APIs available as opt-in modes.

```ts
import { bedrockText } from '@tanstack/ai-bedrock'

// Converse (DEFAULT) — model-agnostic; reaches Claude, Nova, Llama, Mistral, …
const adapter = bedrockText('us.anthropic.claude-haiku-4-5-20251001-v1:0', { region: 'us-east-1' })

// OpenAI-compatible Chat Completions (open-weight models, e.g. gpt-oss)
const chatCompletions = bedrockText('openai.gpt-oss-120b-1:0', { region: 'us-east-1', api: 'chat' })

// OpenAI-compatible Responses (gpt-oss, stateful)
const responses = bedrockText('openai.gpt-oss-120b-1:0', { region: 'us-east-1', api: 'responses' })
```

## Why Converse is the default

Per AWS's [API-compatibility matrix](https://docs.aws.amazon.com/bedrock/latest/userguide/models-api-compatibility.html), the OpenAI-compatible Chat Completions / Responses endpoints reach **open-weight models only** — **Anthropic Claude, Amazon Nova, and Meta Llama are not exposed via those endpoints** (they are Converse / Invoke only). Converse is a **strict superset** for text/chat models: 0 models are ChatCompletions-only, 41 are Converse-only. To "support the most models," Converse must be the default.

`bedrockText` / `createBedrockText` branch on `api: 'converse' | 'chat' | 'responses'` (default `'converse'`). The Converse adapter extends `BaseTextAdapter` directly and implements its own translation layer (messages, tools, streaming, structured output); the Chat Completions and Responses adapters are unchanged, built on `@tanstack/openai-base`.

## ✨ What's included

- **`BedrockConverseTextAdapter`** (default) — streaming, client-side tool calling, structured output (forced single-tool), reasoning-output deltas, and the agent loop, mapping TanStack types ⇄ the Converse wire format.
- **Three per-API type-safe catalogs** (`BEDROCK_CONVERSE_MODELS` / `BEDROCK_CHAT_MODELS` / `BEDROCK_RESPONSES_MODELS`) derived from a generated catalog with per-API support flags + a compile-time `Extract` filter that preserves literal model-id unions.
- **Unified auth** — `apiKey` → `BEDROCK_API_KEY` → `AWS_BEARER_TOKEN_BEDROCK` → **SigV4** via the AWS credential chain. SigV4 is handled natively by `@aws-sdk` for the Converse path and by `@aws-sdk`'s `SignatureV4` signer for the OpenAI-compatible paths. The previous `aws-sigv4-fetch` optional peer is **removed**.
- **Maintainer refresh script** — `scripts/fetch-bedrock-models.ts` (`ListFoundationModels` + a curated API-compatibility seed) regenerates `src/model-catalog.generated.ts`.
- **66 unit tests** across the converters, stream processor, structured output, adapter, factory, auth, and catalogs. `pnpm test:pr` green (sherif, knip, docs, eslint, lib, types, build).

## ⚠️ Maintainer notes (please read)

- **The committed model catalog is a verified 14-model seed, not the full Converse catalog.** Model IDs were not fabricated — there were no AWS credentials in the implementation environment to run `ListFoundationModels`. The architecture reaches every Converse model; **run `scripts/fetch-bedrock-models.ts` with AWS credentials to expand the catalog** to the full set with confirmed IDs/modalities. Per-account/region availability depends on your model access.
- **Converse E2E is deferred** — `@copilotkit/aimock` cannot mock the AWS binary event-stream (`vnd.amazon.eventstream`) Converse protocol, so `bedrock-converse` is not in the Playwright matrix. The translation layer is covered by unit tests; the OpenAI-compatible `bedrock` / `bedrock-responses` entries remain in the E2E matrix. Tracked in #669.
- **Reasoning request-side enable is deferred** (per-model `additionalModelRequestFields` differ); reasoning *output* deltas already surface.
- This PR also fixes a `docs/config.json` regression where a stale rewrite had dropped the React Native quickstart, the Structured Outputs nav section, and the Typed Options guide — the nav diff vs `main` is now purely additive.

## 🚀 Release Impact

- New published package `@tanstack/ai-bedrock` (changeset: minor).

## Summary by CodeRabbit

* **New Features**
  * Added Amazon Bedrock adapter package with three selectable API modes (Converse default, Chat Completions, Responses), plus adapters for `bedrock`, `bedrock-responses`, and `bedrock-converse` supporting streaming, tools, structured output, and reasoning.
  * Added Bedrock multimodal message metadata types.

* **Authentication**
  * Documented flexible auth resolution (explicit key, `BEDROCK_API_KEY`, legacy token, or AWS SigV4).

* **Documentation**
  * Updated Bedrock adapter docs, provider/reference links, and added package-level README and getting-started navigation.

* **Tests / E2E**
  * Expanded unit tests and included Bedrock variants in the E2E provider matrix.

* **Chores**
  * Added generated model catalog and generator, plus tooling/config updates.


AlemTuzlak:
Thanks @coderabbitai — addressed the review in `6ac6e40c` (rebased onto the autofix commit):

**Fixed**
- **`bedrockText()` short-circuiting non-API-key auth** — removed the eager `getBedrockApiKeyFromEnv()` call; the env-key factory now passes config through and lets `withBedrockDefaults`/`resolveBedrockAuth` resolve auth lazily, so `{ auth: 'sigv4' }` no longer throws. Added a regression test.
- **Authoritative `apiKey`** — `createBedrockChat` / `createBedrockResponsesText` now spread `{ ...config, apiKey }` so the explicit parameter wins over a stray `config.apiKey`.
- **SigV4 error passthrough** — the `aws-sigv4-fetch` dynamic import now only remaps genuine module-not-found errors to the install hint and rethrows everything else.
- **adapter-configuration skill** — added `AWS_BEARER_TOKEN_BEDROCK` to the env-var list.
- **refresh script** — `ListFoundationModels` / `ListInferenceProfiles` now paginate via `nextToken`.

**Intentionally not changed**
- **`@tanstack/ai` peer `workspace:^`** — kept to match every sibling adapter (`ai-groq`, `ai-grok`, `ai-openrouter` all use `workspace:^` for the `@tanstack/ai` peer); `test:sherif` passes.
- **`"Structured Outputs (Moved)"` nav label** — that (and other nav churn) came from `sync-docs-config` resyncing a `config.json` that was stale on `main`, unrelated to this adapter. I reverted `docs/config.json` to a surgical diff containing only the new Bedrock adapter entry, so this PR doesn't carry the unrelated nav changes.

packages/ai-bedrock/src/adapters/converse-text.ts
```diff
@@ -0,0 +1,560 @@
...
+  /**
+   * Dynamically import `@aws-sdk/client-bedrock-runtime`. The specifier is held
+   * in a variable (not a string literal) so bundler dep scanners (e.g. Vite/
+   * esbuild optimizeDeps) cannot statically discover the AWS SDK and try to
+   * pre-bundle it for the browser — it would fail on the SDK's Node-only
+   * `fromTokenFile` export chain. The SDK is Node/server-only and is only
+   * reached on a real request. `typeof import(...)` is a type-only reference
+   * (erased at emit) so the imported members keep full typing.
+   */
+  protected importBedrockRuntime(): Promise<typeof BedrockRuntime> {
```
tombeckenham, line 103: Explain this a bit further? Are you saying if this library was imported into a tanstack start app and only ever used on the backend, that it'd still be imported in the frontend? I don't think it would...

packages/ai-bedrock/src/adapters/converse-text.ts
```diff
@@ -0,0 +1,560 @@
...
+            ...(options.temperature !== undefined && {
+              temperature: options.temperature,
+            }),
+            ...(options.topP !== undefined && { topP: options.topP }),
+            ...(options.maxTokens !== undefined && {
+              maxTokens: options.maxTokens,
+            }),
+          }
+        : undefined
+
```
tombeckenham, line 538: These don't pass type check

packages/ai-bedrock/src/converse/message-converter.ts
```diff
@@ -0,0 +1,262 @@
...
+      document: {
+        format: documentFormat(source.mimeType),
+        name: `document-${docIndex}`,
+        source: { bytes: base64ToBytes(source.value) },
+      },
+    }
+  }
+
+  // Fail loud for unsupported part types (audio, video, etc.)
+  const unsupported = (part as ContentPart).type
```
tombeckenham: cast is unnecessary

packages/ai-bedrock/src/utils/auth.ts
```diff
@@ -0,0 +1,79 @@
...
+    // Lazy credential provider: the AWS SDK is Node/server-only, so we defer the
+    // dynamic import until SigV4 actually needs to resolve credentials. The
+    // specifier is held in a variable (not a string literal) so bundler dep
+    // scanners (e.g. Vite/esbuild optimizeDeps) cannot statically discover the
+    // AWS SDK and try to pre-bundle it for the browser — it would fail on the
+    // SDK's Node-only `fromTokenFile` export chain. `typeof import(...)` is a
+    // type-only reference (erased at emit) so we keep full typing.
+    credentials: async (...args) => {
+      const mod = '@aws-sdk/credential-providers'
+      const { fromNodeProviderChain } = (await import(
```
tombeckenham, line 74: Again, do we really need to import this way?

packages/ai-bedrock/src/adapters/converse-text.ts
```diff
@@ -0,0 +1,560 @@
...
+        return block.toolUse.input
+      }
+    }
+  }
+  // Second pass: accept the first toolUse block regardless of name.
+  for (const block of content) {
+    if ('toolUse' in block && block.toolUse) {
+      return block.toolUse.input
+    }
+  }
```
tombeckenham: **Critical — structured output can return the wrong tool's data.** After preferring `STRUCTURED_TOOL_NAME` (or an unnamed tool), the second pass returns the first `toolUse` block *regardless of name*. If the model emits a different tool call (hallucinated or leftover), its arbitrary `input` is handed back as the validated structured result — either failing downstream schema validation with a misdirected error, or silently returning semantically wrong data.

Suggest: match `STRUCTURED_TOOL_NAME` only (the `name === undefined` allowance is fine); if no matching block is found, return `undefined` so the existing `throw` at ~`:228` fires with an accurate message instead of masking the cause.

packages/ai-bedrock/src/index.ts
```diff
@@ -0,0 +1,178 @@
...
+    if (!isResponsesModel(model)) {
+      throw new Error(
+        `Model "${model}" is not available on the Bedrock Responses API. ` +
+          `Responses-capable models: ${BEDROCK_RESPONSES_MODELS.join(', ')}.`,
+      )
+    }
+    return new BedrockResponsesTextAdapter(rest, model)
+  }
+  if (config?.api === 'chat') {
+    return new BedrockTextAdapter(stripApi(config), model as BedrockChatModels)
```
tombeckenham: **Critical — unguarded `as` cast loses type safety on the internal `build` seam.** The `'responses'` branch directly above validates via `isResponsesModel` and throws, but this `'chat'` branch does `model as BedrockChatModels` with no runtime guard. `build` types `model` as the full `BedrockConverseModels` union (a superset of chat), so this silently widens a possibly-chat-incapable id. Harmless today because every catalog entry is chat-capable, but a generated-catalog refresh that adds a `chat: false` model makes this silently wrong (the responses branch would catch the analogous case; this one won't).

Suggest: add a symmetric `isChatModel` guard + descriptive throw, mirroring `isResponsesModel`.

packages/ai-bedrock/README.md
```diff
@@ -0,0 +1,94 @@
...
+```
+
+## Authentication
+
+Auth is resolved in this order:
+
+1. Explicit `apiKey` passed to the factory
+2. `BEDROCK_API_KEY` environment variable
+3. `AWS_BEARER_TOKEN_BEDROCK` environment variable
+4. SigV4 via the AWS credential chain (requires `pnpm add aws-sigv4-fetch`)
```
tombeckenham: **Important — documents a removed peer dependency.** The README tells users SigV4 "requires `pnpm add aws-sigv4-fetch`", but that peer was deliberately replaced by the bundled `@smithy/signature-v4` (`src/utils/openai-sigv4-fetch.ts` comment: "Replaces the old aws-sigv4-fetch peer"). The package has no such dep/peer, and the canonical docs (`docs/adapters/bedrock.md:78`) correctly say "no additional packages required." Remove the `aws-sigv4-fetch` install steps.

(Also in this file: the quick-start uses the superseded `claude-3-7-sonnet` id — the catalog ships `claude-sonnet-4-5` / `claude-haiku-4-5`; and the one-line description omits Converse, which is the default path.)

docs/adapters/bedrock.md
```diff
@@ -0,0 +1,241 @@
...
+
+Set `api: 'chat'` to use Bedrock's OpenAI-compatible Chat Completions endpoint. Returns a `bedrock` adapter.
+
+**Model scope:** Open-weight models only — gpt-oss, DeepSeek V3.x, Gemma, Qwen, Mistral open models, GLM, and similar. Claude, Nova, and Llama are NOT available on this endpoint. See the [AWS API compatibility matrix](https://docs.aws.amazon.com/bedrock/latest/userguide/models-api-compatibility.html) for the current list.
+
+```typescript
+import { bedrockText } from '@tanstack/ai-bedrock'
+import { chat } from '@tanstack/ai'
+
+const adapter = bedrockText('openai.gpt-oss-mini-1:0', {
```
tombeckenham: **Important — non-existent model id in a runnable sample.** `openai.gpt-oss-mini-1:0` is not in the catalog; the only chat/responses-capable ids are `openai.gpt-oss-120b-1:0` and `openai.gpt-oss-20b-1:0`. Because `bedrockText` is typed on the literal `BedrockChatModels` union, this won't type-check (CLAUDE.md requires doc samples to type-check), and `test:docs` only verifies links so CI won't catch it. Use `openai.gpt-oss-20b-1:0` (or `-120b-`).

packages/ai-bedrock/src/utils/auth.ts
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+const DEFAULT_REGION = 'us-east-1'
+
+function readApiKeyFromEnv(): string | undefined {
+  try {
+    return getApiKeyFromEnv('BEDROCK_API_KEY')
+  } catch {
+    try {
+      return getApiKeyFromEnv('AWS_BEARER_TOKEN_BEDROCK')
```
tombeckenham: **Important — blind nested catches + silent auth degradation.** `readApiKeyFromEnv` uses exceptions for control flow with two nested blind `catch`es collapsing to `undefined`, then `'auto'` mode silently falls through to SigV4. Any non-"not set" exception (present-but-empty var, sandboxed env read error) is swallowed, and a user who *intended* bearer auth but set a malformed key gets an opaque downstream SigV4 403 with nothing pointing back to this resolution.

Suggest: read `process.env[...]` directly and treat empty string as absent rather than try/catch; and make the chosen auth branch observable (the resolution is currently invisible — no logger).

packages/ai-bedrock/src/converse/message-converter.ts
```diff
@@ -0,0 +1,262 @@
...
+          parsed !== null &&
+          typeof parsed === 'object' &&
+          !Array.isArray(parsed)
+        ) {
+          input = parsed as DocumentType
+        }
+      } catch {
+        // Malformed / partial JSON — fall back to empty object so the call
+        // can still be forwarded rather than crashing the whole request.
+        input = {}
```
tombeckenham: **Important — malformed tool-call args silently become `{}`.** Both the `catch` and the non-object guard (args that parse to an array/number/string) converge on `input = {}` with no logging. These arguments come from a prior assistant turn the engine already accepted, so malformed JSON here usually signals a real upstream bug (truncated stream accumulation, bad provider response) — and the model then silently re-runs the tool with empty args. `messageToBlocks` has no logger, so this is undebuggable later.

Suggest: thread a logger through and warn/error with the tool name + offending raw string, or surface it to the caller, instead of silently coercing to `{}`.

packages/ai-bedrock/src/adapters/converse-text.ts
```diff
@@ -0,0 +1,560 @@
...
+    options: TextOptions<TProviderOptions>,
+  ): AsyncIterable<StreamChunk> {
+    try {
+      options.logger.request(
+        `activity=chat provider=${this.name} model=${this.model} messages=${options.messages.length} tools=${options.tools?.length ?? 0} stream=true`,
+        { provider: this.name, model: this.model },
+      )
+      const input = this.buildInput(options)
+      const stream = await this.sendStream(input)
+      yield* processConverseStream(stream, () => this.generateId())
```
tombeckenham: **Important — `RUN_STARTED` drops the incoming `threadId`/`parentRunId`.** The chat path generates a fresh `threadId` and never sets `parentRunId`, whereas every sibling adapter (openai-base chat/responses) emits `threadId: options.threadId ?? generate(...)` + `parentRunId: options.parentRunId` — and this adapter's own `structuredOutputStream` (~`:259`, `:291`) threads them correctly. The default Converse streaming path is the inconsistent one; pass the incoming `threadId`/`parentRunId` into `processConverseStream` so the emitted lifecycle matches the rest of the SDK.

packages/ai-bedrock/src/converse/stream-processor.ts
```diff
@@ -0,0 +1,267 @@
...
+        }
+        toolCallsByIndex.delete(stopIndex)
+      }
+      continue
+    }
+
+    if ('messageStop' in ev) {
+      const stopReason = ev.messageStop?.stopReason
+      // Map Converse stopReason to AG-UI's narrower finishReason vocabulary.
+      finishReason =
```
tombeckenham, line 246: **Important — unit gaps where Converse E2E was deferred.** The E2E deferral itself is justified (aimock can't emit the `vnd.amazon.eventstream` binary protocol), but the unit suite meant to substitute for it under-tests exactly the branches E2E would have exercised hardest:

- `messageStop` → `finishReason` mapping (this block) and `usage`/`finishReason` folding into the terminal `RUN_FINISHED` — current tests only assert `types.toContain(RUN_FINISHED)`, never the payload.
- the orphaned tool-call drain (`~:232-243`) for truncated streams — no test omits `contentBlockStop`.
- reasoning→text ordering / `closeReasoning()` (`~:73-83`, `:246-255`).
- `structuredOutputStream` error paths (`converse-text.ts:357-390`: empty-response, parse-error).

These are pure-function tests (no SDK needed) and each guards a real hang/silent-failure. Worth adding before merge to make the deferral sound.

tombeckenham (commented):
Automated review pass (Claude Code, multi-agent). Posting the **Critical** and **Important** findings as inline comments below. Freshness: PR is 0 behind main, mergeable, no rebase needed — these are content findings, not staleness.

**Critical**
1. Structured-output second pass can return the wrong tool's data — `converse-text.ts`
2. Unguarded `model as BedrockChatModels` cast in the chat factory branch — `index.ts`

**Important**
3. README documents a removed `aws-sigv4-fetch` peer dependency
4. Docs chat sample uses a non-existent model id
5. Auth: blind nested catches + silent SigV4 fallback
6. Malformed tool-call args silently coerced to `{}`
7. `chatStream` discards incoming `threadId`/`parentRunId` on `RUN_STARTED`
8. Unit gaps where Converse E2E was deferred (stream finish/usage, tool drain, reasoning order)

packages/ai-bedrock/src/converse/structured-output.ts
```diff
@@ -0,0 +1,24 @@
...
+      {
+        toolSpec: {
+          name: STRUCTURED_TOOL_NAME,
+          description: 'Return the final answer as structured JSON.',
+          inputSchema: { json: schema as DocumentType },
+        },
+      },
+    ],
+    toolChoice: { tool: { name: STRUCTURED_TOOL_NAME } },
+  }
```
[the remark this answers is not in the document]
AlemTuzlak: Skipping — not valid for this caller. `buildStructuredToolConfig` is only called from converse-text's `structuredOutput`/`structuredOutputStream` with `options.outputSchema`, which the activity (ai) layer has already converted to JSON Schema before it reaches the adapter (same contract openai-base relies on: "The outputSchema is already JSON Schema (converted in the ai layer)"). So no Zod object can reach `inputSchema.json` here.

packages/ai-bedrock/src/utils/openai-sigv4-fetch.ts
```diff
@@ -0,0 +1,55 @@
...
+      path: url.pathname + url.search,
+      headers,
+      body,
+    }
+
+    const signed = await signer.sign(request)
+    return baseFetch(url.toString(), {
+      ...init,
+      headers: signed.headers as Record<string, string>,
+    })
```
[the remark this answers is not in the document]
AlemTuzlak: Skipping — the `Request`-input branch isn't reachable in practice. This signed fetch is only handed to the OpenAI SDK client, which always invokes it as `fetch(urlString, init)` (never with a `Request` object), so method/headers/body always arrive via `init` and are preserved. Adding `Request`-merging would be dead defensive code for a path the SDK never exercises.

packages/ai-bedrock/src/adapters/converse-text.ts
AlemTuzlak: You're right that backend-only usage wouldn't pull the AWS SDK into a frontend bundle — a dynamic import already keeps it out of the static graph, so for a TanStack Start app using this only on the server it won't reach the client. The variable-specifier + `@vite-ignore` isn't about that; it suppresses Vite's *dev-time* `optimizeDeps` scanner, which eagerly tries to pre-bundle dynamically-imported deps it can statically see, and the AWS SDK's Node-only `fromTokenFile` export chain breaks that pre-bundle. It's defensive. If you'd rather keep it simple, I'm happy to drop the `mod` indirection and use a plain `import('@aws-sdk/client-bedrock-runtime')` (keeping `@vite-ignore`) — your call on whether the optimizeDeps edge case is worth guarding.

packages/ai-bedrock/src/utils/auth.ts
AlemTuzlak: Same pattern/reasoning as the converse-text dynamic import — it's the dev-time Vite optimizeDeps scanner, not frontend tree-shaking. Happy to simplify both to plain string-literal dynamic imports if you'd prefer; flagging them together so we keep them consistent.

packages/ai-bedrock/src/adapters/converse-text.ts
AlemTuzlak: Resolved — this was the old top-level `options.temperature/topP/maxTokens` read. main moved sampling options into `modelOptions`, so `buildInput` now reads `modelOptions.temperature/top_p/max_completion_tokens` and maps them to Converse `inferenceConfig`. `test:types` passes.

packages/ai-bedrock/src/adapters/converse-text.ts
AlemTuzlak: Fixed in 4bb126ac. `extractStructuredToolInput` now only accepts a block named `STRUCTURED_TOOL_NAME` (or an unnamed block); the second pass that returned any toolUse input is gone, so a hallucinated/leftover tool call returns `undefined` and the existing throw fires with an accurate message. Added a unit test that a non-forced tool-use block rejects.

packages/ai-bedrock/src/index.ts
AlemTuzlak: Fixed in 4bb126ac. Added an `isChatModel` guard mirroring `isResponsesModel`; the `api:'chat'` branch now throws on a non-chat id instead of `model as BedrockChatModels`.

packages/ai-bedrock/src/adapters/converse-text.ts
AlemTuzlak: Fixed in 4bb126ac. `processConverseStream` takes a lifecycle arg now; `chatStream` passes `options.threadId`/`parentRunId`/`model`, threaded onto RUN_STARTED (and the threaded threadId onto RUN_FINISHED). Added a test asserting they propagate.

packages/ai-bedrock/src/converse/message-converter.ts
AlemTuzlak: Fixed in 4bb126ac. Threaded the InternalLogger through `toConverseMessages`/`messageToBlocks`; malformed JSON and non-object parses now log via `logger.errors` with the tool name + raw args instead of silently coercing to `{}`.

packages/ai-bedrock/src/converse/stream-processor.ts
AlemTuzlak: Fixed in 4bb126ac. Added unit tests for: finishReason mapping + usage folding into the terminal RUN_FINISHED (payload asserted), the orphaned tool-call drain on a truncated stream, reasoning->text ordering (REASONING_MESSAGE_END before TEXT_MESSAGE_START), and structuredOutputStream empty-response / parse-error paths.

packages/ai-bedrock/src/converse/message-converter.ts
AlemTuzlak: Fixed in 4bb126ac — dropped the cast, `part.type` is accessible directly after the preceding guards.

docs/adapters/bedrock.md
AlemTuzlak: Fixed in 4bb126ac — `openai.gpt-oss-20b-1:0` (the earlier kiira rollout commit already corrected this; confirmed it type-checks under kiira).

packages/ai-bedrock/README.md
AlemTuzlak: Fixed in 4bb126ac. Removed the `aws-sigv4-fetch` install steps (it was replaced by the bundled `@smithy/signature-v4`), updated the description to mention the native Converse API (default), and replaced the superseded `claude-3-7-sonnet` ids with `claude-haiku-4-5-20251001-v1:0`. Also fixed the stream-loop snippet (`TEXT_MESSAGE_CONTENT` / `chunk.delta ?? ''`).

packages/ai-bedrock/src/utils/auth.ts
AlemTuzlak: Partly addressed in 4bb126ac. Replaced the nested blind try/catch with a direct `process.env` read that treats empty/whitespace as absent — no more swallowing unrelated errors as 'key not set' (and dropped the now-unused `@tanstack/ai-utils` dep). The observability half (logging which auth branch resolved) I left out for now: `resolveBedrockAuth` runs inside the lazy `getClient` with no logger in scope, so surfacing it cleanly means threading a logger into `getClient` across all three adapters. Happy to do that as a follow-up if you want it in this PR.

tombeckenham:
Added examples and tested all the way through. It works. 
image

tombeckenham (approved):
Tested all the way through. Good for me

KevinVandy (approved):

PR #666, opened 2026-05-30
feat: support multimodal (image) tool results (#363)

### LINKED ISSUE #363: Tool results are always stringified, preventing multimodal (image) tool responses with OpenAI Responses API
## Problem

Tool results are always converted to strings via `JSON.stringify()`, which prevents sending multimodal content (e.g. images) in tool responses when using the OpenAI Responses API.

The OpenAI Responses API supports [multimodal tool outputs](https://developers.openai.com/api/docs/api-reference/responses/create#responses_create-input-input_item_list-item-function_tool_call_output-output) via `function_call_output`, where `output` can be either a string **or** an array of content parts:

```json
{
  "type": "function_call_output",
  "call_id": "call_xyz",
  "output": [
    { "type": "input_image", "image_url": "https://example.com/screenshot.png" },
    { "type": "input_text", "text": "Screenshot of the current state" }
  ]
}
```

This is documented in the [migration guide](https://developers.openai.com/api/docs/guides/migrate-to-responses) and the [function calling guide](https://developers.openai.com/api/docs/guides/function-calling) ("For functions that return images or files, you can pass an array of image or file objects instead of a string").

However, TanStack AI forces all tool results to strings at **two points**, making it impossible to use this capability:

### 1. Server-side tool execution — `tool-calls.ts`

**File:** `packages/typescript/ai/src/activities/chat/tools/tool-calls.ts` ([line 181-182](https://github.com/TanStack/ai/blob/main/packages/ai/src/activities/chat/tools/tool-calls.ts))

```typescript
toolResultContent =
  typeof result === string ? result : JSON.stringify(result)
```

### 2. Client-side tool result processing — `processor.ts`

**File:** `packages/typescript/ai/src/activities/chat/stream/processor.ts` ([line ~150](https://github.com/TanStack/ai/blob/main/packages/ai/src/activities/chat/stream/processor.ts))

```typescript
const content = typeof output === "string" ? output : JSON.stringify(output);
```

### 3. OpenAI adapter message conversion — `text.ts`

**File:** `packages/typescript/ai-openai/src/adapters/text.ts` ([line 709-712](https://github.com/TanStack/ai-openai/blob/main/packages/ai-openai/src/adapters/text.ts))

```typescript
output:
  typeof message.content === string
    ? message.content
    : JSON.stringify(message.content),
```

## Use Case

I am building a visual editor where the AI designs on a canvas. After each tool call (e.g. `add_text`, `edit_image`), I need the AI to **see** a screenshot of the canvas to verify its changes and iterate. The canvas screenshot is available via a URL endpoint.

With the current implementation, returning `{ type: "input_image", image_url: "https://..." }` from a tool gets stringified to `"{\"type\":\"input_image\",\"image_url\":\"https://...\"}"` — the model receives it as plain text and cannot actually see the image.

I verified manually that sending the `function_call_output` with `output` as an array (with `input_image` parts) to the OpenAI Responses API works correctly — the model receives and processes the image.

## Proposed Solution

The `ModelMessage` type for `role: "tool"` should support `content` as either `string` or `Array<ContentPart>`. The stringify logic in the three locations above should preserve arrays/objects when they match the expected multimodal format, and only stringify plain objects that are not content part arrays.

The OpenAI adapter should pass the array through to `function_call_output.output` instead of stringifying it.

## Environment

- `@tanstack/ai`: latest
- `@tanstack/ai-openai`: latest  
- `@cloudflare/tanstack-ai`: latest
- OpenAI adapter uses Responses API (`client.responses.create()`)

AlemTuzlak:
## Problem

Tool results were always coerced to a string via `JSON.stringify()` before reaching a provider adapter, making it impossible to return multimodal content (e.g. an image) from a tool — even though the OpenAI Responses, Anthropic, and Gemini APIs all support multimodal tool/function outputs. This blocks use cases like returning a screenshot from a tool so the model can *see* it.

Closes #363.

## Solution

- **Structural, opt-in detection** (`isContentPartArray` / `normalizeToolResult` in `@tanstack/ai`): a tool that returns a non-empty array whose every element is a valid `ContentPart` is passed through unchanged; **strings and all other return values serialize exactly as before**, so there are no breaking changes.
- **`ToolResultPart.content`** widened to `string | Array<ContentPart>` (in both `@tanstack/ai` and `@tanstack/ai-client`).
- **Adapters** convert the parts into each provider's native multimodal tool-output format:
  - OpenAI Responses → `function_call_output.output` (`input_image` / `input_text`)
  - Anthropic → `tool_result` content blocks (also fixes a latent bug where non-string tool content became `''`)
  - Gemini → `functionResponse.parts` (`inlineData` / `fileData`), text → `response.content`
  - Chat Completions path (Groq, Ollama, Grok, OpenRouter chat) keeps the documented stringify fallback — that API has no multimodal tool message.
- **AG-UI compliance**: `TOOL_CALL_RESULT.content` / `TOOL_CALL_END.result` stream events remain string-only per spec; the multimodal array travels on the tool message itself.

No SDK upgrades required — the installed `openai` / `@anthropic-ai/sdk` / `@google/genai` versions already type multimodal tool outputs.

## Testing

- **Unit**: detection predicate truth table; per-adapter conversion tests (OpenAI/Anthropic/Gemini); a regression test on the real server-tool execution path (`buildToolResultChunks`) asserting `ContentPart[]` is preserved as an array to the adapter while plain objects still stringify.
- **E2E**: a deterministic wire-assertion spec proving the OpenAI Responses adapter sends a structured `input_image` in `function_call_output`; Anthropic/Gemini covered end-to-end + by unit tests (aimock's journal normalizes away multimodal tool content).
- **Live demo**: a new `examples/ts-react-chat` `/image-tool-repro` route — a server tool returns an image of a secret number; the model now reads it back correctly (it could not before).

## Summary by CodeRabbit

* **New Features**
  * Tools can return structured multimodal results (text + images/media); adapters send multimodal outputs natively when supported, with string fallbacks where required.

* **Examples**
  * Added a reproducible image example plus a one‑click UI repro page that shows the reference image, lets you run the repro, and displays a pass/fail verdict.

* **Tests**
  * Added unit and e2e tests verifying multimodal handling and wire-format behavior across providers.


tombeckenham (approved):
Good to go. Nice improvement

crutchcorn (approved):

PR #689, opened 2026-06-03
fix: pass and use threadId

WhatIsACore:
## 🎯 Changes

bug: `threadId` is [documented](https://tanstack.com/ai/latest/docs/api/ai-svelte#options) as an option and exists in the type definitions but fails to be exposed or used in the current code. It is also mentioned in this [migration guide](https://tanstack.com/ai/latest/docs/migration/ag-ui-compliance#conversationid--threadid) but does not work as documented

`threadId` can be passed, but all framework wrappers (useChat/createChat) forget to forward it when constructing ChatClient. Because of this, a fallback auto-generated ID is used which cannot consistently identify the same thread. A one-liner addition to every wrapper ensures it is forwarded and can be used by consumers to correlate or resume threads.

## ✅ Checklist

all checks passed

## 🚀 Release Impact

changeset included as `patch` for all 5 wrappers

## Summary by CodeRabbit

* **Bug Fixes**
  * The `threadId` option is now correctly forwarded to the chat client across all framework wrappers, enabling proper thread-scoped chat behavior.
  * Updated support across Preact, React, Solid, Svelte, and Vue framework implementations.

### DIFF
--- a/.changeset/sour-bugs-like.md
+++ b/.changeset/sour-bugs-like.md
@@ -1,0 +1,9 @@
+---
+'@tanstack/ai-preact': patch
+'@tanstack/ai-svelte': patch
+'@tanstack/ai-react': patch
+'@tanstack/ai-solid': patch
+'@tanstack/ai-vue': patch
+---
+
+Forward `threadId` option to `ChatClient` in all framework wrappers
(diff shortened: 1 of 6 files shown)


AlemTuzlak (approved):

AlemTuzlak:
Thank you!

PR #694, opened 2026-06-03
chore(ai-event-client): drop unused @tanstack/ai peerDependency

### LINKED ISSUE #552: chore(ai-event-client): drop unused peerDependency on @tanstack/ai
## Summary

`@tanstack/ai-event-client` declares `@tanstack/ai` as a `peerDependency` but does not import from it. The peer dep manufactures a package-manifest cycle (since `@tanstack/ai` depends on `@tanstack/ai-event-client`) that has to be neutralised at the Nx layer. Removing it is more accurate, removes the workaround, and unblocks devtools-only consumers.

## What's there today

`packages/typescript/ai-event-client/package.json`:

```json
"peerDependencies": {
  "@tanstack/ai": "workspace:*"
}
```

`packages/typescript/ai-event-client/project.json`:

```json
{ "implicitDependencies": ["!@tanstack/ai"] }
```

The `!@tanstack/ai` exists *only* to tell Nx to ignore the edge that the peerDependency would otherwise infer, so the build graph stays a DAG (`ai-event-client` → `ai`, not both ways).

## Why the peer dep is misleading

- `ai-event-client/src/` has zero imports from `@tanstack/ai`. All shared types (`ContentPart*`, `ToolCallPart`, `ChatMiddleware`, `StreamChunk`, etc.) are mirrored locally — see the comment block at `packages/typescript/ai-event-client/src/devtools-middleware.ts:3-7`:

  > Local mirrors of @tanstack/ai middleware types. Do not import from `@tanstack/ai` here — during `@tanstack/ai`'s own build that resolves to `dist/*.d.ts` while the engine compiles from `src/`, producing incompatible duplicate types.

- `aiEventClient` is just an `EventClient` from `@tanstack/devtools-event-client`. `devtoolsMiddleware()` returns a duck-typed object structurally compatible with `ChatMiddleware`. Nothing in the runtime or type surface requires `@tanstack/ai` to be installed.
- `workspace:*` as a peer dep range doesn't give consumers a meaningful version constraint anyway (it's replaced at pack time).

## Proposed change

1. Remove `peerDependencies` from `packages/typescript/ai-event-client/package.json`.
2. Remove `"implicitDependencies": ["!@tanstack/ai"]` from `packages/typescript/ai-event-client/project.json` (or empty the array).
3. Keep the mirrored-types pattern — that's the real decoupling and it's already correct.

## Risks / caveats

- If product intent is "these two ship in lockstep and consumers should see a version-mismatch warning," a peer dep is the wrong enforcement mechanism. Use changesets fixed/linked packages instead.
- The mirrored types can still drift from their `@tanstack/ai` counterparts without compile-time enforcement — that's a separate concern worth a follow-up (some kind of structural compatibility check at CI time), but it's orthogonal to whether the peer dep should exist.

## Acceptance

- [ ] `peerDependencies` removed from `ai-event-client/package.json`
- [ ] `implicitDependencies` workaround removed from `ai-event-client/project.json`
- [ ] `pnpm install`, `pnpm build`, `pnpm test:lib`, `pnpm test:types`, `publint --strict` all pass
- [ ] No new warnings from Nx about cyclic or missing edges

valeriudev:
Closes #552.

`@tanstack/ai-event-client` declared `@tanstack/ai` as a `peerDependency` but imports nothing from it — the middleware/usage types are mirrored locally. The only real edge runs the other way (`@tanstack/ai` depends on `ai-event-client`), so the peer dep just manufactured a manifest cycle that had to be neutralised with a `!@tanstack/ai` Nx `implicitDependencies` hack. This drops both.

**Why not `linked`/`fixed` changesets instead:** the changeset config keeps both empty — independent versioning is the convention — and `updateInternalDependencies: patch` already bumps `ai` when `ai-event-client` changes. So straight removal; happy to revisit if you'd prefer lockstep.

**Verified:** `pnpm install` no longer warns about the cyclic workspace dependency, and `build` / `test:types` / `test:lib` (14 passing) / `publint --strict` / `nx affected build` all pass.

### DIFF
--- a/.changeset/ai-event-client-drop-peer-dep.md
+++ b/.changeset/ai-event-client-drop-peer-dep.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/ai-event-client': patch
+---
+
+Drop the unused `@tanstack/ai` peerDependency. `@tanstack/ai-event-client` mirrors the middleware types it needs locally and imports nothing from `@tanstack/ai`, so the peer dep only manufactured a package-manifest cycle (`@tanstack/ai` already depends on `@tanstack/ai-event-client`). Removing it — and the matching `!@tanstack/ai` Nx `implicitDependencies` workaround — keeps the build graph a clean DAG and unblocks devtools-only consumers.
--- a/packages/ai-event-client/package.json
+++ b/packages/ai-event-client/package.json
@@ -36,9 +36,6 @@
   "dependencies": {
     "@tanstack/devtools-event-client": "^0.4.1"
   },
-  "peerDependencies": {
-    "@tanstack/ai": "workspace:*"
-  },
   "devDependencies": {
     "@vitest/coverage-v8": "4.0.14"
   },
--- a/packages/ai-event-client/project.json
+++ b/packages/ai-event-client/project.json
@@ -1,4 +1,3 @@
 {
-  "name": "@tanstack/ai-event-client",
-  "implicitDependencies": ["!@tanstack/ai"]
+  "name": "@tanstack/ai-event-client"
 }
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -1211,9 +1211,6 @@ importers:
 
   packages/ai-event-client:
     dependencies:
-      '@tanstack/ai':
-        specifier: workspace:*
-        version: link:../ai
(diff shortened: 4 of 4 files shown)


tombeckenham (approved):
I've been wanting to fix this for ages. Thank you!!

PR #698, opened 2026-06-03
fix(ai): normalize AG-UI snapshot messages to UIMessage[] in MESSAGES_SNAPSHOT handler

### LINKED ISSUE #688: TypeError: Cannot read properties of undefined (reading 'find') for MESSAGES_SNAPSHOT
### TanStack AI version

"@tanstack/ai": 0.26.1   "@tanstack/ai-client": 0.16.1   "@tanstack/ai-react": 0.15.1

### Framework/Library version

"@tanstack/ai": 0.26.1   "@tanstack/ai-client": 0.16.1   "@tanstack/ai-react": 0.15.1

An exception occurs when I receive a message of type MESSAGES_SNAPSHOT

```text
data: {"type":"TEXT_MESSAGE_END","messageId":"30c8bcf9-8691-4a08-ac7e-05553cc32fd1"}

data: {"type":"MESSAGES_SNAPSHOT","messages":[{"id":"msg-1780386991166-22zrq","role":"user","content":"...","content":"Could you clarify what you want changed in this DataApp?"}]}

data: {"type":"RUN_FINISHED","threadId":"resp_015228a4252ea0b2006a1e8cb0728481939ad10b5ff635cb0b","runId":"resp_015228a4252ea0b2006a1e8cb0728481939ad10b5ff635cb0b"}
```
Fail with this error

Image

Image

-

_No response_

None

### Terms & Code of Conduct

amitsaroj:
## 🎯 Changes

Fixes #688

### Root cause

`handleMessagesSnapshotEvent` in `processor.ts` was doing an unsafe type cast:

```ts
// Before (unsafe)
this.messages = [...chunk.messages] as Array<UIMessage>
```

AG-UI `MESSAGES_SNAPSHOT` messages have the shape `{ id, role, content }` — they do **not** have the `parts` array that `UIMessage` requires. The `as Array<UIMessage>` cast masked this at compile time.

When any downstream code called `message.parts.find(...)` — for example, the `onToolCallStateChange` devtools handler in `chat-client.ts` — it crashed with:

```
TypeError: Cannot read properties of undefined (reading 'find')
```

### Fix

Normalize each message through `modelMessageToUIMessage` when it does not already have `parts`:

```ts
// After (safe)
this.messages = chunk.messages.map((msg) =>
  'parts' in msg
    ? (msg as UIMessage)
    : modelMessageToUIMessage(msg as any, generateMessageId()),
)
```

This mirrors the pattern already used in `normalizeToUIMessage` (used by `ChatClient.append`) and is consistent with how all other message ingestion paths work in this codebase.

## ✅ Checklist

- [ ] - [ ] I have tested this code locally with `pnpm run test:pr`.
## 🚀 Release Impact

## 🧪 Test Plan

1. Send a chat message that triggers a `MESSAGES_SNAPSHOT` event (e.g., using an AG-UI agent backend that emits snapshot events).
2. 2. Verify no `TypeError: Cannot read properties of undefined (reading 'find')` is thrown.
3. 3. Existing unit tests in `packages/ai/tests/stream-processor.test.ts` for `MESSAGES_SNAPSHOT resets transient state` cover the core normalization path.

## Summary by CodeRabbit

* **Bug Fixes**
  * Normalize streamed chat `MESSAGES_SNAPSHOT` payloads into a consistent UI message format using a converter, ensuring `message.parts` is always available, preserving message IDs, generating fallback IDs when missing, and maintaining correct routing for following events.

* **Tests**
  * Added regression coverage for snapshot normalization (including legacy shapes), empty snapshots, ID preservation/fallback, and subsequent text event routing without downstream crashes.

* **Documentation**
  * Added a changeset describing the snapshot normalization behavior change.


amitsaroj:
Thanks for the review @coderabbitai — great catch on the ID preservation issue.

The second commit (`b091adc`) addresses this: instead of always generating a fresh `generateMessageId()`, the normalization now uses `(msg as { id?: string }).id ?? generateMessageId()` — preserving the AG-UI message's original `id` when present and only falling back to a generated ID when the message has no `id` field.

This ensures downstream event correlation (e.g. `TEXT_MESSAGE_CONTENT`, `TOOL_CALL_*`, `ensureAssistantMessage`) continues to work correctly after a `MESSAGES_SNAPSHOT` reset.

amitsaroj:
Hi @tombeckenham and TanStack maintainers — this is a small but impactful fix normalizing AG-UI `MESSAGES_SNAPSHOT` messages to `UIMessage[]` format, preventing the `TypeError: Cannot read properties of undefined (reading 'find')` crash. CodeRabbit's review is clean, no conflicts, and all checks pass. Could someone approve the pending workflows and merge when ready? Thank you!

tombeckenham:
Would you be able to add some tests for this @amitsaroj ?

amitsaroj:
Hi @tombeckenham — done! Added 6 new tests in `packages/ai/tests/stream-processor.test.ts` inside the existing `describe('MESSAGES_SNAPSHOT')` block (commit 99c4a8b):

1. **Normalizes AG-UI messages (no `parts`) to `UIMessage[]`** — verifies each message gets a `parts` array and text content is surfaced as a `TextPart`
2. 2. **Preserves original AG-UI message `id`** — regression guard for the earlier version that always called `generateMessageId()` and broke downstream event correlation
3. 3. **No `TypeError` on `parts.find()`** — directly reproduces the crash from `onToolCallStateChange` accessing `message.parts` after a snapshot
4. 4. **Pass-through for messages that already have `parts`** — existing `UIMessage` snapshots are kept as-is and not re-normalized
5. 5. **Fallback `id` generation** — handles AG-UI messages that have no `id` field
6. 6. **Stream continuation after snapshot** — `TEXT_MESSAGE_CONTENT` events routing to snapshot-hydrated messages work correctly

packages/ai/src/activities/chat/stream/processor.ts
```diff
@@ -870,8 +870,16 @@ export class StreamProcessor {
...
   ): void {
     this.resetStreamState()
-    // AG-UI Message[] is compatible with UIMessage[] at runtime
-    this.messages = [...chunk.messages] as Array<UIMessage>
+    // Normalize AG-UI messages to UIMessage[] to ensure each message has
+    // a `parts` array. AG-UI snapshot messages carry `content` but no
+    // `parts`, so casting them directly as UIMessage[] is unsafe and causes
+    // "Cannot read properties of undefined (reading 'find')" when any code
+    // later accesses message.parts (e.g. onToolCallStateChange devtools handler).
+    this.messages = chunk.messages.map((msg) =>
```
tombeckenham: You've replaced one cast with another. Our initial cast was incorrect, so lets find a way to fix it without casting again. Please remove `as any` casting. I think you're missing createdAt as well. Can you find a better way to do this?

tombeckenham (changes requested):
Good find, but needs to be fixed more cleanly. We also need a changeset.

danciudev:
Any updates on this?

tombeckenham:
I've rewritten the fix in a typesafe way

tombeckenham (approved):
Thanks for finding this. I've updated the fix to be more consistently typesafe

AlemTuzlak (approved):

PR #727, opened 2026-06-08
feat(ai): add 'error' terminal to ToolCallState (#718)

### LINKED ISSUE #718: ToolCallState has no error terminal — an errored server-tool execution parks at state "input-complete"
### Versions

`@tanstack/ai` 0.28.0 · `@tanstack/ai-client` 0.16.3 · `@tanstack/ai-react` 0.15.4

### Summary

`ToolCallState` has no error/failed member. When a server tool's execute function throws inside the `chat()` agent loop, the corresponding `tool-call` part on the client never reaches a terminal state — it parks at `state: "input-complete"` forever, with the error only inferable from adjacent data:

- the call part's `output` becomes `{ error: "<message>" }` (from `executeServerTool` pushing `result: { error: message }`, JSON round-tripped on the client), and
- a sibling `tool-result` part arrives with `state: "error"` + `error: "<message>"` (`ToolResultState` *does* have an error member).

Internals involved (0.28.0 dist): `processor.js` `handleToolCallEndEvent` / `handleToolCallResultEvent` map `output-error` → `"input-complete"`, and `message-updaters.js` `updateToolCallWithOutput` defaults errored calls to `"input-complete"`.

### Why this matters

Any UI that renders tool-call lifecycle from the part's `state` (the natural reading of a state machine) has no way to distinguish "still executing" from "failed" — `state === "complete"` never matches an errored call, so the obvious `done` check produces an **infinite spinner** on every tool failure. We hit exactly this in production. The workaround is a hand-rolled terminal mapping that treats error-shaped `output` / the sibling `tool-result` part as terminal — it works, but it encodes undocumented internals that nothing type-checks, so every SDK update can silently change the shape.

### Repro

Self-contained (no network, no model): drive the real pipeline with a scripted adapter whose first iteration emits a tool call for a server tool that throws —

```ts
// scripted adapter emits: RUN_STARTED → TOOL_CALL_START/ARGS/END → RUN_FINISHED(tool_calls),
// then (tool result in transcript) TEXT_MESSAGE_* → RUN_FINISHED(stop)
const failing = toolDefinition({
  name: "workflow_status",
  description: "always throws",
  inputSchema: z.object({ workflow_id: z.string() }),
}).server(async () => { throw new Error("Temporal query failed"); });

// chat({ adapter: scripted, messages, tools: [failing] })
//   → toServerSentEventsResponse → fetchServerSentEvents (in-process fetchClient)
//   → ChatClient / StreamProcessor

// After the run:
expect(part.state).toBe("input-complete");            // ← never terminal
expect(part.output).toEqual({ error: "Temporal query failed" });
expect(resultPart.state).toBe("error");               // sibling tool-result IS terminal
```

We maintain this as a permanent contract test in our app (it drives `chat()` → `toServerSentEventsResponse` → `fetchServerSentEvents` → `ChatClient` end-to-end on every suite run) precisely because the error shape is otherwise unpinned. Happy to share the full test file if useful.

### Ask

An explicit error terminal on `ToolCallState` (e.g. `"error"`), set by the stream processor when a tool execution produces an output-error — making the call part's state machine self-describing, symmetrical with `ToolResultState`, and letting UIs render failure without reverse-engineering the output shape.

Related (smaller): a part whose run is aborted mid-drain (user stop, disconnect) also parks at `"input-complete"` with no output and no SDK-level signal that it can never resolve — a terminal for that case would close the remaining "spinner forever" path, but the error terminal above is the primary ask.

season179:
## Summary

Closes #718.

`ToolCallState` had no error terminal. When a tool execution produced an output error, the **tool-call** UIMessage part parked at `"input-complete"` forever — while its sibling **tool-result** part correctly reached `"error"`. UIs that render lifecycle from `part.state` (the natural reading of a state machine) therefore couldn't distinguish "still executing" from "failed" without reverse-engineering the error-shaped `output` or the sibling part, producing an infinite spinner on every tool failure.

This adds an `'error'` member to `ToolCallState` and transitions the tool-call part to it on an output error, making the state machine self-describing and symmetric with `ToolResultState`:

```ts
if (part.type === 'tool-call' && part.state === 'error') {
  // render failure — no more inferring from output shape
}
```

## What changed

- **`'error'` added to `ToolCallState`** in the three hand-synced copies: `@tanstack/ai`, `@tanstack/ai-client`, `@tanstack/ai-event-client`.
- **`StreamProcessor`** maps output errors to `state: 'error'` for the tool-call part in `addToolResult`, `handleToolCallEndEvent`, and `handleToolCallResultEvent` (the tool-result part already used `'error'`).
- **Completion safety net** (`RUN_FINISHED` / `finalizeStream`): still finalizes the call's *internal* bookkeeping (so it remains in `getResult().toolCalls` / `getState()`), but no longer downgrades a rendered `'error'` part back to `'input-complete'` — including the AG-UI ordering where an `output-error` `TOOL_CALL_RESULT` arrives **before** `TOOL_CALL_END`.
- **Empty-message errors**: a failed client tool that throws with an empty message (e.g. `throw new Error()`) now still reaches `'error'` — error-ness derives from the `output-error` state, not message truthiness.
- **`isToolCallIncluded`** gains an explicit `'error'` arm so failed calls stay in conversation history for the LLM.
- Docs: `chat-architecture.md` notes the new terminal.

## Test plan

- Unit: `@tanstack/ai` (`stream-processor`, `message-updaters`) and `@tanstack/ai-client` (`chat-client-context`), incl. new regression tests for the result-before-`END` ordering and the empty-message error.
- E2E: `tool-error.spec.ts` asserts the failed tool-call part reaches `state === 'error'` end-to-end; the full `tools-test/` suite (40 specs) confirms normal completion, parallel tools, approvals, and continuations are unaffected.
- `pnpm test:pr` green across all affected projects (sherif, knip, docs, eslint, lib, types, build).

## Notes

Scoped intentionally to the error terminal. The secondary "aborted mid-drain" terminal mentioned in #718 is left for a follow-up. `ToolCallState` is currently duplicated across three packages (kept in sync by hand); de-duplicating it is out of scope here.

## Summary by CodeRabbit

* **New Features**
  * Tool calls now expose a terminal "error" state so failed executions are shown distinctly from in-progress or completed runs.

* **Bug Fixes**
  * Prevented failed tool calls from being left in an ambiguous "input-complete" state and from being overwritten during finalization.

* **Documentation**
  * Clarified tool-call lifecycle and error handling in the chat/streaming docs.

### DIFF
--- a/.changeset/tool-call-error-terminal.md
+++ b/.changeset/tool-call-error-terminal.md
@@ -1,0 +1,17 @@
+---
+'@tanstack/ai-event-client': minor
+'@tanstack/ai-client': minor
+'@tanstack/ai': minor
+---
+
+Add an `'error'` terminal to `ToolCallState`. When a tool execution produces an output error, the StreamProcessor now transitions the `tool-call` part to `state: 'error'` instead of parking it at `'input-complete'`.
+
+Previously an errored tool call left the tool-call part at `'input-complete'` forever, so UIs that render lifecycle from the part's `state` could not distinguish "still executing" from "failed" without reverse-engineering the error-shaped `output` or the sibling `tool-result` part. The new terminal makes the tool-call state machine self-describing and symmetric with `ToolResultState` (which already has `'error'`):
+
+```ts
+if (part.type === 'tool-call' && part.state === 'error') {
+  // render failure — no more inferring from output shape
+}
+```
+
+The completion safety net (`RUN_FINISHED` / stream finalization) no longer downgrades a failed tool call back to `'input-complete'`, including when an `output-error` result arrives before `TOOL_CALL_END`.
--- a/packages/ai-client/src/chat-client.ts
+++ b/packages/ai-client/src/chat-client.ts
@@ -1220,14 +1220,17 @@ export class ChatClient<
 
     // Add result via processor. `result.state` is the authoritative error
     // signal; `addToolResult` infers error-ness from the error message being
-    // truthy, so a failed tool with an empty message (e.g. `throw new Error()`)
-    // must still pass a non-empty string to reach the terminal 'error' state.
+    // truthy. Pass an error message ONLY for output-error results (falling back
+    // to a default so an empty message like `throw new Error()` still reaches
+    // the terminal 'error' state), and `undefined` otherwise — so error
+    // signalling derives solely from `result.state`, never from a stray
+    // `result.errorText` on a successful result.
     this.processor.addToolResult(
       result.toolCallId,
       result.output,
       result.state === 'output-error'
         ? result.errorText || 'Tool execution failed'
-        : result.errorText,
+        : undefined,
     )
 
     // If stream is in progress, queue continuation check for after it ends
--- a/packages/ai-client/src/chat-client.ts
+++ b/packages/ai-client/src/chat-client.ts
@@ -1218,11 +1218,16 @@ export class ChatClient<
       context,
     )
 
-    // Add result via processor
+    // Add result via processor. `result.state` is the authoritative error
+    // signal; `addToolResult` infers error-ness from the error message being
+    // truthy, so a failed tool with an empty message (e.g. `throw new Error()`)
+    // must still pass a non-empty string to reach the terminal 'error' state.
     this.processor.addToolResult(
       result.toolCallId,
       result.output,
-      result.errorText,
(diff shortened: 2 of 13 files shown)


tombeckenham (approved):
Solid

PR #732, opened 2026-06-09
fix(ai): undo strict-mode null-widening before structured-output validation

### LINKED ISSUE #731: @tanstack/ai-openrouter: optional structured output fields are rewritten as nullable but rejected by original schema validation
### TanStack AI version

0.24.0

### Framework/Library version

@tanstack/ai-openrouter 0.11.0

1. User provides a Standard Schema where a field is optional.  
2. TanStack converts it to JSON Schema.  
3. OpenRouter adapter rewrites optional fields into required + nullable for strict OpenAI/OpenRouter compatibility.  
5. Model returns null for that field, which the provider schema allowed.  
6. TanStack validates the result against the original Standard Schema, which rejects null.  

https://stackblitz.com/edit/dkxh99pz?file=package.json

Image

Maybe, I'll investigate and start debugging

### Terms & Code of Conduct

DrewHoo:
## Summary

Structured output rejects a valid provider response when the schema has an **optional** field. To satisfy OpenAI-style strict schemas, optional fields are widened to `required` + nullable, so the provider returns `null` for an absent optional — but that `null` is then validated against the *original* schema, where `.optional()` means `T | undefined`, not `T | null`. Validation throws.

This adds a schema-aware `undoNullWidening` to `@tanstack/ai-utils` and runs it on the structured-output result before Standard Schema validation, so the synthesized nulls are dropped while genuine `.nullable()`/`.nullish()` nulls are preserved.

Fixes #731 

## Reproduction

```ts
import * as v from 'valibot'
import { toStandardJsonSchema } from '@valibot/to-json-schema'

const schema = toStandardJsonSchema(
  v.object({ title: v.string(), note: v.optional(v.string()) }),
)
// chat({ outputSchema: schema, ... }) with @tanstack/ai-openrouter
// → StandardSchemaValidationError: Invalid type: Expected string but received null
```

The outbound schema sent to the provider widens `note` to `{ "type": ["string", "null"] }` and adds it to `required`, so the model returns `{ "title": "…", "note": null }`. The original Valibot schema rejects that `null`.

## Root cause

`convertSchemaToJsonSchema(schema, { forStructuredOutput: true })` widens optional → `required` + nullable (correct, and required for strict mode). The round trip just never undoes it: the response is validated against the un-widened original via `parseWithStandardSchema`, which rejects the `null` the widening told the model to send.

## Fix

- **`@tanstack/ai-utils`** gains `undoNullWidening(value, schema)` — a schema-aware counterpart to `transformNullsToUndefined`. It consults the original (un-widened) JSON Schema and drops a `null` only where the schema *disallows* one (a synthesized widening), preserving nulls a `.nullable()`/`.nullish()` field genuinely allows, and leaving nulls under shapes the schema doesn't describe untouched.
- **`@tanstack/ai`** runs it inside the structured-output `validate` callback (`activities/chat`), before `parseWithStandardSchema`, using `convertSchemaToJsonSchema(outputSchema)` (no `forStructuredOutput`) as the nullability source of truth.

### Why core, not the adapter

By the time the schema reaches the adapter it's already been widened (`forStructuredOutput: true`), so the adapter can no longer tell a synthesized null from a genuine one. Only the activity layer still holds the original schema, so the normalization has to live there. This also fixes it for every adapter on the validated path at once, rather than per-adapter.

## Tests

- `@tanstack/ai-utils`: unit coverage for `undoNullWidening` — optional vs. nullable, mixed, nested nullable objects (via `anyOf`), array items, schemaless passthrough.
- `@tanstack/ai`: end-to-end `chat({ outputSchema })` regression — a provider `null` on an optional field now resolves (key absent), and a `null` on a `.nullable()` field is preserved.
- Full `@tanstack/ai` suite (1005 tests) + `tsc` + eslint pass.

## Out of scope / follow-up

There's a **symmetric** latent issue in the `@tanstack/openai-base` adapters: they blind-strip *all* nulls (`transformNullsToUndefined`), so a genuinely `.nullable()` **required** field that returns `null` is stripped and then fails validation as missing — the inverse of the bug fixed here.

| field | passthrough adapters (e.g. OpenRouter) | blind-strip adapters (openai-base) |
| --- | --- | --- |
| `optional` (`T \| undefined`) | ❌ before this PR → ✅ | ✅ |
| `nullable` required (`T \| null`) | ✅ | ❌ |

Unifying that is intentionally **not** in this PR: the openai-base null-strip runs inside `structuredOutputStream` (shared by the non-streaming Promise path *and* the streamed path, which re-emits the adapter chunk), so fixing it cleanly means either threading the original schema down to the adapters or normalizing the streaming chunk pipeline — a larger, more invasive change. Happy to follow up with that if you'd prefer a single unified pass; `undoNullWidening` is exported and ready to be reused there.

## Summary by CodeRabbit

* **Bug Fixes**
  * Engine now removes synthesized nulls so optional fields are omitted (not null), preventing validation errors.

* **New Features**
  * Added a schema-aware null-normalization utility and flow that tracks and strips only synthesized nulls while preserving genuine nullable values.
  * Adapters default to preserving provider nulls; engine applies consistent un-widening for both promise and streaming outputs.

* **Documentation**
  * Clarified streaming vs non-streaming validation and normalization behavior.

* **Tests**
  * Expanded unit and e2e tests for optional vs nullable handling, arrays/tuples, streaming, and nested cases.


DrewHoo:
Thanks @coderabbitai — all three addressed in 3f446789:

- **`resolveSchema` ambiguous branches**: now resolves only when exactly one non-null `anyOf`/`oneOf` branch matches the value's shape; ambiguous unions keep the original schema and stop descending, rather than risk stripping a `null` a sibling branch allows.
- **Tuple `items`**: array walking applies `items: [a, b, …]` per index instead of always using the first.
- **Test import order**: fixed — used a separate top-level `import type` rather than the suggested inline `type` specifier, since the repo's `import/consistent-type-specifier-style` rule requires the separate form.

Added unit coverage for both the tuple and ambiguous-union cases.

tombeckenham:
Pushed a refactor of the null-normalization internals (behaviour unchanged; all unit + integration tests green).

**What changed:** the previous approach re-derived *which* nulls were synthetic at validation time by walking the response value against the un-widened schema and pattern-matching `anyOf`/`oneOf` branches (`resolveSchema`), bailing to "leave the null" whenever a branch was ambiguous. That reconstruction was the awkward part — the widening pass already knows exactly what it widened, so the inverse can be recorded rather than guessed.

- `makeStructuredOutputCompatible` now returns `{ schema, nullWidening }`, where `nullWidening` is a `NullWideningMap` marking exactly the positions where it added `null` (and only there — a field already typed nullable via `.nullish()` is left unmarked).
- New `convertSchemaForStructuredOutput(schema)` exposes both; the chat activity threads the map straight into `undoNullWidening`, replacing the second (un-widened) `convertSchemaToJsonSchema` call.
- `undoNullWidening(value, map)` now consumes the map instead of a schema. `resolveSchema`/`allowsNull` and the `JsonSchemaNode` type are gone.

**Why it's better:**
- **Exact, not guessed** — the map *is* what the widener did; no value-vs-branch matching.
- **Drift-proof** — the record lives inside the widening function, so the two can't fall out of sync.
- **Closes a real gap** — the old ambiguity bailout (a union of object types) silently left synthesized nulls in place; with the map there's nothing to disambiguate. `.nullish()` nulls are now preserved by construction.

The `ai-utils` tests were rewritten to drive maps directly; the `chat-structured-output-null-normalization.test.ts` behavioural test is unchanged and still passes.

tombeckenham (commented):
Thanks for finding this @DrewHoo. This pointed me to an adjust bug in the open-ai base as well.

tombeckenham (approved):

DrewHoo:
> Thanks for finding this @DrewHoo. This pointed me to an adjust bug in the open-ai base as well.

@tombeckenham thanks for the update and thanks for taking a look! i'm working on the ci failure now

DrewHoo:
Fixed the `react-native-smoke` CI failure in 0c1be6c1. The smoke fixture resolves workspace packages to source via explicit per-tool mappings; since `@tanstack/ai` now pulls in `@tanstack/ai-utils` (chat activity + schema-converter), I registered it — mirroring `@tanstack/ai-event-client` — in the tsconfig `paths`, metro `packageEntryPoints`, the esbuild alias map, and the import-surface walker.

`pnpm --filter @tanstack/ai-react-native-smoke smoke` now passes locally end-to-end (typecheck, import surface (80 visits), expo export, esbuild).

AlemTuzlak (approved):

PR #749, opened 2026-06-11
fix(ai-client): settle status to ready on a message-less terminal run (#421)

valeriudev:
Fixes #421

## 🎯 Changes

`useChat` stayed in a non-ready status after a client tool call when the continuation run closed with a bare `RUN_FINISHED { finishReason: 'stop' }` and no assistant message.

Status only reaches `ready` through the processor's `onStreamEnd` callback, which `StreamProcessor.finalizeStream()` fires only when an assistant message exists — so a message-less terminal run left status stuck at `submitted`. The fix normalizes status to `ready` in `streamResponse`'s finally block on the terminal, non-continuing path, guarded on `status !== 'ready'` so the normal `onStreamEnd` path is a no-op. Kept in `ai-client` rather than the core processor: `onStreamEnd` is message-scoped by design (enforced by existing processor tests) and status is a client-owned concern.

Regression coverage in `chat-client-client-tool-status.test.ts`: the #421 repro, an exactly-once `ready` guard on the normal text path, and the first-run bare-`RUN_FINISHED` case.

## ✅ Checklist

- [x] I have followed the steps in the Contributing guide.
- [x] I have tested this code locally (`@tanstack/ai-client`: `test:lib`, `test:types`, `test:eslint`).

## 🚀 Release Impact

- [x] This change affects published code, and I have generated a changeset (`@tanstack/ai-client` patch).

### DIFF
--- a/.changeset/chat-client-tool-status-ready.md
+++ b/.changeset/chat-client-tool-status-ready.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/ai-client': patch
+---
+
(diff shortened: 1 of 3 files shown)


tombeckenham (approved):
Thanks for this.

PR #750, opened 2026-06-11
feat(ai-isolate): add native QuickJS Code Mode isolate driver for Bun

lithdew:
## 🎯 Changes

Adds a new Code Mode isolate driver, `@tanstack/ai-isolate-quickjs-bun`, that runs QuickJS natively on Bun via `bun:ffi` (through [`quickjs-bun`](https://github.com/superpowerdotcom/quickjs-bun)) rather than WebAssembly.

It implements the existing `IsolateDriver` contract from `@tanstack/ai-code-mode`, so it's a drop-in replacement for `@tanstack/ai-isolate-quickjs` on Bun servers:

```ts
import { createQuickJSBunIsolateDriver } from '@tanstack/ai-isolate-quickjs-bun'
import { createCodeModeTool } from '@tanstack/ai-code-mode'

const executeTypescript = createCodeModeTool({
  driver: createQuickJSBunIsolateDriver(),
  tools: [myTool],
})
```

### Why

On Bun, the existing WASM driver (`quickjs-emscripten`) is both slower and, in my testing, unreliable for async host tool calls. Its asyncify bridge crashes the shared WASM module (`memory access out of bounds`) and hangs once a single execution makes ≥ 4 sequential awaited tool calls (reproduced on Node 22 and Bun 1.3.14). `quickjs-bun` maps the QuickJS C API directly through `bun:ffi`, giving each context its own native runtime with no asyncify and no shared-VM serialization.

### Benchmarks

`packages/ai-isolate-quickjs-bun/benchmarks/compare-with-wasm.ts`, both drivers driven through the public `IsolateDriver` interface (Apple M-series, darwin/arm64; Bun 1.3.14 for the native driver, Node 22 for the WASM driver as its native habitat):

| Scenario (fresh context per run) | QuickJS Bun | QuickJS WASM (Node) |
| -------------------------------- | ----------: | ------------------: |
| Cold start (first context + run) |     ~130 ms |              ~20 ms |
| `return 1 + 1`                   |     ~0.7 ms |   ~10 ms (≈14× WASM)|
| 3 sequential tool calls          |     ~0.9 ms |          see note ¹ |
| 8 sequential tool calls          |     ~1.0 ms |          see note ¹ |
| compute (`fib(20)`)              |     ~5.0 ms |          see note ¹ |
| `return 1 + 1` (reused context)  |    ~0.04 ms |                   — |

¹ The WASM driver's asyncified host tool calls repeatedly crash the shared WASM module and hang subsequent executions (≥ 4 sequential awaited host calls per process), on both Node 22 and Bun 1.3.14. Sync-only workloads are unaffected. WASM wins cold start (one-time WASM instantiate vs TinyCC compile of the QuickJS sources); the native driver wins steady-state per-execution by ~14× on the trivial case.

### Notes

The suites are gated with `describe.skipIf(typeof Bun === 'undefined')`, mirroring how `@tanstack/ai-isolate-node` skips when its native addon is unavailable — would like to know if it is fine to add a `oven-sh/setup-bun` job (`pnpm --filter @tanstack/ai-isolate-quickjs-bun test:bun`) so that the full test suite is ran in CI.

## ✅ Checklist

## 🚀 Release Impact

## Summary by CodeRabbit

* **New Features**
  * Added a Bun-native QuickJS isolate driver for Code Mode (`@tanstack/ai-isolate-quickjs-bun`, drop-in for Bun) using `bun:ffi`, with configurable `timeout`, `memoryLimit`, `maxStackSize`, and `maxToolCalls` (default 1000) plus normalized limit/timeout error behavior.
* **Documentation**
  * Updated Code Mode and isolate-driver documentation, comparisons, READMEs, and skills to add the QuickJS Bun driver and guidance for choosing it on Bun servers.
* **Examples**
  * Updated the Code Mode web example to add a `quickjs-bun` driver option.
* **Tests & Benchmarks**
  * Added Bun-only isolation/escape and driver behavior tests, plus benchmarks comparing native Bun QuickJS vs WASM.


lithdew:
Rebased to the latest origin/main and fixed merge conflicts.

docs/code-mode/code-mode-isolates.md
```diff
@@ -2,31 +2,35 @@
...
   - isolate driver
   - isolated-vm
   - quickjs
+  - quickjs-bun
+  - bun
+  - bun:ffi
   - cloudflare workers
   - sandbox
   - secure execution
----
```
tombeckenham, line 19: **[bug]** The YAML frontmatter opening `---` (line 1) is never closed before the body starts at line 19. Every other Code Mode doc (`code-mode.md`, `client-integration.md`, etc.) closes frontmatter with `---` on its own line. The diff removed the closing delimiter. Parsers (e.g. `typedoc-plugin-frontmatter`) will treat lines 2–34—including the comparison table and headings—as frontmatter or mis-parse the page.

**Suggestion:** Restore the closing `---` immediately after the `keywords` block (after line 17), matching the other docs in `docs/code-mode/`.

docs/code-mode/code-mode-isolates.md
```diff
@@ -90,17 +96,61 @@ const driver = createQuickJSIsolateDriver({
...
 | `maxStackSize` | `number` | `524288` | Maximum call stack size in bytes (default: 512 KiB). Increase for deeply recursive code; decrease to catch runaway recursion sooner. |
 
+
 ### How it works
 
 QuickJS WASM uses an asyncified execution model — the WASM module can pause while awaiting host async functions (your tools). Executions are serialized through a global queue to prevent concurrent WASM calls, which the asyncify model does not support. Fatal errors (memory exhaustion, stack overflow) are detected, the VM is disposed, and a structured error is returned. Console output is captured and returned with the result.
 
-> **Performance note:** QuickJS interprets JavaScript rather than JIT-compiling it, so compute-heavy scripts run slower than with the Node driver. For typical LLM-generated scripts that are mostly waiting on `external_*` tool calls, this difference is not significant.
+> **Performance note:** QuickJS interprets JavaScript rather than JIT-compiling it, so compute-heavy scripts run slower than with the Node driver. For typical LLM-generated scripts that are mostly waiting on `external_`* tool calls, this difference is not significant.
```
tombeckenham: **[nit]** Broken inline-code formatting: `` `external_` ``* tool calls — the backtick closes before `_`, leaving a stray `*` and incorrect rendering.

**Suggestion:** Use `` `external_*` `` (or "external_* tool calls" without partial backticks).

docs/code-mode/code-mode-isolates.md
```diff
@@ -90,17 +96,61 @@ const driver = createQuickJSIsolateDriver({
...
+```
+
+### Options
+
+
+| Option         | Type     | Default  | Description                                                                                                                          |
+| -------------- | -------- | -------- | ------------------------------------------------------------------------------------------------------------------------------------ |
+| `memoryLimit`  | `number` | `128`    | Maximum heap memory for the QuickJS runtime, in megabytes.                                                                           |
+| `timeout`      | `number` | `30000`  | Maximum wall-clock time per execution, in milliseconds.                                                                              |
+| `maxStackSize` | `number` | `524288` | Maximum call stack size in bytes (default: 512 KiB). Increase for deeply recursive code; decrease to catch runaway recursion sooner. |
```
tombeckenham, line 147: **[suggestion]** The QuickJS Bun "Options" table documents `memoryLimit`, `timeout`, and `maxStackSize` but omits `maxToolCalls`, which is a first-class driver option (default 1000) documented in the package README, changeset, and `QuickJSBunIsolateDriverConfig`. Users reading the canonical isolate-driver doc will not discover this safety limit.

**Suggestion:** Add a `maxToolCalls` row to the Options table and mention it briefly in "How it works", consistent with `packages/ai-isolate-quickjs-bun/README.md`.

packages/ai-isolate-quickjs-bun/package.json
```diff
@@ -0,0 +1,64 @@
...
+  ],
+  "engines": {
+    "bun": ">=1.3.14"
+  },
+  "scripts": {
+    "build": "vite build",
+    "clean": "premove ./build ./dist",
+    "lint:fix": "eslint ./src --fix",
+    "test:build": "publint --strict",
+    "test:bun": "bun test ./tests",
```
tombeckenham, line 42: **[suggestion]** The substantive test suite (`tests/*.test.ts`, ~880 lines) is gated with `describe.skipIf(typeof Bun === 'undefined')` and only runnable via `test:bun`. Standard CI targets (`test:lib` / `test:pr`) use Vitest on Node and therefore exercise only the single "rejects createContext on Node.js" test. Escape-attempt, timeout, memory-limit, maxToolCalls, and concurrency behavior are unverified in CI.

I'm unsure whether we should add in bun to the main test suite. I'm thinking that you'll need a second CI workflow to run the bun tests separately. We'll need core approval to add that in.

examples/ts-code-mode-web/src/components/ToolSidebar.tsx
```diff
@@ -103,6 +103,12 @@ export const DEFAULT_ISOLATE_VM_OPTIONS: Array<IsolateVMOption> = [
     description: 'Lightweight JavaScript engine',
     available: true,
   },
+  {
+    id: 'quickjs-bun',
+    name: 'QuickJS Bun',
+    description: 'Native QuickJS engine (requires running the server with Bun)',
+    available: true,
```
tombeckenham, line 112: **[suggestion]** The new `quickjs-bun` VM option is marked `available: true` unconditionally. When the example server runs under Node (the default `pnpm` workflow), selecting it fails at `createContext` with a runtime error. The description warns about Bun, but the UI still presents it as available—unlike the Node driver path in `create-isolate-driver.ts`, which falls back to QuickJS on failure.

**Suggestion:** Gate `available` on server runtime (e.g. env flag or server-side probe), or add a `quickjs-bun` fallback/error message in `create-isolate-driver.ts` similar to the Node driver pattern.

packages/ai-isolate-quickjs-bun/package.json
```diff
@@ -0,0 +1,64 @@
...
+    "code-mode",
+    "quickjs",
+    "bun",
+    "ffi",
+    "isolate",
+    "sandbox",
+    "code-execution"
+  ],
+  "dependencies": {
+    "quickjs-bun": "0.1.2"
```
tombeckenham, line 62: **[suggestion]** Runtime behavior depends on `quickjs-bun@0.1.2`, a very early third-party package that compiles QuickJS via TinyCC/`bun:ffi`. The driver correctly isolates this with dynamic import and Bun version checks, but consumers inherit supply-chain and platform-support risk (Windows requires `QUICKJS_BUN_NATIVE_LIBRARY`).

**Suggestion:** Document minimum `quickjs-bun` compatibility in the README/changeset, pin with a narrow range once API stabilizes, and consider vendoring or upstreaming critical fixes if TanStack plans long-term support.

packages/ai-isolate-quickjs-bun/package.json
```diff
@@ -0,0 +1,64 @@
+{
+  "name": "@tanstack/ai-isolate-quickjs-bun",
+  "version": "0.0.0",
+  "description": "Native QuickJS sandbox driver for TanStack AI Code Mode TypeScript execution on Bun via bun:ffi.",
+  "author": "",
```
tombeckenham: **[nit]** Package metadata is incomplete compared to sibling drivers (`@tanstack/ai-isolate-quickjs`): empty `author`, no `homepage`/`bugs`/`funding`, and `version: "0.0.0"` (likely intentional pre-release, but inconsistent with published siblings at `0.1.x`).

**Suggestion:** Align metadata with other `@tanstack/ai-isolate-*` packages before first publish.

packages/ai-isolate-quickjs-bun/src/isolate-context.ts
```diff
@@ -0,0 +1,642 @@
...
+        // exhausted to allocate an Error object. (A literal `throw null` in
+        // sandbox code is indistinguishable and is treated the same way.)
+        if (value.type === 'null') {
+          return memoryLimitError(error.stack)
+        }
+        // A thrown plain object/array surfaces through quickjs-bun's
+        // JSException as the generic "QuickJS object was thrown"; recover its
+        // `message` (and `name`) so the model still gets useful feedback for
+        // self-correction — parity with the WASM driver.
+        if (value.type === 'object' || value.type === 'array') {
```
tombeckenham, line 620: **[bug]** Fatal OOM / stack-limit errors thrown as Error *objects* (not a bare `null`) are misclassified as non-fatal, so an exhausted VM gets reused.

This branch recovers `message`/`name` from a thrown object and **returns early — before** `normalizeError`'s fatal-limit heuristics run. QuickJS only throws a bare `null` on the *most* severe OOM (when it can't even allocate the Error); with some headroom it throws a real `InternalError: out of memory` **object**, and stack overflow surfaces as an `InternalError`/`RangeError` object. Those hit this branch, come back named `InternalError`, so `isFatalQuickJSLimitError()` returns `false`, `fail()` skips `releaseVmAfterFatalLimit()`, and the VM is driven on after a fatal limit. This matters more here than in the WASM driver: native FFI QuickJS has no containment boundary, so continuing on a heap-exhausted native VM is a worse failure mode.

**Suggestion:** recover the user-facing fields but still route them through `normalizeError` so classification is uniform:
```ts
const synthetic = new Error(message)
synthetic.name = this.readValueProp(value, 'name') ?? error.name
if (error.stack !== undefined) synthetic.stack = error.stack
return normalizeError(synthetic)
```
This also fixes a parity nit — a synchronous timeout tripped by `evalCode`'s `timeoutMs` interrupt currently reaches the model as `InternalError: interrupted` instead of a clean `TimeoutError`, for the same early-return reason. Worth confirming the object-vs-null OOM shapes under Bun, but the fix is safe either way.

packages/ai-isolate-quickjs-bun/src/isolate-driver.ts
```diff
@@ -0,0 +1,183 @@
...
+   */
+  maxToolCalls?: number
+}
+
+/**
+ * Create a QuickJS isolate driver for the Bun runtime
+ *
+ * This driver runs QuickJS natively through `bun:ffi` (via `quickjs-bun`)
+ * instead of WebAssembly. Each context gets its own QuickJS runtime with
+ * dedicated memory and stack limits, so sandboxes are fully isolated from
```
tombeckenham: **[suggestion]** "fully isolated ... from the host" overstates the boundary. Per-context heaps and scope isolation hold and are well-tested, but this driver runs QuickJS compiled by TinyCC and called through `bun:ffi` — **directly in the host process address space, with no sandbox boundary**. The memory "limit" is QuickJS's internal accounting (`JS_SetMemoryLimit`), not a hardware/OS/VM wall. Unlike Node `isolated-vm` (separate V8 isolate) or the WASM driver (WebAssembly linear-memory sandbox), a memory-safety bug in QuickJS or a mishandled FFI handle here is a host-process compromise, not a contained fault.

**Suggestion:** keep the accurate per-context-independence framing, but reword to state the trade-off honestly (native FFI trades WASM's/V8's containment boundary for speed — fits trusted-ish or otherwise-contained deployments, weaker fit where the sandbox wall itself is the security requirement). Same applies to the README's "with the same sandboxing guarantees" line — the guarantees are *not* the same, only the engine and scope surface are.

packages/ai-isolate-quickjs-bun/src/isolate-context.ts
```diff
@@ -0,0 +1,642 @@
...
+        deferred.dispose()
+        task.settle()
+      }
+    }
+
+    // The wrapper always passes a JSON string, but degrade gracefully if
+    // the handle dumps to something else.
+    let argsJson = '{}'
+    try {
+      const dumped = vm.dump(argsHandle ?? vm.undefined)
```
tombeckenham, line 518: **[suggestion]** When `vm.dump` **throws** here (sandbox heap exhausted while materializing the args string), the catch swallows the JSException and the code falls through with `argsJson = '{}'` — and then the async IIFE below still calls `binding.execute({})`. So a host tool runs with empty args instead of the caller's real args (a real side effect with the wrong inputs for something like `readFile`/`deleteRecords`), and the underlying OOM is never surfaced or classified.

The "dump returned a non-string" case degrading to `{}` is fine (the wrapper guarantees a JSON string). But the **throw** case should not invoke the binding — record it via `hostSettleError ??= this.toNormalizedError(error)` (or settle the sandbox promise with the classified error) so the execution loop surfaces it and releases the VM if it was fatal.

packages/ai-isolate-quickjs-bun/src/isolate-context.ts
```diff
@@ -0,0 +1,642 @@
...
+
+    await waitForPrev
+
+    // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition -- dispose() may run while awaiting the queue
+    if (this.disposed) {
+      release()
+      return this.disposedResult()
+    }
+
+    this.logs.length = 0
```
tombeckenham, line 189: **[suggestion]** Reused-context state bleed: `execute()` resets host-side counters here but nothing drains the QuickJS job queue between runs. After a timeout/abort, the previous program's un-run promise reactions stay queued; on the next `execute()` the loop's `executePendingJob()` runs those stale jobs first — attributing their wall-clock, tool-call budget, and `console.*` output to the new execution.

Latent today (the main consumer creates a fresh context per call and disposes it), but this class explicitly supports multi-`execute` — that's the whole point of the `execQueue` serialization + per-run reset. Either drain/reset the VM job queue on entry, or document that a context is single-execute.

packages/ai-isolate-quickjs-bun/src/isolate-context.ts
```diff
@@ -0,0 +1,642 @@
...
+/**
+ * Caps on captured console output. Logs are held on the host and flow back
+ * to the model, so an unbounded `while (true) console.log(...)` loop in
+ * untrusted code could grow host memory without ever tripping the sandbox
+ * heap limit (the sandbox only holds one string at a time). Once either cap
+ * is reached a single truncation marker is appended and further output is
+ * dropped for the rest of the execution.
+ */
+const MAX_LOG_ENTRIES = 10_000
+const MAX_LOG_BYTES = 1_000_000
```
tombeckenham: **[nit]** `MAX_LOG_BYTES` is compared against `msg.length` in `pushLog`, but `String.length` is UTF-16 code units, not bytes — while the constant name and the doc comment say "bytes". Worst-case multi-byte content lets the host log buffer reach ~3–4 MB before truncation. Still bounded, so low risk. Rename to `MAX_LOG_CHARS`, or measure real byte length, to match the stated intent.

examples/ts-code-mode-web/src/lib/create-isolate-driver.ts
```diff
@@ -19,6 +19,12 @@ export async function createIsolateDriver(
       driver = createQuickJSIsolateDriver()
       break
     }
+    case 'quickjs-bun': {
```
tombeckenham, line 77: **[suggestion]** The `quickjs-bun` driver case added here is effectively **unreachable end-to-end**, which makes the new sidebar option impossible to actually exercise:

1. **The server never passes the selected VM through.** All nine server call sites hardcode `createIsolateDriver('node')` (`api.execute-prompt.ts:31`, `api.product-codemode.ts:63`, `api.database-demo.ts`, `api.banking-demo.ts`, `api.reports.ts`, `api.report-event.ts`, `api.structured-output.ts`, `reports/*.ts`). Selecting "QuickJS Bun" (or anything) in `ToolSidebar` changes nothing — the server always builds the Node `isolated-vm` driver.
2. **There's no way to run the server under Bun.** `dev` is `vite dev` (Node), there's no `bun` script, and the README has no Bun instructions (its whole troubleshooting section is about `isolated-vm` on Node). But this driver requires the *server process* itself to be Bun.

So as shipped, the option is cosmetic. To make it usable: thread the selected `IsolateVM` from the request into `getDriver()` (keyed cache per VM), add a `"dev:bun"` script that runs the server under Bun, and document the Bun run path in the README. Otherwise the sidebar entry should be removed until the example can actually run it.

tombeckenham (changes requested):
It's good work generally. Thee's a few issues raised through the AI review you should address.

I've got 2 main concerns. I'm not sure about how we add the bun tests to the suite cleanly, and I couldn't properly get the example app to run in bun. Let me know how you'd address that.

jherr:
@tombeckenham — I went back through all of your open threads on this branch and mapped each to where it now lands, to make a re-review easier. The branch is up to date with `main` (trivial merge, `4d044f10`) and all reporting checks are green; the substantive Bun test suite is green locally (52 pass / 1 skip). Details below, grouped by kind.

## Substantive correctness fixes (in code + covered by tests)

**1. Fatal OOM/stack objects misclassified → exhausted VM reused** (`isolate-context.ts:518`→`:620` region, [r3528366410](https://github.com/TanStack/ai/pull/750#discussion_r3528366410))
The recovered-object branch no longer returns early before the fatal-limit heuristics. It now recovers the user-facing `name`/`message` and routes them **back through `normalizeError`** so classification is uniform (`isolate-context.ts:633`). A real `InternalError: out of memory` *object* (not just the bare-`null` OOM) is now classified fatal, so `fail()` releases the VM instead of driving it on. This also fixes the timeout-parity nit you flagged (a sync `evalCode` interrupt now surfaces as `TimeoutError`, not `InternalError: interrupted`).

**2. `vm.dump` throw ran the host tool with `{}`** (`isolate-context.ts:518`, [r3528366562](https://github.com/TanStack/ai/pull/750#discussion_r3528366562))
The **throw** case no longer falls through to `binding.execute({})`. It records the classified error via `this.hostSettleError ??= this.toNormalizedError(error)` (`:529`), disposes the deferred, and settles — so the OOM is surfaced and the VM released rather than a host side effect running with the wrong args. The benign "dump returned a non-string" degrade-to-`{}` path is unchanged, as you noted it should be.

**3. Reused-context job-queue bleed after timeout/abort** (`isolate-context.ts:189`, [r3528366636](https://github.com/TanStack/ai/pull/750#discussion_r3528366636))
`execute()` now drains the QuickJS job queue on entry before the run starts: `drainStaleJobs()` (`:186`/`:561`), a bounded loop (`MAX_STALE_JOB_DRAIN = 10_000`) that runs out any leftover promise reactions from a prior program so their wall-clock, tool-call budget, and `console.*` output can't be attributed to the new execution.
**Regression test added** — *"does not leak a timed-out run's queued jobs into the next execution"*: a first run queues `Promise.resolve().then(() => console.log('STALE_FROM_FIRST_RUN'))` then busy-loops into a timeout; the test asserts the second run's logs do **not** contain the stale marker. Verified it fails without the drain (`["SECOND_RUN","STALE_FROM_FIRST_RUN"]`) and passes with it.

## CI for the Bun-gated suite (`package.json:42`, [r3527371410](https://github.com/TanStack/ai/pull/750#discussion_r3527371410))

You were right that the ~880-line suite only ran the "rejects on Node.js" case under the Node PR job. I've added a **separate, path-filtered `Bun Tests` workflow** (`.github/workflows/bun-test.yml`): it installs Bun, builds the package's workspace deps via nx (`@tanstack/ai-code-mode` is a runtime peer dep imported as `wrapCode`), and runs `test:bun`. It's path-filtered to the package + `ai-code-mode` + the workflow file so it stays off unrelated PRs (keeping the runner-cost impact you raised contained). It currently shows `action_required` — that's the standard maintainer-approval gate on *all* workflows for a fork PR, not a config issue.

This is the one item I'd flag for a **core decision**: whether to (a) keep it as an informational check, or (b) make it a *required* status check for merge. I've wired it so either is a one-line settings change; I didn't presume to make it required.

## Example sidebar option (`ToolSidebar.tsx:111` + `create-isolate-driver.ts:22`, [r3527371416](https://github.com/TanStack/ai/pull/750#discussion_r3527371416), [r3528379467](https://github.com/TanStack/ai/pull/750#discussion_r3528379467))

Being straight with you here: I took the **minimal, honest-gating** route rather than the full end-to-end wiring. `create-isolate-driver.ts` now **falls back to the WASM QuickJS driver with a warning** when the server isn't running under Bun, so selecting the option degrades gracefully instead of failing at `createContext`, and the sidebar label says so.

I did **not** thread the selected `IsolateVM` through the nine hardcoded `createIsolateDriver('node')` server call sites, nor add a `dev:bun` script — so under the default `pnpm dev` (Node) the option still resolves to WASM rather than the native Bun driver end-to-end. That's a deliberate scope call to keep this PR focused on the driver itself. If you'd rather the example not present an option it can't fully exercise, I'm happy to either (a) fully wire it (per-VM `getDriver()` cache + `dev:bun` + README run path) as a follow-up, or (b) remove the sidebar entry until the example can run under Bun — your call.

## Docs, metadata, wording (all fixed)

- **Frontmatter / formatting** (`code-mode-isolates.md`, [r3527371395](https://github.com/TanStack/ai/pull/750#discussion_r3527371395), [r3527371402](https://github.com/TanStack/ai/pull/750#discussion_r3527371402)): restored the closing `---`; fixed `` `external_*` `` backticks.
- **`maxToolCalls` documented** ([r3527371406](https://github.com/TanStack/ai/pull/750#discussion_r3527371406)): added a row to the Options table (`code-mode-isolates.md:148`) and a mention in "How it works".
- **Supply-chain / pin** ([r3527371418](https://github.com/TanStack/ai/pull/750#discussion_r3527371418)): `quickjs-bun` pinned to exact `0.1.2` with a compatibility + Windows `QUICKJS_BUN_NATIVE_LIBRARY` note in the README and changeset.
- **Package metadata** ([r3527371422](https://github.com/TanStack/ai/pull/750#discussion_r3527371422)): added `author`/`homepage`/`bugs`/`funding` to match sibling drivers.
- **"fully isolated from the host" overstatement** ([r3528366488](https://github.com/TanStack/ai/pull/750#discussion_r3528366488)): reworded in both the driver doc-comment and README — it now states plainly that native FFI QuickJS runs in the host process address space (no OS/VM/WASM boundary; `JS_SetMemoryLimit` is internal accounting), and points to `ai-isolate-node`/WASM for a hard sandbox wall.
- **`MAX_LOG_BYTES` misnamed** ([r3528366746](https://github.com/TanStack/ai/pull/750#discussion_r3528366746)): renamed `MAX_LOG_CHARS` (`isolate-context.ts:55`) to match that it bounds UTF-16 code units.

---

**Net:** all the correctness threads are fixed with tests, the docs/metadata/wording nits are done, and the Bun suite now has CI. The two things that need *your* input rather than more code are (1) whether the Bun workflow should be a required check (needs core sign-off) and (2) the example-wiring scope call above. If those look reasonable, the last blocker is this review's `changes-requested` state — a re-review would clear it.

---

*Attribution: this analysis, the review-response fixes, and this CI workflow were done by **Claude (Claude Code)** working in Jack Herrington's (@jherr) environment. The original driver is @lithdew's. Jack has not personally reviewed this line-by-line.*

tombeckenham (approved):
I've created a way to run the ts-code-mode-web example with bun. Run pnpm dev:bun or bun dev:bun.

AlemTuzlak (approved):

PR #776, opened 2026-06-16
fix(ai-react): sync tools to ChatClient when tools prop changes

### LINKED ISSUE #775: @tanstack/ai-react: useChat does not update the underlying ChatClient when tools changes
### TanStack AI version

v0.29.0

### Framework/Library version

v0.15.5

The React `useChat` hook accepts a `tools` option and passes it to `ChatClient` during initial construction. However, if the `tools` array changes during the component lifecycle, the existing ChatClient is not updated.

This causes the client-side tool registry to become stale. Tools with same name can have different implementaion overtime. For eg, an App with "worspaces" or "projects" toggle can contain a tool like `updateTask`  which updates tasks in the current project and on project toggle that implementation will change due to their dependency on things like `projectId` or `workspaceId`

The core `ChatClient` already supports this via `client.updateOptions({ tools })` https://github.com/TanStack/ai/blob/2cbf4f80d0fb2c2d5d065101a408ed7ed53db8f5/packages/ai-client/src/chat-client.ts#L46 The fix surface seems small.

**Expected Behavior**
When `tools` changes on client-side(React), `useChat` should update the existing `ChatClient`:
```tsx
  useEffect(() => {
    client.updateOptions({
      tools: options.tools ?? [],
    })
  }, [client, options.tools])
```

**Reproducible Example**
A project management app with a persistent AI chat sidebar. The chat component stays mounted while the user switches between projects.

Steps - 
1. User opens Project A.
2. useChat creates a ChatClient with updateTaskTool closed over projectId = "A".
3. User switches to Project B.
4. React re-renders ProjectChat with projectId = "B".
5. A new updateTaskTool is created, now closed over projectId = "B".
6. But useChat does not call client.updateOptions({ tools }).

```tsx
function ProjectChat({ projectId }: { projectId: string }) {
  const updateTaskTool = useMemo(
    () =>
      updateTaskDef.client(async ({ taskId, status }) => {
        // Captures the current projectId
        await updateTaskInProject(projectId, taskId, status)

        return {
          ok: true,
          projectId,
        }
      }),
    [projectId],
  )

  const tools = useMemo(
    () => clientTools(updateTaskTool),
    [updateTaskTool],
  )

  const chat = useChat({
    tools,
  })

  return <ChatUI chat={chat} />
}
```

Shared Code, reproducible example would require API keys

_No response_

None

### Terms & Code of Conduct

citizen204:
## Summary

`useChat` initialises the `ChatClient` with the `tools` option passed at construction time, but never calls `client.updateOptions({ tools })` when the `tools` prop changes during the component lifecycle. This means the client-side tool registry becomes stale when callers swap tools — e.g. a project-management app where `updateTaskTool` closes over `projectId` and the user switches projects.

`ChatClient.updateOptions()` already accepts `tools` and rebuilds `clientToolsRef` when it receives them. The fix is to add `tools` to the existing `updateOptions` effect that already syncs `body`, `forwardedProps`, and `context`.

## Changes

- `packages/ai-react/src/use-chat.ts`: conditionally spread `options.tools` into the `updateOptions` call and add it to the dependency array, matching the existing pattern for `forwardedProps`.

Fixes #775

## Test plan

- Mount `useChat` with a `tools` array that closes over a value (e.g. `projectId`).
- Change `projectId` while the component stays mounted.
- Confirm the new tool implementation is used on the next LLM response (not the stale closure from mount time).
- Callers should `useMemo` their tools array (as shown in the issue) to avoid unnecessary re-registrations.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved chat option handling so updates to tool availability apply reliably and take effect immediately.
  * Refined option update behavior to reduce unintended re-runs when unrelated chat settings change.

### DIFF
--- a/packages/ai-react/src/use-chat.ts
+++ b/packages/ai-react/src/use-chat.ts
@@ -168,20 +168,29 @@ export function useChat<
     }
   }, [client])
 
+  // Sync each wire-payload slot in its own effect so an unrelated option
+  // changing doesn't re-run the others. `updateOptions` declares strict-optional
+  // fields and rejects explicit `undefined` under EOPT, so guard the optional
+  // slots before passing them.
   useEffect(() => {
-    // Conditional spread: `updateOptions` declares strict-optional
-    // fields and rejects explicit `undefined` under EOPT.
-    client.updateOptions({
-      body: options.body,
-      ...(options.forwardedProps !== undefined && {
-        forwardedProps: options.forwardedProps,
-      }),
-      ...(options.tools !== undefined && {
-        tools: options.tools,
-      }),
-      context: options.context,
-    })
-  }, [client, options.body, options.forwardedProps, options.tools, options.context])
+    client.updateOptions({ body: options.body })
+  }, [client, options.body])
+
+  useEffect(() => {
+    if (options.forwardedProps !== undefined) {
+      client.updateOptions({ forwardedProps: options.forwardedProps })
+    }
+  }, [client, options.forwardedProps])
+
+  useEffect(() => {
+    if (options.tools !== undefined) {
+      client.updateOptions({ tools: options.tools })
+    }
+  }, [client, options.tools])
+
+  useEffect(() => {
+    client.updateOptions({ context: options.context })
+  }, [client, options.context])
 
   useEffect(() => {
     if (options.live) {
--- a/packages/ai-react/src/use-chat.ts
+++ b/packages/ai-react/src/use-chat.ts
@@ -176,9 +176,12 @@ export function useChat<
       ...(options.forwardedProps !== undefined && {
         forwardedProps: options.forwardedProps,
       }),
+      ...(options.tools !== undefined && {
+        tools: options.tools,
+      }),
       context: options.context,
     })
-  }, [client, options.body, options.forwardedProps, options.context])
+  }, [client, options.body, options.forwardedProps, options.tools, options.context])
 
   useEffect(() => {
     if (options.live) {


packages/ai-react/src/use-chat.ts
```diff
@@ -176,9 +176,12 @@ export function useChat<
       ...(options.forwardedProps !== undefined && {
         forwardedProps: options.forwardedProps,
       }),
+      ...(options.tools !== undefined && {
+        tools: options.tools,
+      }),
       context: options.context,
     })
-  }, [client, options.body, options.forwardedProps, options.context])
+  }, [client, options.body, options.forwardedProps, options.tools, options.context])
```
manish-baghel, line 194: wouldn't it be better to have separate `useEffect` to avoid updating other fields repeatedly when one changes

the underlying code https://github.com/TanStack/ai/blob/2cbf4f80d0fb2c2d5d065101a408ed7ed53db8f5/packages/ai-client/src/chat-client.ts#L1439

just updates all fields present in the provided options.

With separate `useEffects`, updates in let's say `forwardedRef` wouldn't update the tools

manish-baghel: @coderabbitai

manish-baghel:
@AlemTuzlak 
I think other clients will also need this change. I originally didn't create a PR as I realized that pretty much all clients will need this update and I personally don't know anything other than React.

AlemTuzlak:
@manish-baghel i'll apply this to other frameworks and get it over the finish line, tysm!

manish-baghel:
Thanks @AlemTuzlak  and thanks @citizen204  for raising a PR. Really appreciate it~!

tombeckenham (approved):
Thanks for this and for the comment. I implemented the suggestion

PR #785, opened 2026-06-18
feat: add AI persistence and generation resumability

AlemTuzlak:
## Summary
- Adds generation artifact persistence on top of `withPersistence`, including input/output artifact refs, blob-backed media storage, `generation:artifacts` events, and custom extraction/name hooks.
- Adds generation run identity, replay, request parsing helpers, and lightweight client resume snapshots for generation hooks.
- Wires resumable generation APIs across React, Solid, Vue, Svelte, and Angular, with docs for resumable generations, artifacts, Cloudflare/R2, and hook API fields.

## Validation
- `pnpm test:docs`
- `git diff --check HEAD`
- Focused typechecks for `@tanstack/ai`, `@tanstack/ai-client`, `@tanstack/ai-persistence`, `@tanstack/ai-event-client`, React, Solid, Vue, Svelte, and Angular packages
- Focused generation/persistence/hook test suites for core, ai-client, ai-persistence, React, Solid, Vue, Svelte, and Angular

Full `pnpm test:pr` was not run per request; validation was scoped to the changed packages and docs.


willwillems:
Awesome to see that this is in progress, it was one of the final bits that prevented it from feeling completely batteries included!

docs/persistence/controls.md
```diff
@@ -0,0 +1,212 @@
...
+  adapter: anthropicText('claude-sonnet-4-6'),
+  messages: [{ role: 'user', content: 'Draft a release note.' }],
+})
+```
+
+There is no resume cursor, no server transcript, and no durable run record. This
+is enough for one-shot server work, tests, or prototypes where reload recovery
+does not matter.
+
+## Lever 2: client snapshots
```
tombeckenham: These levels are really difficult to understand. I've read it several times, and I'm still not sure when I'd use what level. It needs something visualy, like an svg. Perhaps even a hyperframes video. It's also unclear if I can use the native conversations ids like Google interactions and OpenAI responses. For example, I could use local storage on the browser and a Google interaction id if I just want to continue a conversation and for Google to retain the context

docs/persistence/overview.md
```diff
@@ -0,0 +1,135 @@
...
+
+declare const messages: MessageStore
+declare const runs: RunStore
+declare const publicEvents: PublicEventStore
+declare const internalEvents: InternalEventStore
+declare const interrupts: InterruptStore
+declare const metadata: MetadataStore
+declare const locks: LockStore
+declare const artifacts: ArtifactStore
+declare const blobs: BlobStore
```
tombeckenham: It would be useful in this doc to explain what each of these storages is for. A one liner against each

docs/persistence/drizzle.md
```diff
@@ -0,0 +1,67 @@
...
+events, interrupts, metadata, and other SQL-backed stores while keeping schema
+deployment in your existing Drizzle workflow.
+
+## Generate the migration
+
+Create a Drizzle migration file for your dialect.
+
+```sh
+pnpm exec tanstack-ai-persistence-drizzle --dialect postgres
+```
```
tombeckenham: This is all wrong.... we should never be generating migrations directly. We should instead generate schemas in typescript. if we generate the migrations, the schemas will be out of sync at all times. Look at how better-auth works

tombeckenham: Also Drizzle now has 2 different migration formats, and different types of migrations that depend on what db is used behind it.

tombeckenham:
## Scalability note: per-chunk event persistence on D1, and the path forward

Wrote up a load analysis of the durable-replay hot path with the Cloudflare backend. TL;DR: the current per-chunk write path falls over well before quotas matter — the fixes are a batched append seam in the store contract and, longer term, a DO-per-thread hot tier.

### What the hot path costs today

With `cloudflarePersistence(...)`, durable replay is **on by default** (`defaultFeatures` enables it whenever a `publicEvents` store exists — `packages/ai-persistence/src/middleware.ts:99-101`), and `onChunk` awaits an append for **every AG-UI event** — every text delta (`middleware.ts:975`). D1 has no interactive transactions, so the driver's `transaction()` is a pass-through (`packages/ai-persistence-cloudflare/src/index.ts:382-384`), which makes the CAS append (`packages/ai-persistence-sql/src/stores.ts:206-244`) **4 sequential D1 round trips per chunk**: read-at-target, `SELECT MAX(seq)`, INSERT, read-back. A typical streamed response is ~200-300 events.

### Which limits break, in order (100 concurrent users)

1. **Per-invocation subrequest cap — breaks on a single long response.** D1 queries count as subrequests (1,000 paid / 50 free). At 4 queries/chunk, any response past ~250 chunks dies mid-stream with "Too many subrequests". Needs zero concurrency to trigger.
2. **Write serialization — breaks with a handful of concurrent streams.** One D1 database is single-writer SQLite. 100 streams × ~30-50 deltas/sec ≈ 3-5k appends/sec (12-20k queries/sec) against a store realistically good for hundreds to ~1k serial writes/sec. Because `onChunk` awaits inline, the queue backpressures every user's stream.
3. **Free-tier daily writes — minutes.** 100k rows/day ÷ ~250 rows/message ≈ 400 messages total.
4. **10 GB database cap — weeks.** ~200-500 B/event; there's no pruning/retention API, and `public_events` grows forever, including for completed runs whose messages are already durably saved.
5. **Paid monthly write quota — actually fine.** ~$1/M rows overage; cost is not the constraint.

### Recommendations

**1. Add `appendBatch` to `PublicEventStore` + a buffering decorator.** Buffer N events (usually-sync `onChunk`), flush on count / boundary events (`RUN_FINISHED`, interrupt outcomes) / terminal drain in `onFinish`/`onError`/`onAbort`, with a high-water mark that degrades to per-chunk awaiting when the backend is slow. Each backend gets its atomic verb: `db.batch()` on D1, multi-row `INSERT` on Postgres/MySQL, `TransactWriteItems` on DynamoDB (note: `BatchWriteItem` can't carry the CAS condition), pipelined `XADD` on Redis. At N=25 a 300-chunk response drops from ~1,200 D1 subrequests to ~12. Keeping one row per event preserves the idempotent re-append/retry safety the stores already have.

**2. Adopt clamp-and-replay resume semantics.** With batching, a crashed run can leave a client cursor up to N−1 events past `latestSeq`; `resume-source.ts` currently throws on beyond-latest. The portable contract should be "resume from `min(clientSeq, latestSeq)`" — the client harmlessly re-receives a few events it already rendered. This matters most on platforms without `waitUntil` (Lambda, Cloud Run), where a hard kill can eat a buffered tail.

**3. Offer a `waitUntil` escape hatch** — e.g. `withPersistence(p, { waitUntil: ctx.waitUntil })` — so the final flush can ride past the response on CF/Vercel/Netlify-edge. Default stays drain-before-close, which is correct everywhere (Lambda freezes pending promises; Cloud Run throttles CPU post-response). The writer must be per-run, never module scope (Vercel Fluid / Lambda reuse).

**4. Longer term on Cloudflare: DO-per-thread hot tier.** Running the stream inside a per-thread Durable Object makes appends local synchronous SQLite writes (zero subrequests, no CAS needed — single-writer by construction, `locks` becomes redundant), and WebSocket hibernation provides the `stream.publish` fan-out natively. On `RUN_FINISHED`, one batched flush of messages/run status/interrupts to D1 or Postgres-via-Hyperdrive (three orders of magnitude less traffic — comfortably inside D1's envelope), raw event log optionally archived to R2 (~$0.015/GB-mo vs 10-50× that for DO/D1) or just dropped, then the DO prunes itself via alarm. The store contracts already split cleanly along this line; the missing piece is a DO-backed `PublicEventStore` with the completion-time flush.

**5. Docs callout in the meantime:** `features: ['messages', 'interrupts']` disables durable replay while keeping thread persistence — worth stating prominently in `docs/persistence/cloudflare.md`, since the default silently opts every D1 user into the per-token write path. A retention/pruning story for `public_events` is needed regardless.

Off-Cloudflare, for reference: a single Postgres instance absorbs this workload with an order of magnitude of headroom (the `ai-persistence-postgres` backend is the boring right answer); Redis Streams in front of Postgres is the optimal split if live-tail fan-out becomes a product feature.

tombeckenham:
Delivery durability (resumable streams) has been split out of this branch into its own PR: #955. This PR is now state persistence only — middleware, store contract, backends, and interrupt resume. The two features share no code; see #955 for the transport-level StreamDurability seam, `@tanstack/ai-durable-stream`, and the resumable-SSE client.

https://claude.ai/code/session_01A6Arc9bWdLq1aRRnDRCLMt

tombeckenham:
Superseded by the split this branch was reference material for. The work now lands as three stacked PRs:

- #984 `feat/persistence-core` → `main` — server persistence (`@tanstack/ai-persistence`) + client browser-refresh durability
- #987 `feat/generation-persistence` → #984 — generation persistence (client snapshot + durable media bytes)
- #988 `feat/persistence-sandbox` → #984 — sandbox persistence (SandboxStore + shared lock)

Three things changed materially versus this PR, so it is not a straight re-cut:

- **No backend packages.** `-drizzle`, `-prisma`, and `-cloudflare` are not published. A backend package has to own DDL, which puts it in permanent conflict with the migration tooling the consuming app already runs. `@tanstack/ai-persistence` ships the store contracts and a conformance testkit; per-stack agent skills read your existing ORM config and write a `chat-persistence.ts` against it.
- **Locks are not a store.** `AIPersistence.stores` accepts exactly four keys and throws on anything else; coordination is wired separately with `withLocks`.
- **The middleware no longer builds `ChatResumeToolState`.** The shipped ephemeral engine reconstructs resume state from the resume batch plus interrupt bindings in the server-loaded history; persistence records interrupts and gates new input.

Closing so the stack is the single live thread. The branch stays put if anything needs pulling from it.

PR #807, opened 2026-06-23
fix(ai-sandbox-cloudflare): make the Claude Code edge agent actually stream

tombeckenham:
## Summary

The Cloudflare sandbox Claude Code agent never started its turn — runs sat at `status:running`, `lastSeq:-1` forever and streamed nothing. Diagnosed live against `wrangler dev` (Docker Desktop) down to two stacked root causes, both fixed:

1. **`@tanstack/ai-claude-code`** — the adapter defaults `--permission-mode bypassPermissions`, which Claude Code maps to `--dangerously-skip-permissions` and **refuses to run as root**. Sandbox containers (Docker/Cloudflare) run as root, so `claude` died instantly. The adapter now sets **`IS_SANDBOX=1`** in the CLI env (Claude Code's documented escape hatch for skip-permissions in an isolated environment), merged over any caller-provided env.
2. **`@tanstack/ai-sandbox-cloudflare`** — `spawn()` used `@cloudflare/sandbox`'s background-process API (`startProcess` + `streamProcessLogs`), whose `onOutput`/`onExit` callbacks **never fire** (verified: the process runs and exits cleanly, yet zero log events arrive), so the stdout-NDJSON harness hung forever. `spawn()` now streams over **`exec({ stream: true, onOutput })`** — the same proven command path as one-shot `exec` — and resolves the exit code from its result. The caller's `AbortSignal` is **no longer forwarded** across the Durable Object RPC boundary (Workers RPC cannot serialize an `AbortSignal` — `"AbortSignal serialization is not enabled"` — which previously threw *before* the command ran and masked the failure as `status:done, lastSeq:-1`). A failed command now rejects `wait()` so the adapter surfaces a `RUN_ERROR` rather than a silent zero-output run.

Mid-run cancellation is therefore unavailable on the Cloudflare provider (documented in the code; a stuck run is bounded by the coordinator watchdog and the DO lifecycle).

This PR also wires the **Sandbox Agent** page into the `ts-code-mode-web` example (nav link, route, POST-then-WebSocket → SSE proxy, README section) and bumps the reference worker's deploy config (`max_instances`, `PUBLIC_HOSTNAME`).

## Test plan

- [x] `pnpm test:lib` — `@tanstack/ai-sandbox-cloudflare` (34, +1 `wait()`-rejects-on-failure test) and `@tanstack/ai-claude-code` (34, +1 `IS_SANDBOX=1` assertion) green
- [x] `pnpm test:types` + `pnpm test:eslint` green on both packages
- [x] **Live end-to-end** (`wrangler dev`, Docker Desktop): `POST /runs` → WS replay shows `RUN_STARTED` → `TEXT_MESSAGE_CONTENT` deltas (`hello from docker desktop`) → `RUN_FINISHED` (stop), `status:done`
- [ ] No CI runtime coverage for the Workers/container runtime — unit tests mock the SDK; live verification was manual


tombeckenham:
This is a duplicate

PR #829, opened 2026-06-24
ci: keep npm publish and GitHub releases coupled, self-heal release gaps

AlemTuzlak:
## Problem

GitHub releases stopped being created after **2026-06-15** (`@tanstack/ai-react@0.15.5` was the last one), even though npm kept receiving new versions through CI (`ai-react` is now at `0.15.12`).

Root cause: GitHub releases are created by **one thing only** — the `changesets/action` step in `release.yml` — and that step has not completed cleanly on any release-triggering run since mid-June:

- **Jun 16–23:** every `ci: Version Packages` merge run (#773, #778, #787, #792, #808, #813) passed **Run Tests** and then **failed at the "Run Changesets" step**. CI ran `changeset publish` and **npm advanced**, but the step died **before the tag-push / GitHub-release phase** — so no git tags (`0.15.6`–`0.15.12` don't exist) and no releases were created.
- **Jun 24:** runs now die even earlier, at the **test gate** — `root:test:kiira` (added to `test:ci` in #805) fails, so the changesets step is never reached.

(The exact in-step error from the Jun 16–23 runs is no longer recoverable — those runs' logs have expired — but the structural signature is clear: test gate green, changesets step red, npm published, no tags, no releases.)

## What this PR does

- **Split into `test` (gate) + `release` (`needs: test`) jobs.** A flaky run blocks **both** npm and GitHub releases together — never one without the other.
- **Tightened permissions** — top-level `contents: read`; write scoped to the `release` job.

## What this PR does NOT do

It does **not** diagnose or fix why the changesets step itself was failing after publishing to npm (logs expired). If that step still fails mid-way, npm and GitHub can still drift — this PR relies on the changesets step succeeding. The next post-merge release run is the place to capture its `Run Changesets` log before it expires.

## Follow-ups (not in this PR)

- **Backfill** the missing releases (`0.15.6`–`0.15.12`, etc.) as a one-off, with accurate per-version commits.
- **Diagnose** the changesets-step failure from a fresh run's logs.

## Testing

Workflow changes can only be fully exercised on `main`. Validated locally: YAML parses and the job graph is `test → release`.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -12,13 +12,11 @@ env:
   NX_CLOUD_ACCESS_TOKEN: ${{ secrets.NX_CLOUD_ACCESS_TOKEN }}
 
 permissions:
-  contents: write
-  id-token: write
-  pull-requests: write
+  contents: read
 
 jobs:
-  release:
-    name: Release
+  test:
+    name: Test
     if: github.repository_owner == 'TanStack'
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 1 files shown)


AlemTuzlak:
Thanks @coderabbitai — addressing both findings:

**Tag pointing at the wrong commit** — fixed. The tag now pins explicitly to the release commit (`git tag "$tag" "$GITHUB_SHA"`) instead of relying on implicit `HEAD`. In this step's scope that's the correct commit by construction: the step only runs on an actual release attempt (`if` gates on `published == 'true'` or the changesets step failing), and the workflow is checked out at the commit that bumped the version and triggered the publish. So `$GITHUB_SHA` *is* the commit that produced the published version — no drift.

**Only the current version is backfilled, not earlier missing ones** — intentional, not fixing here. This step's job is the coupling guarantee for the **current** release (publish → release, in one run), not a historical sweep. Auto-backfilling every earlier npm version would require resolving the correct commit *per old version*, which can't be reliably reconstructed after the fact — synthesizing those tags at the release commit would create a pile of wrong-commit tags across ~30 packages. Restoring the existing `0.15.6`–`0.15.12` gap is therefore a deliberate one-off (noted in the PR description as a follow-up), done with accurate per-version commits rather than guessed ones.

KevinVandy (approved):

AlemTuzlak:
Update: removed the self-heal/backfill step entirely. The two findings above were both about that step, so they're now moot. The PR is reduced to the structural changes: test/release job split (gated together), explicit GITHUB_TOKEN on the changesets step, and least-privilege permissions.

KevinVandy (approved):

PR #835, opened 2026-06-24
fix(ai-grok): remove Record<string,unknown> index sig from GrokTextProviderOptions

### LINKED ISSUE #821: bug(ai-grok): grokSummarize is not assignable to summarize()'s adapter param for any current model
## Summary

`grokSummarize(...)` (and `createGrokSummarize(...)`) is **not assignable to `summarize()`'s `adapter` param for any current Grok model** (`grok-4.3`, `grok-build-0.1`). This is a real type regression introduced by `main`'s provider-options rewrite; it was never caught because the only call sites live in `testing/**`, which CI excludes from `test:types` (see #820).

## Repro

```ts
import { summarize } from '@tanstack/ai'
import { grokSummarize } from '@tanstack/ai-grok/adapters'

summarize({ adapter: grokSummarize('grok-4.3'), text: 'x' })       // ❌ type error
summarize({ adapter: grokSummarize('grok-build-0.1'), text: 'x' }) // ❌ type error
```

```
Type 'ChatStreamSummarizeAdapter<"grok-4.3", GrokTextProviderOptions>' is not assignable to
  'SummarizeAdapter<string, object>'.
  Types of property 'summarize' are incompatible.
    Type '(options: SummarizationOptions<GrokTextProviderOptions>) => Promise<SummarizationResult>'
      is not assignable to '(options: SummarizationOptions<object>) => Promise<SummarizationResult>'.
        ... Type 'SummarizationOptions<object>' is not assignable to 'SummarizationOptions<GrokTextProviderOptions>'.
          Type 'object' is not assignable to type 'GrokTextProviderOptions'.
```

OpenAI (`openaiSummarize('gpt-5.2')`) passes — confirming this is Grok-specific.

## Root cause

`SummarizeAdapter.summarize` is declared as a **property** (arrow-function type), not a method shorthand (`packages/ai/src/activities/summarize/adapter.ts:49`). Under `strictFunctionTypes`, property-position function params are checked **contravariantly**. So assignability to the constraint `SummarizeAdapter<string, object>` (the bound on `summarize<TAdapter extends SummarizeAdapter<string, object>>`) reduces to:

> is `object` assignable to `TProviderOptions`?

For Grok, `TProviderOptions` is `GrokTextProviderOptions`, which **extends `Record<string, unknown>`** (`packages/ai-grok/src/text/text-provider-options.ts:35-36`):

```ts
export interface GrokTextProviderOptions
  extends GrokBaseOptions, Record<string, unknown> { ... }
```

`object` is **not** assignable to `Record<string, unknown>` (an index-signature type), so the check fails. `grok-build-0.1` resolves to `GrokBuildProviderOptions = Omit<GrokTextProviderOptions, 'reasoning'> & { reasoning?: never }`, which retains the index signature and fails the same way.

OpenAI's options (`ExternalTextProviderOptions`) are a pure intersection of interfaces with **all-optional named props and no index signature**, and `object` *is* assignable to an all-optional object type — which is why OpenAI passes. So the index signature on the Grok options is exactly what breaks it; it is not present on the other providers.

### Minimal isolation

```ts
const r1: Record<string, unknown> = {} as object  // ❌ object not assignable to index-sig type
const r2: { a?: number }          = {} as object  // ✅ object assignable to all-optional type
```

## Fix direction

Make Grok's options match the all-optional, no-index-signature shape the other providers use:

- Drop `extends ... Record<string, unknown>` from `GrokTextProviderOptions`; keep only the named optional props.
- Re-check `GrokBuildProviderOptions`'s `Omit<..., 'reasoning'>` — with the index signature gone, `Omit` will preserve the named keys correctly instead of collapsing them into `string`.

Confirm afterward that per-model provider-options narrowing (the `reasoning?: never` distinction for `grok-build-0.1`) still holds.

## Test coverage

- Add a call-site type assertion in `packages/ai-grok/tests/` (an *included* package) so this can never regress silently again:
  ```ts
  summarize({ adapter: grokSummarize('grok-4.3'), text: '' })
  summarize({ adapter: grokSummarize('grok-build-0.1'), text: '' })
  ```
  Constructing the adapter (current coverage) is not enough — the constraint only instantiates at the `summarize()` call site.

## Related

- #820 — CI gap that allowed this to slip through (`test:types` excludes `examples/**` and `testing/**`).

citizen204:
## Summary

`grokSummarize('grok-4.3')` and `grokSummarize('grok-build-0.1')` are not assignable to `summarize()`'s `adapter` parameter, producing a type error at every call site.

Fixes #821

## Root cause

`GrokTextProviderOptions` extends both `GrokBaseOptions` **and** `Record<string, unknown>`. Under `strictFunctionTypes`, function parameters are checked contravariantly. The constraint on `summarize<TAdapter extends SummarizeAdapter<string, object>>` means the adapter's `summarize` method must accept `SummarizationOptions<object>`. Contravariance requires `object` to be assignable to `GrokTextProviderOptions`. But `object` is **not** assignable to `Record<string, unknown>` (index-signature types require a compatible index signature from the source), so the check fails.

OpenAI's provider options have no index signature and all-optional named fields, so they pass: `object` satisfies any all-optional interface.

## Fix

Remove `Record<string, unknown>` from the `extends` clause of `GrokTextProviderOptions`. All fields are explicitly typed optional members — the index signature was not needed. After this change, `object` is assignable to `GrokTextProviderOptions` (all-optional, no index signature), fixing the contravariant check.

`GrokBuildProviderOptions` (`Omit<GrokTextProviderOptions, 'reasoning'>`) is not affected: the named fields are preserved correctly by `Omit` since there is no longer an index signature to collapse.

The `TProviderOptions extends Record<string, unknown>` constraint on `GrokTextAdapter` remains satisfied: a named-property interface with `unknown`-compatible values is structurally assignable to `Record<string, unknown>`.

## Changes

- `packages/ai-grok/src/text/text-provider-options.ts`: Remove `Record<string, unknown>` from `GrokTextProviderOptions` extends clause
- `.changeset/fix-grok-provider-options-index-sig.md`: Patch changeset for `@tanstack/ai-grok`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved TypeScript type compatibility for Grok summarization adapters, eliminating strict type errors at `summarize()` call sites.
  * Updated Grok text provider options to rely only on explicitly defined fields, removing the overly broad option shape that caused mismatches under strict function typing.

### DIFF
--- a/.changeset/fix-grok-provider-options-index-sig.md
+++ b/.changeset/fix-grok-provider-options-index-sig.md
@@ -1,0 +1,5 @@
+---
+"@tanstack/ai-grok": patch
+---
+
+Remove `Record<string, unknown>` index signature from `GrokTextProviderOptions` so that `grokSummarize` adapters are assignable to `SummarizeAdapter<string, object>`. Under `strictFunctionTypes`, the index signature caused `object` to be un-assignable to `GrokTextProviderOptions` (contravariant parameter check), making `grokSummarize('grok-4.3')` a type error at every `summarize()` call site. All fields on `GrokTextProviderOptions` are explicitly typed optional members, so the index signature was unnecessary.
--- a/packages/ai-grok/src/text/text-provider-options.ts
+++ b/packages/ai-grok/src/text/text-provider-options.ts
@@ -32,8 +32,7 @@ export interface GrokBaseOptions {
  * Grok-specific provider options for text/chat
  * Based on xAI Responses API options
  */
-export interface GrokTextProviderOptions
-  extends GrokBaseOptions, Record<string, unknown> {
+export interface GrokTextProviderOptions extends GrokBaseOptions {
   /**
    * Temperature for response generation (0-2)
    * Higher values make output more random, lower values more focused


AlemTuzlak (approved):

AlemTuzlak:
Thank you so much for this, Tom has fixed it upstream!

Issue #605, opened 2026-05-21
Route structured-output through native combined-mode where supported (skip finalization round-trip)


tombeckenham:
## Background

PR #600 added middleware coverage to the structured-output finalization step. That step exists because, historically, providers couldn't combine `tools` with schema-constrained output in a single call — so the engine runs the agent loop with `tools`, then makes a separate finalization call against the structured-output adapter to get the typed answer.

That premise has changed for most providers in the last six months.

## Provider landscape (May 2026)

| Provider | Native streaming + structured output + tools in one call? | Notes |
|---|---|---|
| OpenAI Chat Completions (gpt-4o-2024-08-06+) | Turn-based | Schema applies on the natural final turn; no separate call needed |
| OpenAI Responses API (gpt-5, o-series, 4.1) | Single-pass interleaved | True interleaved stream — `function_call` + schema-constrained `message` items in one SSE |
| Anthropic (Claude 4.5/4.6/4.7) | Single-pass interleaved | **GA 2026-01-29** via `output_config.format` + `tools` together |
| Google Gemini 3.x | Single-pass interleaved | Docs explicitly support combined mode (Gemini 3 only) |
| Google Gemini 2.x | Unsupported / brittle | Treat as unsupported |
| xAI Grok 4.x family (`grok-4.3`, `grok-4.20-*`) | Yes | Docs explicitly: "structured outputs + tools" gated to Grok 4 family. Tool-call args arrive atomically in one SSE chunk (not token-streamed). |
| xAI Grok 2/3 | One-or-the-other | Schema OR tools, not both |
| Groq | **API-rejected** | Docs explicit: *"Streaming and tool use are not currently supported with Structured Outputs."* Request returns 400. Same family as Ollama. |
| Ollama | Sampler conflict | GBNF schema mask suppresses tool-call prefix tokens (issue [ollama-python#546](https://github.com/ollama/ollama-python/issues/546)) |
| OpenRouter | Depends on upstream | Routes to an underlying provider; combined-mode support inherits from whichever model is selected |

Net: the finalization round-trip is **strictly overhead** for modern OpenAI / Anthropic 4.5+ / Gemini 3+ / Grok 4+ adapters, and load-bearing for **Groq, Ollama, Gemini 2.x, Grok 2/3, and legacy model versions**.

[...]

PR #836, opened 2026-06-24
feat(ai-openrouter): per-request native combined tools + outputSchema mode

### LINKED ISSUE #612: feat(ai-openrouter): per-request native combined tools + outputSchema mode (extends #605)
Follow-up to #605, which landed native combined mode (\`supportsCombinedToolsAndSchema()\`) for OpenAI (Chat Completions + Responses), Anthropic Claude 4.5+, Gemini 3.x, and the Grok 4 family. OpenRouter was deliberately deferred because it has a fundamentally different design problem from the others.

## What native combined mode buys

\`chat({ outputSchema, tools, stream: true })\` against a capable adapter wires the schema into the regular \`chatStream\` request alongside \`tools\` — the agent loop's final-turn text is the schema-constrained JSON, the engine harvests it from accumulated content, and the separate finalization adapter call is skipped. See #605 for the engine plumbing.

## Why OpenRouter is harder than the others

The other native-combined adapters declare capability statically (or via a model-meta set check on \`this.model\`). OpenRouter is a routing layer — a single OpenRouter adapter instance can route to many upstream models (\`anthropic/claude-opus-4.6\`, \`openai/gpt-5.2\`, \`x-ai/grok-4.1-fast\`, etc.), each with its own combined-mode capability. The capability for an OpenRouter call depends on which upstream model is resolved at request time, not on the adapter.

## Design choices to consider

1. **Static model-string prefix check.** The \`supportsCombinedToolsAndSchema(modelOptions)\` method receives \`modelOptions\` from the call site. The OpenRouter adapter inspects \`this.model\` (e.g. \`anthropic/claude-opus-4.6\`) plus any \`modelOptions.model\` / \`modelOptions.variant\` overrides and matches against a known-good prefix list:
   - \`anthropic/claude-sonnet-4-5\` and later
   - \`anthropic/claude-opus-4-5\` and later
   - \`openai/gpt-4o-2024-08-06\` and later
   - \`openai/gpt-5*\`, \`openai/o3*\`, \`openai/o4*\`
   - \`google/gemini-3*\`
   - \`x-ai/grok-4*\`
   
   Pro: stays adapter-local, no cross-package coupling. Con: duplicates the upstream gating that's already in each native adapter's own model-meta set.

2. **Delegate to upstream adapter capability sets.** OpenRouter imports the model-meta sets from \`@tanstack/ai-anthropic\`, \`@tanstack/ai-gemini\`, etc., strips the \`upstream/\` prefix from the model id, and checks each set in turn. Stays in sync automatically but creates dependency edges OpenRouter probably wants to avoid.

3. **Per-model registry on \`OpenRouterTextAdapter\`.** A static map keyed by OpenRouter's catalog model ids. Maintained inside \`ai-openrouter/src/model-meta.ts\` alongside the existing catalog. Pro: explicit and inspectable. Con: yet another list to keep current as new upstream models land.

Suggested: (3) is the most idiomatic match for how OpenRouter already organizes its catalog. The \`supportsCombinedToolsAndSchema(modelOptions)\` method inspects \`this.model\` (and \`modelOptions.model\` override if present) against a \`OPENROUTER_COMBINED_TOOLS_AND_SCHEMA_MODELS: Set<string>\` exported from model-meta.

## Wiring

Both OpenRouter adapter variants need updating:

- \`OpenRouterTextAdapter\` (chat-completions endpoint): when \`options.outputSchema\` is set AND the resolved model is in the set, attach \`response_format: { type: 'json_schema', json_schema: { strict: true, schema } }\` to the upstream request alongside \`tools\`. Skip when not in the set so the engine takes the legacy finalization path.
- \`OpenRouterResponsesTextAdapter\` (Responses-beta endpoint): same pattern but with \`text.format\` instead.

## Acceptance

- \`OPENROUTER_COMBINED_TOOLS_AND_SCHEMA_MODELS\` exported from \`@tanstack/ai-openrouter/model-meta\`, listing the canonical OpenRouter model ids that support combined mode (sourced from the underlying upstream-model gates that #605 / its follow-ups maintain).
- Both adapters' \`supportsCombinedToolsAndSchema(modelOptions)\` honors the set.
- Both adapters' \`mapOptionsToRequest\` (or equivalent) attaches the schema field when set.
- Unit test mirroring the per-adapter test pattern in #605 (verify request payload shape on the combined path; verify the legacy path is taken for unsupported upstream models).
- E2E: extend existing structured-output coverage to exercise OpenRouter with at least one combined-mode-capable upstream model.

season179:
Closes #612 (extends #605).

## What changed

This adds native OpenRouter combined mode for `tools + outputSchema` in both text adapters:

- Chat Completions now sends `responseFormat: { type: 'json_schema', jsonSchema: { strict: true, schema } }` on the combined path.
- Responses Beta now sends the schema through `text.format` on the combined path.
- Models that are not known to support native combined mode continue to use the existing finalization path.

Because OpenRouter is a routing layer, the capability check is intentionally conservative. The adapter now checks the primary model plus any `modelOptions.models` fallbacks and only enables combined mode when every possible routed model is in the combined-capable set. `:variant` suffixes such as `:nitro` are still ignored for capability purposes because they are routing directives, not different upstream model capabilities.

The new `OPENROUTER_COMBINED_TOOLS_AND_SCHEMA_MODELS` set is exported from `@tanstack/ai-openrouter/model-meta`, as requested in #612. The list is derived from the upstream provider gates rather than the broader OpenRouter `responseFormat` flag, so older structured-output-capable models that do not support native tools + schema stay on the legacy path.

## Tests

Added OpenRouter unit coverage for:

- combined-capable and unsupported models on both adapters
- chat `responseFormat` payload
- Responses `text.format` payload
- fallback routing disabling combined mode when any fallback is unsupported
- `:variant` suffixes not changing capability
- preserving caller-supplied `text.*` fields on Responses
- set integrity against `OPENROUTER_CHAT_MODELS`

I also added the `model-meta` package subpath to the build entries so the requested export is actually published.

## Local verification

Passed:

- `CI=true pnpm --filter @tanstack/ai-openrouter test:lib -- openrouter-combined-structured-output.test.ts`
- `CI=true pnpm --filter @tanstack/ai-openrouter test:types`
- `CI=true pnpm --filter @tanstack/ai-openrouter test:eslint`
- `CI=true pnpm --filter @tanstack/ai-openrouter build`
- `CI=true pnpm --filter @tanstack/ai-openrouter test:build`
- `CI=true pnpm --filter @tanstack/ai-e2e exec playwright test tests/agentic-structured-stream.spec.ts --grep "openrouter"`

I also ran `pnpm test:pr`. The OpenRouter affected targets passed, but the full gate currently stops in `root:test:kiira` on unrelated docs snippets in `docs/adapters/grok.md` and `docs/media/video-generation.md`.

## Summary by CodeRabbit

* **New Features**
  * OpenRouter now supports combining tool calls with structured output schemas in a single request for compatible models.
  * Available through both chat-completions and Responses-style adapters.
  * Supported requests can return the final structured response from the same stream, reducing extra processing.
  * Capability detection accounts for fallback models and model variants.

* **Bug Fixes**
  * Older structured-output models continue using the existing flow when combined mode isn’t supported.


tombeckenham:
Rebased onto latest `main` (was 1 commit behind) and force-pushed with `--force-with-lease`. Conflict in `testing/e2e/src/lib/feature-support.ts` resolved by keeping both `openrouter` and `byteplus` in the `agentic-structured-stream` support set. CI approval for the prior head was also submitted.

.changeset/openrouter-combined-tools-and-schema.md
```diff
@@ -0,0 +1,7 @@
+---
+'@tanstack/ai-openrouter': minor
+---
+
+Add native combined tools + `outputSchema` mode to both OpenRouter text adapters (chat-completions and Responses). When the resolved upstream model supports emitting a schema-constrained final answer alongside tool calls in a single pass, `chat({ outputSchema, tools, stream: true })` now wires the JSON Schema into the same streaming request as the tools and harvests the final-turn JSON, skipping the separate finalization round-trip.
```
AlemTuzlak, line 5: **[suggestion]** This is a user-visible behavior change (`chat({ outputSchema, tools, stream: true })` on capable OpenRouter models skips the extra finalization call; middleware no longer sees `structuredOutput`; a new public export `@tanstack/ai-openrouter/model-meta` lands). The PR updates no docs and does not bump `updatedAt` in `docs/config.json`. `docs/adapters/openrouter.md` still says nothing about combined mode or the new export. `docs/advanced/middleware.md` still lists only "modern OpenAI Chat Completions, OpenAI Responses, Claude 4.5+, Gemini 3.x, Grok 4.x family" as native-combined adapters.

**Suggestion:** Add a short combined-mode section to `docs/adapters/openrouter.md` (show both the chat-completions and Responses request, plus `OPENROUTER_COMBINED_TOOLS_AND_SCHEMA_MODELS`). Add OpenRouter to the native-combined list in `docs/advanced/middleware.md` (and `docs/structured-outputs/streaming.md` / `with-tools.md` if those pages talk about which adapters skip finalization). Set `updatedAt` on the touched `docs/config.json` entries.

packages/ai-openrouter/src/model-meta.ts
```diff
@@ -16315,3 +16315,116 @@ export const OPENROUTER_IMAGE_MODELS = [
...
+  'anthropic/claude-opus-4.6',
+  'anthropic/claude-opus-4.7',
+  'anthropic/claude-opus-4.7-fast',
+  'anthropic/claude-opus-4.8',
+  'anthropic/claude-opus-4.8-fast',
+  'anthropic/claude-sonnet-4.5',
+  'anthropic/claude-sonnet-4.6',
+  'anthropic/claude-sonnet-5',
+
+  // Google — Gemini 3.x family
```
AlemTuzlak: **[suggestion]** The comment says the set is "Gemini 3.x only" and "Grok 4.x (tool-capable)". On this PR's catalog that list is complete. Current `main` already has more catalog ids that pass the same `responseFormat` + `toolChoice` `satisfies` check and sit in the upstream gates: `google/gemini-3.5-flash-lite`, `google/gemini-3.6-flash`, `google/gemini-3.7-flash` (all in `GEMINI_COMBINED_TOOLS_AND_SCHEMA_MODELS`), plus `x-ai/grok-4.5` and `x-ai/grok-4.6`. After merge they will silently stay on the legacy two-call path. The new unit test only checks `set ⊆ OPENROUTER_CHAT_MODELS`, so this drift will not fail CI.

**Suggestion:** Rebase onto current `main`, add every Gemini 3.x / Grok 4.x catalog id that has `responseFormat` + `toolChoice` (still skip image-only and `grok-4.20-multi-agent`). Tighten the integrity test so every catalog Gemini 3.x / Grok 4.x text id is in the set, not only the reverse inclusion. Plan to bump the list whenever the catalog generator adds a new 3.x / 4.x id.

testing/e2e/src/lib/feature-support.ts
```diff
@@ -199,11 +199,17 @@ export const matrix: Record<Feature, Set<Provider>> = {
...
+  // openrouter (#612): its default test model `openai/gpt-4o` is a member of
+  // OPENROUTER_COMBINED_TOOLS_AND_SCHEMA_MODELS, so the chat adapter's
+  // `supportsCombinedToolsAndSchema()` returns true and the engine takes the
+  // native combined path — same single-request shape this fixture models.
   'agentic-structured-stream': new Set([
     'openai',
     'anthropic',
     'gemini',
     'grok',
+    'openrouter',
```
AlemTuzlak, line 209: **[suggestion]** Only `'openrouter'` (chat-completions, default `openai/gpt-4o`) is added to `agentic-structured-stream`. The Responses adapter got the same `supportsCombinedToolsAndSchema` + `text.format` wiring, and the e2e provider `'openrouter-responses'` already exists with the same default model. Issue #612 asked both adapters to wire the schema; harvest/parse for Responses is untested. The new unit file never calls `chatStream` either — it only casts to reach protected `mapOptionsToRequest` — so a Responses stream/harvest break would not be caught.

**Suggestion:** Add `'openrouter-responses'` to this set (default model is already combined-capable). Mirror the openai-base #605 test: call `chatStream({ tools, outputSchema })` and assert the SDK payload, then add one `chat({ tools, outputSchema, stream: true })` harvest case so the engine actually parses the final-turn JSON.

packages/ai-openrouter/src/adapters/text.ts
```diff
@@ -1191,6 +1193,24 @@ export class OpenRouterTextAdapter<
       ? convertToolsToProviderFormat(options.tools)
       : undefined
 
+    // Native combined mode (#612): the engine populates `options.outputSchema`
```
AlemTuzlak: **[suggestion]** The new comments narrate #612, the engine harvest path, and "legacy `structuredOutput*` methods strip `outputSchema`". The same block is copied on `supportsCombinedToolsAndSchema` and again in `responses-text.ts`. That is design history, not a local constraint.

**Suggestion:** Cut this to one short why: attach `responseFormat` / `text.format` only when `outputSchema` is set and every routed model is in the combined set. Drop the issue-number walkthrough and the engine-harvest explanation.

packages/ai-openrouter/src/model-meta.ts
```diff
@@ -16315,3 +16315,116 @@ export const OPENROUTER_IMAGE_MODELS = [
...
+  'google/gemini-3.1-pro-preview',
+  'google/gemini-3.1-pro-preview-customtools',
+  'google/gemini-3.5-flash',
+
+  // OpenAI — strict-json_schema era, tool-capable text models
+  'openai/gpt-4o',
+  'openai/gpt-4o-2024-08-06',
+  'openai/gpt-4o-2024-11-20',
+  'openai/gpt-4o-mini',
+  'openai/gpt-4o-mini-2024-07-18',
```
AlemTuzlak: **[suggestion]** The allowlist comment says OpenAI combined mode starts at `gpt-4o-2024-08-06` (strict `json_schema` era) and the tests call out `openai/gpt-4o-2024-05-13` as excluded for that reason. `openai/gpt-4o-mini-2024-07-18` is still in the set. That snapshot predates the 2024-08-06 structured-outputs launch the comment uses as the cut. If OpenRouter routes `json_schema` + `tools` + `strict: true` to that pin, this is a false positive; the engine will skip finalization and harvest may fail.

**Suggestion:** Drop `openai/gpt-4o-mini-2024-07-18` unless there is a recorded OpenRouter response that proves that pin accepts `tools` + `responseFormat: json_schema`. Keep the unpinned `openai/gpt-4o-mini` alias.

packages/ai-openrouter/src/adapters/openrouter-combined-structured-output.test.ts
```diff
@@ -0,0 +1,260 @@
...
+    it('omits responseFormat when any fallback model is unsupported', () => {
+      const req = buildChatRequest('openai/gpt-4o', {
+        models: ['openai/gpt-4o-2024-05-13'],
+      })
+      expect(req.responseFormat).toBeUndefined()
+      expect(req.models).toEqual(['openai/gpt-4o-2024-05-13'])
+      expect(req.tools).toBeDefined()
+    })
+
+    it('keys capability off the bare model id, ignoring the :variant suffix', () => {
```
AlemTuzlak: **[nit]** This test passes `{ variant: 'nitro' }`. The gate reads `this.model` (`openai/gpt-4o`) and `modelOptions.models`, not `modelOptions.variant`. `stripOpenRouterModelVariant` is never hit. A constructor id or fallback like `openai/gpt-4o:nitro` / `openai/gpt-4o-2024-05-13:nitro` is untested.

**Suggestion:** Add a case with `models: ['openai/gpt-4o:nitro']` (gate stays true) and `models: ['openai/gpt-4o-2024-05-13:nitro']` (gate goes false).

AlemTuzlak (approved):

PR #864, opened 2026-06-30
refactor(ai-react): construct useChat ChatClient via useState lazy init

jingeon27:
## Summary

`useChat` constructs its internal `ChatClient` with `useMemo(() => …, [clientId])`. React's docs are explicit that [`useMemo` is a performance hint, not a semantic guarantee](https://react.dev/reference/react/useMemo#caveats) — React may discard a memoized value and recompute it. For an instance you need to keep stable, that's the wrong contract: a spurious recompute constructs a *second* `ChatClient`, and each one owns a `StreamProcessor`, a devtools bridge, and a connection.

This switches client construction to a `useState` lazy initializer — the documented "runs once per mount" guarantee, which is what React recommends for [creating instances](https://react.dev/reference/react/useState#avoiding-recreating-the-initial-state). The client is still recreated synchronously when `id` changes: `useState` has no dependency array, so this uses the documented ["adjust state during render"](https://react.dev/reference/react/useState#storing-information-from-previous-renders) pattern (track the previous `id`, swap during render when it differs). The existing `[client]` effect still disposes the superseded instance, so lifecycle is unchanged.

While here, also removes an unused `isFirstMountRef`.

## Changes

- `packages/ai-react/src/use-chat.ts`:
  - Replace `const client = useMemo(() => {…}, [clientId])` with a `createClient()` factory + `useState(createClient)` lazy initializer, recreating on `id` change via the previous-`clientId`-in-state pattern.
  - Remove the unused `isFirstMountRef`.
  - Update the adjacent `messagesRef` render-sync comment (it referenced `useMemo`).

## Test plan

- `@tanstack/ai-react` unit suite: **152/152 pass, unchanged** — including the `client recreation` tests, notably *"return new client messages during the id change render"*, which asserts the swap stays synchronous.
- `tsc`, `eslint`, `build`, and `publint` all green.

No public API change and no observable behavior change — the instance is still created during render (same timing as `useMemo`); this only upgrades the memoization primitive to the one with a runtime guarantee.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved chat hook stability by making its internal client creation more predictable during React renders.
  * When the chat identifier changes, the client is now refreshed immediately and consistently.
  * Removed an unnecessary internal dependency, reducing the chance of extra recomputation.

### DIFF
--- a/packages/ai-react/src/use-chat.ts
+++ b/packages/ai-react/src/use-chat.ts
@@ -51,23 +51,29 @@ export function useChat<
   const messagesRef = useRef<Array<UIMessage<TTools>>>(
     options.initialMessages || [],
   )
-  const isFirstMountRef = useRef(true)
   const activeClientRef = useRef<ChatClient | null>(null)
   const cleanupInvalidationRef = useRef<ReturnType<typeof setTimeout> | null>(
     null,
   )
 
-  // Update ref synchronously during render so it's always current when useMemo runs.
+  // Update ref synchronously during render so it's always current when the
+  // client is created (createClient runs during render via the useState lazy
+  // initializer / the clientId-reset below).
   messagesRef.current = messages
 
   // Track current options in a ref to avoid recreating client when options change
   const optionsRef = useRef<UseChatOptions<TTools, TSchema, TContext>>(options)
   optionsRef.current = options
 
-  // Create ChatClient instance with callbacks to sync state
-  const client = useMemo(() => {
+  // Create the ChatClient eagerly, but exactly once per `clientId`.
+  // `useState`'s lazy initializer runs once per mount (a semantic guarantee),
+  // whereas `useMemo` is only a performance hint React may discard and
+  // recompute — which would spuriously construct a second ChatClient (it owns
+  // a StreamProcessor, a devtools bridge, and a connection). When `clientId`
+  // changes we swap synchronously via the "adjust state during render"
+  // pattern; the `[client]` effect below disposes the superseded instance.
+  const createClient = (): ChatClient<TTools, TContext> => {
     const messagesToUse = options.initialMessages || []
-    isFirstMountRef.current = false
 
     // Build options with conditional spreads for fields whose source
     // type is `T | undefined` but the ChatClient target uses a strict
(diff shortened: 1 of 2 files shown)


packages/ai-react/src/use-chat.ts
```diff
@@ -159,7 +165,18 @@ export function useChat<
     })
     activeClientRef.current = instance
     return instance
-  }, [clientId])
+  }
+
+  const [client, setClient] =
+    useState<ChatClient<TTools, TContext>>(createClient)
```
AlemTuzlak: **[suggestion]** This is dead weight. `useState(createClient)` plus `trackedClientId` plus during-render `setClient(createClient())` does not fix a production React 18/19 bug and does not fix Strict Mode leaks. It is more code for the same committed behavior, and it is slower on `id` change (discarded render still reads the old `client`, then a retry). The PR body even says "no observable behavior change — this only upgrades the memoization primitive." An upgrade with no production effect is not a reason to land. The `isFirstMountRef` delete is fine and does not need any of this.

**Suggestion:** Revert the client construction to the previous one-liner. Keep the unused-ref delete if you want that cleanup.

```ts
const client = useMemo(() => {
  const messagesToUse = options.initialMessages || []
  // ... existing ChatClient construction unchanged ...
  activeClientRef.current = instance
  return instance
}, [clientId])
```

packages/ai-react/src/use-chat.ts
```diff
@@ -51,23 +51,29 @@ export function useChat<
...
+  // initializer / the clientId-reset below).
   messagesRef.current = messages
 
   // Track current options in a ref to avoid recreating client when options change
   const optionsRef = useRef<UseChatOptions<TTools, TSchema, TContext>>(options)
   optionsRef.current = options
 
-  // Create ChatClient instance with callbacks to sync state
-  const client = useMemo(() => {
+  // Create the ChatClient eagerly, but exactly once per `clientId`.
```
AlemTuzlak: **[suggestion]** The new comments restate React docs and overclaim. "`useState`'s lazy initializer runs once per mount (a semantic guarantee)" is false in Strict Mode, where the initializer runs twice and the extra `ChatClient` is dropped on the floor (never disposed). The `messagesRef` comment and the id-change block comment do the same: they explain WHAT the code does and WHY React might one day discard `useMemo`, not a non-obvious constraint in this file. Comments like this rot the moment the next person copies the pattern into `useGeneration`.

**Suggestion:** If the `useMemo` stays, delete these comments. If this pattern somehow stays, one short line is enough: "Recreate the client when `clientId` changes." Do not quote the `useMemo` caveat in source.

AlemTuzlak:
I just checked this PR out and analyzed the current state and I don't think this is needed right now as the chat client is not reconstructed and unstable, neither on react 18 nor 19 so I wouldn't prematurely optimize it yet. Thank you for the PR it's really appreciated!

PR #867, opened 2026-06-30
fix(ai): preserve tool-call metadata across MESSAGES_SNAPSHOT so addToolResult works (#859)

### LINKED ISSUE #859: MESSAGES_SNAPSHOT + addToolOutput
### TanStack AI version

Lastest

### Framework/Library version

React 19

addToolOutput() doesn't work properly when the chat contains a MESSAGES_SNAPSHOT-type message. If I delete it, I can add output without any problems.

-

_No response_

None

### Terms & Code of Conduct

IdiotTIQS:
## 🎯 Changes

Fixes #859 — `addToolResult` / `addToolOutput` silently no-op'd after a `MESSAGES_SNAPSHOT`.

### Root cause

`handleMessagesSnapshotEvent` replaces `this.messages` with the normalized snapshot. The AG-UI `MESSAGES_SNAPSHOT` wire shape cannot reconstruct client-side tool-call metadata a server may omit:

- a `role: 'tool'` message only carries `toolCallId` + `content`, and
- an assistant snapshot message may drop the `toolCalls` the client already observed via `TOOL_CALL_*` events.

After the snapshot, the matching `tool-call` part could disappear, so a later `addToolResult(toolCallId)` couldn't locate the call:

```
[StreamProcessor] Could not find message with tool call <id>
```

and silently no-op'd. Removing the snapshot (the reporter's "if I delete it") leaves the streamed assistant message intact, which is why it then worked.

### Fix

`handleMessagesSnapshotEvent` now reconciles the normalized snapshot against the pre-snapshot state in a small post-pass (`reconcileSnapshotToolCalls`):

- **(a) consistency** — detached `tool-result`-only assistant messages (produced by `aguiSnapshotMessageToUIMessage` for `role: 'tool'` wire messages) are anchored into the preceding assistant message, matching the streaming fan-out shape `assistant: [text, tool-call, tool-result, ...]`. This also removes a duplicate `tool-result` that previously appeared when both the assistant `toolCalls` and a separate `tool` message were present.
- **(b) bug fix** — when a `tool-result` references a `toolCallId` whose `tool-call` part is absent from the snapshot, the `tool-call` part is carried forward from the pre-snapshot state so `addToolResult(toolCallId)` can still find it.

The trigger for (b) is deliberately narrow: it only fires when the snapshot explicitly references a `toolCallId` (via a `tool-result`) but does not itself supply the corresponding `tool-call` part. This frames the change as recovering metadata the snapshot payload cannot express, not as making the snapshot non-authoritative.

### Tests

Added unit coverage in `stream-processor.test.ts` (the same location as the prior snapshot fix #698):

- anchors a snapshot `tool` message into the preceding assistant message
- preserves an in-flight `tool-call` the snapshot omits so `addToolResult` still works (the #859 repro)
- does not duplicate a `tool-call` the snapshot already supplies

## ✅ Checklist

- [x] I have tested this code locally.

### Local verification

All `@tanstack/ai` targets are green locally:

- `test:lib` — 164 tests passing (incl. 3 new)
- `test:types`, `test:eslint` (0 errors), `build`, `test:build`

I also ran the full repo-wide `pnpm run test:pr`. The only failures were in **packages unrelated to this change** and are platform/environment artifacts on my Windows machine, not regressions:

- `@tanstack/ai-sandbox` › `tests/harness-cwd.test.ts` — a path-separator assertion (`path.join` produces `\` on Windows; the code returns `/`). Fails only on Windows; unrelated to this PR.
- `@tanstack/ai-angular:build` — an intermittent npm peer-dep install crash (`Cannot read properties of null (reading 'matches')`); re-running it on a clean `main` checkout succeeds.

Neither touches `@tanstack/ai`. Linux CI should be unaffected — happy to adjust if CI surfaces anything.

## 🚀 Release Impact

- [x] This change affects published code, and I have generated a changeset (`@tanstack/ai` patch).

---

Note on E2E: there is no existing `MESSAGES_SNAPSHOT` harness in `testing/e2e` (snapshots aren't part of the recorded provider flow), and the directly analogous prior snapshot fix (#698) was covered with unit tests in `stream-processor.test.ts` rather than E2E. I followed that precedent — happy to add an E2E scenario if you'd prefer one.

I raised the two fix directions on the issue first (#859 comment) and implemented the combined approach with the focus on the actual bug fix (b). Let me know if you'd like the scope narrowed/widened.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved snapshot hydration to reconcile detached `tool-result` messages with the originating assistant turn.
  * Preserved client-observed tool-call metadata when snapshots omit `toolCalls`, keeping follow-up tool result handling working.
  * Avoided duplication of tool-call parts when snapshot data already includes them, and added warnings when recovery isn’t possible.
* **Tests**
  * Expanded streaming snapshot hydration coverage for correct anchoring, metadata carry-forward across snapshots, and warning behavior.


NovolineGast:
On the **E2E coverage** requirement in CONTRIBUTING — flagging this explicitly rather than leaving it buried in the PR description, so a maintainer can make the call.

This fix lives entirely in the client-side `StreamProcessor` (reconciling a `MESSAGES_SNAPSHOT` against pre-snapshot tool-call state). I looked at `testing/e2e` and there is **no path that emits `MESSAGES_SNAPSHOT`**: the harness drives real provider adapters against aimock, and aimock only replays ordinary LLM streams (text / tool-call / tool-result). `MESSAGES_SNAPSHOT` is an AG-UI server feature no provider in the matrix produces, so the existing `providersFor('feature')` model can't exercise this path without new infrastructure (a dedicated snapshot-emitting route + UI page + spec).

Given that, I covered the behaviour with unit tests in `stream-processor.test.ts`, which is exactly where the directly analogous prior snapshot fix (#698, "normalize AG-UI snapshot messages") was tested — no E2E was added there either.

Options, your call:
1. **Keep unit-only** (current state), consistent with #698.
2. I add an **`@tanstack/ai-client` integration test** that drives a real `ChatClient` through a mock connection which injects a `MESSAGES_SNAPSHOT` chunk mid-stream, then calls `addToolResult` — proves the end-to-end client behaviour without Playwright.
3. I build out a **full Playwright E2E**: a new snapshot-emitting route + tools page + spec in `testing/e2e`.

I'd lean (1) or (2); (3) feels like a lot of new harness for a client-internal fix. Happy to do whichever you prefer.

NovolineGast:
Re: the review about rebuilding tool-call routing/state after snapshot reconciliation (`toolCallToMessage` / `messageStates` not rehydrated after `resetStreamState()`).

Good catch, and the mechanism is real — after a `MESSAGES_SNAPSHOT`, `handleToolCallArgsEvent` (bails at `toolCallToMessage.get(...)` then `getMessageState(...)`), `TOOL_CALL_END`, `TOOL_CALL_RESULT`, approval, and ui-resource routing all key off maps that `resetStreamState()` clears. So follow-up streaming events for a reconciled tool call can be dropped.

Two notes on scope:

1. **Pre-existing, not introduced here.** On `main` before this PR, `handleMessagesSnapshotEvent` already did `this.resetStreamState()` then `this.messages = chunk.messages.map(...)` — the routing maps were always cleared on every snapshot and never rebuilt. This PR does not touch those maps.

2. **Orthogonal to the bug this PR fixes.** #859 is about `addToolResult(toolCallId)`, which resolves purely against `this.messages` and never consults `toolCallToMessage` / `messageStates`. The fix carries the `tool-call` part forward into `this.messages` so `addToolResult` can find it. Restreaming into a reconciled call is a separate concern.

Rehydrating per-message tool state is a reasonable robustness improvement, but (as the review notes) it is a heavy lift that widens this PR beyond #859. I would prefer to keep this PR minimal and open a dedicated follow-up issue, unless you would rather I fold it in here — happy to do whichever you prefer.

tombeckenham (approved):

PR #896, opened 2026-07-04
feat: add self-describing billed usage to TokenUsage

### LINKED ISSUE #816: Add a required `billingUnit` field to TokenUsage so billed quantities are self-describing
## Problem

`TokenUsage` reports billed *quantities* but never the *unit* they're counted in. Today:

- `unitsBilled?: number` is a bare count — the doc explicitly says *"the unit itself (megapixels, seconds, images, …) is provider-defined and **not carried here**."* (`packages/ai-event-client/src/index.ts`)
- `durationSeconds?: number` carries seconds, but only by convention/field name.

A consumer that receives `unitsBilled: 5` cannot tell whether that's 5 seconds of video, 5 images, or 5 megapixels. The unit is exposed **nowhere machine-readable** — the OTel middleware emits `tanstack.ai.usage.units_billed` unitless, and the only "seconds"/"images" strings in the codebase are hardcoded UI labels and JSDoc.

This already bites the media example (`examples/ts-react-media/src/components/VideoGenerator.tsx`), which has to *guess* the unit from a proxy signal — it branches on whether `cost` is present to decide between rendering "N seconds of video" vs "N fal units". That's fragile and only works because the two current providers happen to differ on `cost` presence.

It also drives inconsistency between adapters for the same concept: video length is reported as `durationSeconds` by the Gemini video adapter but `unitsBilled` by the fal/grok video adapters.

## Proposal

Add a field that names the billed unit, alongside the quantity:

```ts
interface TokenUsage<TProviderDetails = ProviderUsageDetails> {
  // ...existing fields...

  /**
   * The unit the billed quantity is counted in. Lets consumers label/aggregate
   * usage without out-of-band knowledge of the provider/activity.
   */
  billingUnit: BillingUnit
}

type BillingUnit =
  | 'tokens'        // chat, summarize, embedding
  | 'seconds'       // video, audio/music, transcription, some TTS
  | 'characters'    // some TTS (e.g. OpenAI)
  | 'images'        // image generation
  | 'megapixels'    // image generation
  | 'units'         // opaque provider-defined unit (e.g. fal's "fal units")
  | (string & {})   // escape hatch for provider-specific units
```

(Open to a better name — `billingUnit`, `billedUnit`, `usageUnit`. `billingUnit` pairs cleanly with the existing `unitsBilled` quantity. The `(string & {})` member keeps the common values autocompletable while allowing genuinely opaque units.)

## Can it be required? Yes.

Every activity has a billing unit that the adapter knows **statically** from its own pricing model — independent of whether the provider echoes a quantity:

| Activity | Unit |
| --- | --- |
| chat / summarize / embedding | `tokens` |
| generateTranscription | `seconds` |
| generateSpeech (TTS) | `characters` or `seconds` |
| generateAudio | `seconds` |
| generateImage | `images` / `megapixels` / `units` |
| generateVideo | `seconds` |

Because the unit kind is always knowable, making `billingUnit` **required** is feasible and desirable — it forces every adapter to declare its unit rather than emit an ambiguous bare count. The `'units'` / open-string members keep it honest for opaque providers, so "required" never forces a lie.

### Caveats to settle in design
- **Breaking change.** Every adapter that constructs a `TokenUsage` must set `billingUnit`. Coordinated migration across all adapter packages.
- **Multi-dimensional billing.** Chat bills input + output tokens (both `tokens`, fine). Some media calls also charge input tokens on top of the media unit; `billingUnit` should describe the primary non-token billed unit while tokens remain in the token fields. Worth documenting.
- **Alternative/complement:** the unit is fundamentally a property of model pricing (`model-meta.ts`), so it could additionally live there; echoing it on `TokenUsage` is what gives consumers the label at the point of use.

## Acceptance
- `billingUnit` added to `TokenUsage` (required) with the union above.
- All adapters set it for every activity.
- Media example reads `billingUnit` for labels instead of guessing from `cost` presence.
- Gemini/fal/grok video adapters align on the same unit representation.

season179:
## Summary

Closes #816.

Non-token billed quantities (seconds of audio/video, fal's endpoint units) are currently reported as bare counts on `unitsBilled` / `durationSeconds`, so consumers have to guess the unit from provider identity. This PR adds an optional structured `billed` field to `TokenUsage` that pairs the quantity with the unit it is denominated in:

```ts
usage.billed // { quantity: 8, unit: 'seconds' }
```

### Design notes (where this deviates from the issue proposal)

The issue proposed a required flat `billingUnit` field. This PR uses an optional `billed: { quantity, unit }` object instead, for three reasons:

- A flat unit field doesn't say **which** quantity it describes — `durationSeconds` and `unitsBilled` can coexist, so `billingUnit: 'seconds'` would still be ambiguous. The object shape forces quantity and unit to travel together.
- Making it required would be a runtime lie: usage objects cross the SSE wire and server-function JSON boundaries, where version skew between client and server would surface `undefined` in a field typed as required. Optional matches reality — token-only activities simply don't set it (the token fields are already self-describing).
- Required would also break every existing `TokenUsage` literal (~50 test files and any userland constructor) for no information gain.

`BillingUnit` is an open string union (`'tokens' | 'seconds' | 'characters' | 'images' | 'videos' | 'megapixels' | 'requests' | 'units' | (string & {})`) so common units autocomplete while provider-specific ones stay representable. `'units'` marks an opaque provider-defined unit (fal's endpoint units, priced via fal's pricing API).

One premise of the issue didn't hold up: the Gemini video adapter reports no usage at all — its `durationSeconds` is request config sent to Veo, and the operation response carries no billing signal — so there was nothing to align there.

### Changes

- **`@tanstack/ai-event-client`**: `BillingUnit` + `BilledUsage` types; `TokenUsage.billed`; `unitsBilled` / `durationSeconds` deprecated (still populated). Re-exported from `@tanstack/ai`.
- **Producers**: fal adapters report `{ quantity, unit: 'units' }` from the `x-fal-billable-units` header; Grok Imagine video `{ quantity, unit: 'seconds' }`; OpenAI (gpt-4o duration usage + whisper verbose_json) and Grok STT transcription `{ quantity, unit: 'seconds' }`.
- **`otelMiddleware`**: emits the pair as `tanstack.ai.usage.billed_quantity` / `tanstack.ai.usage.billed_unit` span attributes, guarded so a non-finite quantity never emits a dangling unit. Deprecated attributes still emitted for backward compatibility.
- **Example**: `ts-react-media` labels billed usage from `billed.unit` instead of inferring from cost/provider.
- **Docs**: media generation pages, OTel attribute tables (the `units_billed` attribute row was previously missing entirely), Grok adapter page, media-generation skill.

### Tests

- Unit tests across ai-fal / ai-grok / ai-openai / ai for the new field, including a NaN-guard case for the OTel attributes.
- New E2E spec: drives `generateTranscription` (whisper-1, duration-billed) through `otelMiddleware` against the aimock fixture and asserts the billed span attributes end-to-end. The two ~100-line duplicated in-memory tracers in the otel E2E routes were deduplicated into a shared helper along the way.
- Note: `testing/e2e/fixtures/transcription/basic.json` turned out to be a dead fixture — its `userMessage` match clause can never match because aimock builds transcription requests with empty `messages`. The live mock is the programmatic `onTranscription` in `global-setup.ts`, which now carries the `duration` the new spec asserts on. Left the dead fixture untouched; can remove it here or in a follow-up if preferred.

`pnpm test:pr` and the full E2E suite pass locally.

## Summary by CodeRabbit

* **New Features**
  * Added first-class `usage.billed: { quantity, unit }` for media and transcription, with legacy billed fields still populated for compatibility.
  * OpenTelemetry middleware now emits `tanstack.ai.usage.billed_quantity` / `tanstack.ai.usage.billed_unit`.
* **Bug Fixes**
  * Aligned provider adapters to report billed quantities/units consistently (including duration-billed transcription and billed video seconds).
* **Documentation**
  * Updated docs, examples, and media/skill guidance to use `usage.billed` instead of `usage.unitsBilled`.
* **Tests**
  * Expanded unit and end-to-end coverage, including a new `/api/otel-transcription` test asserting billed span attributes.


AlemTuzlak:
@tombeckenham leaving this one to you!

AlemTuzlak (approved):

AlemTuzlak:
I like this idea! cc @tombeckenham you'll probably find this useful as well

tombeckenham:
Me too. It's a better design

AlemTuzlak:
Merged latest `main` into this branch (`4a5cbe770`).

Conflicts were in:
- `docs/config.json` — kept today's Grok `updatedAt`
- `packages/ai-grok/src/adapters/video.ts` — kept main's reference/edit/extend docs, and `usage.billed`
- `packages/ai/skills/ai-core/media-generation/SKILL.md` — same: main's Grok 1.5 text + `usage.billed`

The PR is mergeable again. CI is running.

AlemTuzlak (approved):

AlemTuzlak:
@season179 thank you and sorry for the wait

PR #915, opened 2026-07-08
feat(ai): native Files API support across providers (upload adapters + file content source)

### LINKED ISSUE #909: feat(ai): native Files API support across providers (upload adapters + `file` content source)
## Summary

Add first-class support for provider **Files / storage APIs** so callers can upload media once and reference it by a provider-issued handle, instead of re-sending base64 or relying on the provider to fetch a public URL every request.

Motivated by the discussion on #907 (Gemini image URL handling) and the original ask from @L4Ph about `createGeminiImage()` supporting the Files API.

## Why

Today `ContentPartSource` is a two-arm union — `{ type: 'data' }` (inline base64) or `{ type: 'url' }` (HTTP(S)/`gs://`/data URI). That covers a lot, but it can't express a **provider file handle**, and it doesn't give us a place to hang native upload.

Benefits: large/reused inputs uploaded once (lower latency + bandwidth), no re-buffering of base64 on memory-constrained runtimes (the #907 OOM class), and access to provider-side file lifecycle (TTL, deletion).

## Provider support (verified against the vendored SDKs in this repo)

| Provider | Native upload surface | Handle referenced as | Notes |
|----------|----------------------|----------------------|-------|
| **Gemini** | `ai.files.upload()` | `fileData.fileUri` (a `generativelanguage.googleapis.com/...` URL) | Handle is a URL, so it already round-trips through `{ type: 'url' }` today (image-gen fixed in #907). |
| **OpenAI** | `openai.files.upload()` | `input_image` / `input_file` `file_id` (`file-...`) | ID is **not** a URL — no representation today. |
| **Anthropic** | `client.beta.files.upload()` (beta) | `file_id` message source | ID is **not** a URL — no representation today. |
| **fal** | `fal.storage.upload()` | resulting storage URL | fal-client already auto-uploads `Blob`/`File` inputs (see `ai-fal/src/utils/client.ts`); we convert `data:` URLs to Blobs so it can. Could be surfaced explicitly. |
| **xAI / Grok** | *(no dedicated files API found)* | `image.url` / `images[].url` | Grok's image edits accept URLs directly (passthrough), so no upload step is needed for URL inputs. |

## Proposed shape

**1. Tree-shakeable `files` adapter kind** (following the existing `Text`/`Embed`/`Image` split), for providers that have one:

```ts
import { geminiFiles } from '@tanstack/ai-gemini/adapters'
const files = geminiFiles()
const handle = await files.upload({ data, mimeType }) // → { id, uri?, expiresAt? }
await files.get(handle.id)
await files.delete(handle.id)
```

`openaiFiles()`, `anthropicFiles()`, `falFiles()` (wrapping `fal.storage.upload`) likewise. Grok has nothing to wrap — URL inputs already work.

**2. New `{ type: 'file' }` arm on `ContentPartSource`:**

```ts
{ type: 'image', source: { type: 'file', value: fileIdOrUri, provider: 'openai' } }
```

Per-provider typing should enforce that a file reference only flows to the provider that issued it (a `file-...` ID sent to Gemini is a bug). Adapters map it to the right wire field (`file_id` / `fileData.fileUri`).

**3. Optional follow-up:** an auto-upload helper that, above a size threshold, uploads a `{ type: 'data' }` input via the provider's Files API and swaps in the handle — the large-input path #907 called out. fal already does a form of this internally.

## Related: server-side URL buffering audit (#907 class)

Sweeping the repo for adapters that fetch + buffer image URL **inputs** locally:

- **Gemini image** — was buffering; **fixed in #907** (now `fileData` passthrough).
- **Gemini Veo** (`imagePartToVeoImage`) — still buffers HTTP URLs, but **required**: the predict API's `Image` type only accepts `imageBytes` or `gcsUri`, no URI passthrough.
- **OpenAI image edits** (`imagePartToFile`) — still buffers, but **required**: `/v1/images/edits` is multipart and needs real bytes; there's no "fetch this URL" option. A `files.upload` handle would be the memory-safe alternative here.
- Everyone else passes URLs through without buffering: Gemini chat, fal image, Grok image, Mistral vision, OpenAI vision. (OpenAI/Grok/`arrayBufferToBase64` hits elsewhere are all *output* handling — downloading generated media — not input buffering.)

## Scope / open questions

- Capability gating: which models/endpoints accept file handles varies (Imagen vs native Gemini image, Responses vs Chat Completions). Needs `feature-support` / `model-meta` wiring.
- E2E: aimock coverage for upload endpoints (likely unit-first, mirroring the image-gen mapping note in `feature-support.ts`).
- Not all provider file handles are cross-compatible with all generation types — the type layer should reflect that.

## References

- #907 — Gemini image URL passthrough (the immediate trigger)
- Gap analysis notes Files API as a broader Gemini gap (`.agent/gap-analysis/`)

tombeckenham:
Upload media once with `uploadFile()`, then pass a `{ type: 'file' }` handle in chat, image, or video. The provider keeps the bytes. You do not re-send base64 on every request.

CI on `d8aebb66f` is green (Test, E2E, Preview). Review threads are resolved. The E2E wire route only treats `RUN_ERROR` as failure when the message matches `/file/` (mapping or lookup miss). Aimock can emit `RUN_ERROR` after a successful OpenAI `file_id` mapping because it does not model that field.

Closes #909.

## Changes

- New tree-shakeable files adapters: `openaiFiles()`, `anthropicFiles()`, `geminiFiles()`, `falFiles()`.
- New `{ type: 'file', reference: { openai: 'file-…' } }` source. Each adapter reads only its own key.
- Fail-closed preflight: `chat()` / `generateImage()` / `generateVideo()` / `embed()` reject file sources unless the adapter sets `supportsFileSources`.
- Worked example: `examples/ts-react-media` uploads Gemini reference images and fal start frames once.

## Checklist

## Release Impact

### Commands run (this push)

1. CI Test on `d8aebb66f` — pass (`32394661757`, 10m24s).
2. CI E2E on `d8aebb66f` — pass (`32394661795`, 16m7s).
3. Full `pnpm test:pr` was **not** re-run locally after `d8aebb66f`. CI is the gate.

### Manual test

1. Upload a PDF or image with `uploadFile({ adapter: openaiFiles(), input })`.
2. Send that handle in `chat()` with `openaiText` via `fileSourceFromHandle(handle)`.
3. Confirm the request uses `file_id`, not base64.
4. Repeat with `grokText` (or any adapter without `supportsFileSources`). Confirm it throws before a network call.
5. Call `chat({ outputSchema, messages: [file source] })` with no tools on an unsupported adapter. Confirm it throws in structured-output finalization.

- Unit tests: `packages/ai/tests/files-source.test.ts`, per-provider `files-source.test.ts` / `files-adapter.test.ts`.
- E2E: `testing/e2e/tests/file-source-wire.spec.ts`.
- Example: `examples/ts-react-media`.

## Linked issues

Closes #909.

## Risk / rollback

Unsupported adapters now throw at the activity layer, including `embed()` and schema-only `chat({ outputSchema })`. That is the intended fail-closed contract. Revert the PR to undo.

## Public API change

**Before**

```ts
chat({
  adapter: openaiText('gpt-5.5'),
  messages: [
    {
      role: 'user',
      content: [
        { type: 'image', source: { type: 'data', value: base64, mimeType: 'image/png' } },
      ],
    },
  ],
})
```

**After**

```ts
const handle = await uploadFile({
  adapter: openaiFiles(),
  input: { data: base64, mimeType: 'image/png' },
})

chat({
  adapter: openaiText('gpt-5.5'),
  messages: [
    {
      role: 'user',
      content: [
        { type: 'image', source: fileSourceFromHandle(handle) },
      ],
    },
  ],
})
```

## Summary by CodeRabbit

* **New Features**
  * Added Files API support for OpenAI, Anthropic, Gemini, and fal.
  * Upload files, retrieve or delete them where supported, and reuse provider-specific file handles across chat, image, and video requests.
  * Added support for combining references from multiple providers.
  * Added browser-safe helpers for creating file-based content sources.

* **Bug Fixes**
  * Added clear, early errors for unsupported providers, endpoints, and file-reference combinations.

* **Documentation**
  * Added Files API and multimodal usage guides with provider limitations and examples.


tombeckenham:
Rebased onto main (was 47 behind / conflicting) and completed a post-review pass. Summary of what changed beyond the rebase:

**Sweep completion**
- **ai-byteplus** (landed on main after this branch): added the file-source guards to its text/image/video adapters + tests — previously a handle would have been passed to Ark as a URL.
- **Gemini Interactions video path** (`mediaPartToInteractionsContent`): added the missing `assertOwnFileSource` guard the sibling text-interactions adapter already had.
- **openai-base Responses subclasses**: new `supportsFileIdInput` gate (true only on `OpenAITextAdapter`), so Grok/Bedrock/compatible adapters now throw "unsupported file source" instead of inheriting the `file_id` mapping; the chat-completions error copy only suggests `openaiText` for OpenAI itself.

**Types**
- `FileHandle<TProvider>` / `FilesAdapter<TName>` thread each adapter's `as const` name literal through `uploadFile()`; `getFile()`/`deleteFile()` accept the handle itself — cross-provider lifecycle calls are now compile errors (no brands; wire-serialized handles still fit via the `string` default).
- `fileSourceFromHandle` + `FileHandle` exported from `@tanstack/ai/client` so the documented client flow doesn't pull the server entry.

**Docs / skills**
- Fixed the broken `openaiText` import in multimodal-content.md; documented the `uploadFile`/`getFile`/`deleteFile` dispatchers; corrected the fal `get`/`delete` and explicit-key claims (kiira green).
- Agent skills updated: adapter-configuration §7 (Files adapters) + notes in chat-experience and media-generation.

**Tests**
- New: gemini files-source mapping + foreign-handle rejection, OpenAI Responses document-arm behavior change, `toFileHandle` normalizer tests (openai/gemini/fal), grok + byteplus rejection tests, handle-object lifecycle dispatch.
- New aimock E2E (`file-source-wire`): round-trip per issuer + cross-provider rejection end-to-end (aimock's journal normalisation strips user-message image parts, so wire-shape assertions stay in the unit tests).

**Quality gates**: `test:pr` green except three failures verified pre-existing on clean main on this machine (`ai-grok-build` durability-warning tests under nx, `ai-sandbox-local-process` reaper conformance, and the `durable-takeover` e2e "real disconnect, then attach" test — flaky at ~3/8 on clean main). Full e2e: 446 passed, 1 failed (that same pre-existing flake).

tombeckenham:
Maintainer sweep: rebased onto main (force-with-lease).

packages/ai-mistral/src/adapters/text.ts
```diff
@@ -1015,6 +1016,9 @@ export class MistralTextAdapter<
     }
 
     if (part.type === 'image') {
+      if (isFileSource(part.source)) {
```
AlemTuzlak: **[bug]** Adding `{ type: 'file' }` to `ContentPartSource` (no `value`) makes every `source.value` access without an `isFileSource` narrow a compile error. This file guards images, but the document branch at line 1037 still does `part.source.value`. The same hole exists in files this PR did not touch: `packages/ai-openrouter/src/adapters/video.ts:71` (image adapter was guarded, video was not), `packages/ai-cohere/src/adapters/embedding.ts:272`, and `packages/ai-bedrock/src/adapters/embedding.ts:438`. That matches the CI `Test` failures on this head (`@tanstack/ai-mistral`, `@tanstack/ai-openrouter`, `@tanstack/ai-cohere`, `@tanstack/ai-bedrock` build / `test:types`). `embed()` also has no `assertPromptFileSourceSupport` (the helper comment cites `generateAudio`, which only takes a string prompt), so Cohere / Bedrock Titan image embeddings would not fail closed at the activity layer even after the type errors are fixed.

**Suggestion:** Guard every remaining `source.value` site the same way as the image branch here (`isFileSource` throw, then url/data). Cover OpenRouter video and both embedding adapters. Add an `embed()` preflight (or reuse `assertPromptFileSourceSupport` on resolved image parts). Add a patch changeset for `@tanstack/ai-cohere` when you touch it.

packages/ai/src/activities/chat/index.ts
```diff
@@ -1486,6 +1487,13 @@ class TextEngine<
       )
     }
 
+    // Fail closed on `{ type: 'file' }` sources for adapters that haven't
+    // declared support — an adapter written before the file arm existed would
+    // otherwise fall through to its URL/data branch and silently mis-map the
+    // reference. Checked per model call so tool results added mid-loop are
+    // covered too.
+    assertMessagesFileSourceSupport(this.adapter, this.messages)
```
AlemTuzlak: **[bug]** `assertMessagesFileSourceSupport` runs only in `streamModelResponse` (the `chatStream` path). Schema-only structured output with no tools sets `skipAgentLoop` (around line 1173) and goes straight to `runStructuredFinalization`, which never calls the preflight. Middleware `onStructuredOutputConfig` can also change `this.messages` after that point (`applyMiddlewareConfig` at line 3347). Current issuer/base adapters still throw inside mapping, so this is not a silent leak today — but the whole point of the activity-layer gate is that a future adapter (or a missed mapping site) cannot fall through. That is the BytePlus-class hole this PR claims to close, and it is the path people will use for "upload a PDF, extract JSON".

**Suggestion:** Call `assertMessagesFileSourceSupport(this.adapter, this.messages)` in `runStructuredFinalization` after `applyMiddlewareConfig(postOnConfig)`, and add a unit test for `chat({ adapter: grokText(...), outputSchema, messages: [file source] })` with no tools.

packages/ai/src/activities/generateImage/index.ts
```diff
@@ -306,6 +307,10 @@ async function runGenerateImage<
 
   await runGenerationStart(middleware, mwCtx)
 
+  // Fail closed on `{ type: 'file' }` sources for adapters that haven't
+  // declared support (see assertPromptFileSourceSupport).
+  assertPromptFileSourceSupport(adapter, rest.prompt)
```
AlemTuzlak: **[bug]** `assertPromptFileSourceSupport` runs after `runGenerationStart` (line 308) and outside the `try` that calls `runGenerationError` / `logger.errors`. An unsupported file source therefore opens middleware (persistence, otel) as started and never finishes or fails the run. `generateVideo` does this correctly: it preflights before submit / `RUN_STARTED`.

**Suggestion:** Move the existing `try` so it starts before this preflight (keep the check inside it), or run the preflight before `runGenerationStart`.

docs/config.json
```diff
@@ -702,7 +702,13 @@
         {
           "label": "Multimodal Content",
           "to": "advanced/multimodal-content",
-          "addedAt": "2026-04-15"
+          "addedAt": "2026-04-15",
+          "updatedAt": "2026-08-07"
+        },
+        {
+          "label": "Files API",
+          "to": "advanced/files-api",
```
AlemTuzlak: **[bug]** GitHub still reports `mergeable=false`, `mergeable_state=dirty`, label `merge-conflicts`. Files changed on both this branch and current `main` since `c092231d` include this nav file, `packages/ai/src/activities/chat/index.ts`, `packages/ai/src/types.ts`, `packages/ai-fal/src/adapters/video.ts`, `examples/ts-react-chat/src/routes/index.tsx`, `examples/ts-react-media/src/lib/server-functions.ts`, `packages/ai/skills/ai-core/media-generation/SKILL.md`, and `testing/e2e/src/routeTree.gen.ts`. The PR cannot merge until it is rebased.

**Suggestion:** Rebase onto current `main`, resolve those files, and rerun `pnpm test:pr` plus the E2E suite on the rebased head.

packages/ai/src/utilities/tool-result.ts
```diff
@@ -25,6 +25,14 @@ export function isContentPart(value: unknown): value is ContentPart {
   const source = part.source
   if (typeof source !== 'object' || source === null) return false
   const src = source as Record<string, unknown>
+  // `file` sources carry a non-empty provider→reference record instead of a
+  // `value` string.
+  if (src.type === 'file') {
+    const reference = src.reference
+    if (typeof reference !== 'object' || reference === null) return false
```
AlemTuzlak: **[suggestion]** `typeof reference === 'object' && reference !== null` accepts arrays such as `{ type: 'file', reference: ['file-abc'] }`. `Object.values` then sees strings, so `isContentPart` returns true. `fileReferenceFor` looks up by provider name and would throw later, but a multimodal tool result would already have been treated as a file source. The docstring on line 14 still says every non-text part has a string `value`, which is false for the file arm.

**Suggestion:** Require `!Array.isArray(reference)` and that every key/value is a non-empty string. Update the docstring to describe a provider-keyed record, not `value`.

examples/ts-react-chat/src/routes/index.tsx
```diff
@@ -357,8 +357,9 @@ function Messages({
                     )
                   }
 
-                  // Render image parts
-                  if (part.type === 'image') {
+                  // Render image parts (file references have no local bytes
+                  // or URL to render, so only url/data sources get an <img>)
+                  if (part.type === 'image' && 'value' in part.source) {
```
AlemTuzlak: **[suggestion]** The image render branch now requires `'value' in part.source`, so a file-handle image has nothing to show. `hasRenderablePart` (line 110, unchanged) still returns true for every `part.type === 'image'`. A user message that is only a file-handle image still renders an empty bubble.

**Suggestion:** Count an image as renderable only when the source has a `value` (url/data), matching this guard.

packages/ai-anthropic/src/adapters/files.ts
```diff
@@ -0,0 +1,85 @@
...
+
+/**
+ * Create an Anthropic Files adapter with an explicit API key.
+ */
+export function createAnthropicFiles(
+  apiKey: string,
+  config?: Omit<AnthropicFilesConfig, 'apiKey'>,
+): AnthropicFilesAdapter {
+  return new AnthropicFilesAdapter({ apiKey, ...config })
+}
```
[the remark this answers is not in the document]
AlemTuzlak: Addressed in 50237a52cf29: `create*Files(apiKey, config)` now spreads `{ ...config, apiKey }` so the explicit key wins in openai, anthropic, and gemini files adapters.

packages/ai/src/activities/generateImage/index.ts
```diff
@@ -281,6 +282,10 @@ async function runGenerateImage<
...
+
```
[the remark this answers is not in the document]
AlemTuzlak: Already fixed before this comment's line: `assertPromptFileSourceSupport` now runs in public `generateImage()` before middleware start and before `stream: true` emits RUN_STARTED. See 50237a52cf29 / `packages/ai/src/activities/generateImage/index.ts`.

packages/ai/src/utilities/tool-result.ts
```diff
@@ -11,7 +11,7 @@ const CONTENT_PART_TYPES = new Set([
 /**
  * Structural check for a single `ContentPart`. A text part must carry a string
  * `content`; every other modality must carry a `source` with `type` of
- * `'url' | 'data'` and a string `value`.
+ * `'url' | 'data' | 'file'` and a string `value`.
```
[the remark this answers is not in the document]
AlemTuzlak: Already fixed with `!Array.isArray(reference)` and non-empty string values. This is a structural type guard on tool results, not a schema parse, so it stays in `isContentPart` without Zod.

testing/e2e/src/routes/api.file-source-wire.ts
```diff
@@ -0,0 +1,97 @@
...
+export const Route = createFileRoute('/api/file-source-wire')({
+  server: {
+    handlers: {
+      POST: async ({ request }) => {
+        const url = new URL(request.url)
+        const provider = (url.searchParams.get('provider') ??
+          'openai') as Provider
+        const handleProvider =
+          url.searchParams.get('handleProvider') ?? provider
+        const testId = url.searchParams.get('testId') ?? undefined
```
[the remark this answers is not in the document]
AlemTuzlak: Addressed in 50237a52cf29: the route now type-guards `provider` / `handleProvider` against `ALL_PROVIDERS` (400 on unknown values). Spec-side Zod parse of `{ ok }` is extra on an internal test route that already asserts those fields.

testing/e2e/src/routes/api.file-source-wire.ts
```diff
@@ -0,0 +1,97 @@
...
+          for await (const chunk of chat({
+            ...createChatOptions({ adapter }),
+            messages,
+          })) {
+            if (
+              chunk.type === 'RUN_ERROR' &&
+              /file/.test(chunk.message ?? '')
+            ) {
+              runError = chunk.message
+            }
```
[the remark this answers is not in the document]
AlemTuzlak: Addressed in 50237a52cf29: every `RUN_ERROR` chunk sets `runError`. The `/file/` filter is gone.

packages/openai-base/src/adapters/responses-text.ts
```diff
@@ -1829,6 +1835,16 @@ export abstract class OpenAIBaseResponsesTextAdapter<
...
+        if (isFileSource(part.source)) {
+          if (this.supportsFileSources !== true) {
+            throw unsupportedFileSourceError(this.name)
+          }
+          return {
+            type: 'input_image',
+            file_id: fileReferenceFor(part.source, this.name),
+            detail: imageMetadata?.detail || 'auto',
+          }
+        }
```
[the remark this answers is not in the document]
AlemTuzlak: No code change. `openai-base` Responses is the shared mapping for every Responses-based text adapter (OpenAI, Grok, Bedrock Responses, …). File `file_id` belongs there next to url/data, not in a new adapter kind. The guideline about separate `/adapters` exports is for new adapter kinds (`files` vs `text`), which this PR already adds as `openaiFiles()` / `anthropicFiles()` / `geminiFiles()` / `falFiles()`.

testing/e2e/src/routes/api.file-source-wire.ts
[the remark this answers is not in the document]
AlemTuzlak: Follow-up in d8aebb66f: the previous change treated every RUN_ERROR as a route failure and broke the OpenAI round-trip E2E. Aimock emits RUN_ERROR after a successful file_id mapping because it does not model that field. This route only fails on file-source mapping/lookup errors (message matches /file/) or thrown lookup misses. Provider query params still 400 on unknown values.

testing/e2e/src/routes/api.file-source-wire.ts
```diff
@@ -0,0 +1,117 @@
...
+        // Report both the same way so the spec has one shape to assert.
+        let runError: string | undefined
+        try {
+          for await (const chunk of chat({
+            ...createChatOptions({ adapter }),
+            messages,
+          })) {
+            if (chunk.type === 'RUN_ERROR') {
+              runError = chunk.message
+            }
```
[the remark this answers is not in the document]
AlemTuzlak: No code change. This suggestion was already tried in 50237a52c (treat every RUN_ERROR as a route failure). That broke the OpenAI file-source-wire E2E: aimock emits RUN_ERROR after a successful file_id mapping because it does not model that field. The route in d8aebb66f only fails when the RUN_ERROR message matches /file/ (mapping or lookup miss) or when lookup throws. An empty RUN_ERROR is the aimock gap, not a mapping failure. See the later thread on the same route.

AlemTuzlak:
Reply to the last CodeRabbit finding (`provider=elevenlabs` on `/api/file-source-wire`):

No code change. That route is an internal E2E helper. The spec only posts `openai`, `anthropic`, and `gemini`. `createTextAdapter('elevenlabs')` throws on purpose (“no text/chat adapter”). The same throw exists on sibling E2E chat routes. Special-casing `elevenlabs` in `isProvider()` is a one-off; `handleProvider` must stay on `ALL_PROVIDERS` so a foreign ElevenLabs handle can still drive the rejection path.

`openaiFiles()` stays on the package root (`import { openaiFiles } from '@tanstack/ai-openai'`), same as `openaiText()`. There is no `@tanstack/ai-openai/adapters` subpath in this repo; tree-shakeable factories are the package exports.

CI on `d8aebb66f`: Test, E2E, Preview, autofix, zizmor, Socket, CodeRabbit all pass.

tombeckenham:
Maintainer sweep: rebased onto main (force-with-lease).

docs/advanced/files-api.md
```diff
@@ -0,0 +1,245 @@
...
+On the server, attach the file part before the run starts:
+
+```typescript
+import { chat, fileSourceFromHandle } from '@tanstack/ai'
+import { openaiText } from '@tanstack/ai-openai'
+import type { FileHandle, ModelMessage } from '@tanstack/ai'
+
+export function runTurn(messages: Array<ModelMessage>, handle: FileHandle) {
+  const last = messages[messages.length - 1]
+  if (last?.role === 'user' && Array.isArray(last.content)) {
```
AlemTuzlak, line 205: **[bug]** The documented client/server flow is `sendMessage('Describe this', { body: { handle } })` plus `runTurn()`. `sendMessage` with a string makes the last user message `content` a string. `runTurn` only attaches the file when `Array.isArray(last.content)`, so the `if` is false, `fileSourceFromHandle` never runs, and `chat()` continues with text only. The file is dropped with no error. This is the common path, not an edge case.

**Suggestion:** Always rebuild the last user message. If `content` is a string, turn it into `[{ type: 'text', content }, filePart]`. If it is an array, copy it and append. Do not use `Array.isArray` as a silent skip. Show reading the handle from `chatParamsFromRequest` / `forwardedProps.handle`.

packages/ai/src/utilities/ag-ui-wire.ts
```diff
@@ -447,7 +447,21 @@ function collectUserContent(
...
       p.type === 'document'
     ) {
-      out.push(p)
+      // The AG-UI wire schema's `InputContentSource` has only `data` and `url`
+      // arms — there is no way to carry a `{ type: 'file' }` provider handle
+      // across it. Throw rather than drop the part silently; build the file
+      // source server-side (after the handle crosses in your own payload)
+      // until @ag-ui/core gains a file arm.
+      const source = p.source
+      if (source.type === 'file') {
```
AlemTuzlak, line 456: **[bug]** The new throw is correct for the browser send path (`uiMessagesToWire` in `buildRunAgentInputBody`). `chat()` also uses this helper for interrupt `MESSAGES_SNAPSHOT`. The docs then tell the server to `last.content.push({ type: 'file', ... })` on the same `messages` array that `chat()` snapshots. `convertMessagesToModelMessages` keeps that object. On a `beforeModel` / approval interrupt, `buildMessagesSnapshotChunk` serializes those messages and throws. A supporting adapter never gets a request. Any chat that uses tools, approvals, or a `beforeModel` interrupt cannot follow the documented attach-on-`messages` pattern.

**Suggestion:** Keep the throw on the client send path. For snapshots, skip `{ type: 'file' }` parts (or put the handle in metadata) so an interrupt can finish. Document the safe server pattern: leave `messages` as the client sent them, and put `fileSourceFromHandle(handle)` only on `providerMessages` from middleware `onConfig`.

packages/ai-gemini/src/adapters/files.ts
```diff
@@ -0,0 +1,84 @@
...
+    this.client = createGeminiClient(config)
+  }
+
+  async upload(input: FileUploadInput): Promise<FileHandle<'gemini'>> {
+    const { blob, mimeType } = normalizeFileUploadInput(input)
+    const file = await this.client.files.upload({
+      file: blob,
+      ...(mimeType ? { config: { mimeType } } : {}),
+    })
+    return toFileHandle(file)
```
AlemTuzlak, line 33: **[bug]** `GeminiFilesAdapter.upload()` returns as soon as `client.files.upload()` returns. It does not poll `FileState.PROCESSING`, and it does not fail on `FAILED`. The same package already has `uploadGeminiFile()` in `packages/ai-gemini/src/files/index.ts`, which waits until ACTIVE, times out, and throws on FAILED / missing URI. Docs tell people to use `geminiFiles()` / `uploadFile()` and then `chat()` / `generateImage()` right away. A PDF or video handle is often not ready yet, so Gemini rejects the next call. `toFileHandle` also omits `uri` when the API has not issued one, so `fileSourceFromHandle` falls back to the resource name `files/…`.

**Suggestion:** Reuse the poll loop from `uploadGeminiFile` (or call it) inside `upload()`. Wait for ACTIVE, throw on FAILED / timeout / missing URI, then build the handle.

examples/ts-react-media/src/lib/server-functions.ts
```diff
@@ -145,6 +152,32 @@ function asImageToVideoPrompt(
...
+/**
+ * Upload each inline (base64 `data`) image input to the provider's Files API and
+ * swap in a `{ type: 'file' }` handle. A reference image / start frame is then
+ * uploaded once via the tree-shakeable files adapter (`geminiF()` /
+ * `falF()`, the BYOK-keyed wrappers below) instead of being re-sent inline
+ * as base64 on the generation
+ * request — the memory-safe path for large inputs. URL and already-uploaded
+ * sources pass through untouched.
+ */
+async function uploadInlineImageInputs(
```
AlemTuzlak, line 164: **[suggestion]** `uploadInlineImageInputs` uploads every inline `data` image on every `generateImage` / `generateVideo` request. The comment says the image is uploaded once. The browser still sends base64 each time, the server never returns a `FileHandle`, and the next request uploads again. This example does not show handle reuse. It only avoids sending base64 on the generation call.

**Suggestion:** Return the handle to the client, store it, and skip upload when the prompt already has `{ type: 'file' }` or `{ type: 'url' }`. If the example is only "upload then generate in one server call", say that. Do not say "uploaded once".

testing/e2e/src/routes/api.file-source-wire.ts
```diff
@@ -0,0 +1,123 @@
...
+        // errors fail this route.
+        let runError: string | undefined
+        try {
+          for await (const chunk of chat({
+            ...createChatOptions({ adapter }),
+            messages,
+          })) {
+            if (
+              chunk.type === 'RUN_ERROR' &&
+              /file/.test(chunk.message ?? '')
```
AlemTuzlak, line 100: **[suggestion]** The route treats a `RUN_ERROR` as failure only when `chunk.message` matches `/file/`. Any other mapping error is ignored, and the handler returns `{ ok: true }`. The happy-path spec only checks `ok: true` (and a non-empty journal for OpenAI). It does not check `file_id` / `fileData.fileUri` on the wire. A mapping that sends a request and then gets a non-file `RUN_ERROR` from aimock still passes.

**Suggestion:** Fail the route on every `RUN_ERROR` from mapping (or match the known lookup / unsupported messages explicitly). For the happy path, assert the native field in unit tests (already done) and say in the spec that E2E only checks "request left the adapter", not the JSON shape.

packages/ai/skills/ai-core/adapter-configuration/SKILL.md
```diff
@@ -456,6 +456,66 @@ compatible providers speak.
 > Verify the provider's current `baseURL` and model ids against its live docs —
 > they drift. See `docs/adapters/openai-compatible.md` for the full provider table.
 
+### 7. Files Adapters (upload once, reference by handle)
+
+Four providers expose a native Files/storage API as a tree-shakeable `files`
```
AlemTuzlak, line 461: **[suggestion]** Changeset and `docs/advanced/files-api.md` include `grokFiles()` and Grok public-URL handles. This skill still says "Four providers" and lists only `openaiFiles()`, `anthropicFiles()`, `geminiFiles()`, and `falFiles()`. Lifecycle text omits Grok even though `grokFiles()` has `get` / `delete` / `revokePublicUrl`. The PR description lists the same four factories and tells reviewers to confirm `grokText` throws because it has no `supportsFileSources`. That is stale: `GrokTextAdapter` sets `supportsFileSources = true` and maps the public URL.

**Suggestion:** Add `grokFiles()` everywhere the other factories are listed. Fix the PR description manual test: `grokText` must accept a Grok handle, not throw.

packages/ai/src/client.ts
```diff
@@ -318,10 +318,18 @@ export type { AdapterYieldChunk } from './utilities/adapter-yield-chunk'
 export { getChunkRunId, getChunkThreadId } from './utilities/chunk-ids'
 export type { WireMessage } from './utilities/ag-ui-wire'
 
+// A browser client that received an uploaded handle from its server can build
+// the `{ type: 'file' }` content source itself — `fileSourceFromHandle` is a
+// pure object builder, so exporting it here keeps the documented client flow
+// from pulling in the server entry.
+export { fileSourceFromHandle } from './activities/files/index'
```
AlemTuzlak, line 325: **[suggestion]** Exporting `FileHandle` from `@tanstack/ai/client` is right. Exporting `fileSourceFromHandle` plus the comment that the browser can build the `{ type: 'file' }` source itself fights the docs. If a browser puts that source on a `useChat` message, `uiMessagesToWire` throws. It does not drop the part. The documented flow only needs the `FileHandle` type on the client.

**Suggestion:** Keep the `FileHandle` type export. Either drop `fileSourceFromHandle` from the client entry, or say it is only for a custom (non-AG-UI) transport. Do not call it the documented client flow.

packages/ai/src/activities/files/adapter.ts
```diff
@@ -0,0 +1,120 @@
...
+ * filename) so provider adapters can hand it straight to their SDK. A `Blob`
+ * input passes through; base64 `{ data }` is decoded to bytes. Shared so
+ * provider files adapters don't each re-implement the decode.
+ */
+export function normalizeFileUploadInput(input: FileUploadInput): {
+  blob: Blob
+  mimeType?: string
+  filename?: string
+} {
+  if (input instanceof Blob) {
```
AlemTuzlak, line 96: **[suggestion]** Docs say a `Blob` is the preferred upload input. A browser `File` is a `Blob`, so this branch runs and returns no `filename`. OpenAI / Anthropic / Grok then call `toFile(blob, undefined, …)`. The original `File.name` (for example `report.pdf`) is lost.

**Suggestion:** If `input instanceof File`, pass `input.name` (and `input.type`) through as `filename` / `mimeType`.

AlemTuzlak (commented):
## Summary

This PR adds a real Files API: per-provider `files` adapters, `{ type: 'file', reference }` sources, fail-closed `supportsFileSources` preflight, and typed `FileHandle`s. Adapter mapping looks correct (OpenAI/Anthropic `file_id`, Gemini URI, fal/Grok URL). The main risk is the documented two-endpoint chat flow, not the provider mappers.

## Issue counts by severity

- bugs: 3
- suggestions: 5
- nits: 0

AlemTuzlak (approved):

Issue #914, opened 2026-07-08
sandbox file-diff: make failures observable + correct (follow-up to #892)


tombeckenham:
Follow-up to #892 (typed custom events + sandbox file-diff hooks). A post-merge review surfaced a set of correctness and observability gaps in the new **sandbox file-diff** subsystem. They share a single root cause — `buildFileHookEvent` and the watcher are built on a *"never throws, fall back to `''`/empty-map, don't log"* contract — so a git/exec/fs failure silently becomes an empty value with no signal. Grouping them here since a single fix pass addresses the theme.

The **typed-events** half of #892 is sound and is intentionally excluded from this issue (type-hardening items will be filed separately). Severity overall: **medium** — the feature is new and opt-in (no prod consumers yet), no security/data-loss angle, but it silently delivers wrong/empty data, which is the worst failure mode for an observability feature.

Line references are against `main` post-merge.

## 1. New (untracked) files produce an empty diff — `packages/ai-sandbox/src/file-diff.ts:44-77`

For a `create` event in a **git-backed** workspace (`baseSha !== ''`), `diff()` runs `git diff <baseSha> -- <path>`. Git does not show untracked files, so it exits 0 with empty stdout and `diff()` returns `''`. The `synthesizeAddPatch` fallback that would render the new file's contents only runs when `baseSha === ''` (non-git). Net effect: with `fileEvents: { diff: true }`, **every file the agent creates** — the most common agent action — emits a `sandbox.file.diff` chunk carrying `diff: ''`, and the accessors become internally inconsistent (`after()` has content, `diff()` is empty).

*Fix idea:* when `git diff` yields empty for a `create`, fall back to `synthesizeAddPatch` (or `git add -N` / `git diff --no-index /dev/null <path>`).

## 2. A failed `find` poll fabricates a delete→create event storm — `packages/ai-sandbox/src/watch.ts:204-206`

```ts
return result.exitCode === 0
  ? parseFindOutput(result.stdout, root)
  : new Map<string, string>()
```

A non-zero `find` exit collapses the snapshot to **empty**; the next `diffSnapshots(previous, next)` then emits a `delete` for every tracked file, and a `create` for each when the poll recovers. One transient blip (slow container, `find` not on PATH, permissions) fans phantom deletes/creates out to user hooks and the stream. Note the asymmetry: the *thrown*-exec path in `tick()` (`watch.ts:220`) correctly preserves `previous` — the non-zero-exit path must do the same, and log `stderr`.

[...]

PR #892, opened 2026-07-03
feat(ai): typed custom events + sandbox file-diff hooks

AlemTuzlak:
## What

Makes every framework-emitted CUSTOM event **type-constrained on the chat stream**, and gives sandbox file hooks lazy git-backed diffs.

- **Typed custom events.** Every framework CUSTOM event (`sandbox.file`, `file.changed`, `${adapter}.session-id`, `code_mode:*`, `skill:registered`, plus the already-typed structured-output/tool-interaction/ui events) is now a discriminated interface unified as `KnownCustomEvent`. The chat stream is retyped `ChatStream = AsyncIterable<Exclude<StreamChunk, CustomEvent> | KnownCustomEvent>`, so a plain `if (chunk.type === 'CUSTOM' && chunk.name === '<literal>')` narrows `chunk.value` with **no helper and no cast**. No guard utilities by design.
- **Sandbox file-hook accessors.** `onFile*` hooks receive a `SandboxFileHookEvent` with lazy, git-backed `before()` / `after()` / `diff()` (computed on call — path-only hooks pay nothing). The serialized `sandbox.file` chunk stays path-only.
- **Opt-in live diffs.** `defineSandbox({ fileEvents: { diff: true } })` emits a per-file `sandbox.file.diff` `{ path, diff }` chunk.

## Tests / quality

- Unit tests for the accessors, wire-projection (no accessor leak), and event typing; a **real-handle integration test** (`localProcessSandbox` + real git) that caught a genuine bug — `diff()` was passing the virtual `/workspace` path to `git diff` instead of relativizing, silently yielding `''` on every local-process sandbox.
- Command-injection hardening: all `git` exec args are shell-quoted.
- `kiira` green (774/774 snippets); `@tanstack/ai` and `@tanstack/ai-sandbox` typechecks green. Docs + `ag-ui-protocol`/`middleware` skills updated.
- The full 50-project `test:pr` sweep was not completed locally (monorepo-wide fan-out); CI runs it.

## Notes for reviewers

- **E2E:** the `testing/e2e/` harness has no sandbox wiring (aimock LLM-only), so browser-level sandbox E2E is structurally infeasible here — covered by the real-handle integration test instead. True sandbox E2E infra is a suggested follow-up.
[...]

PR #917, opened 2026-07-08
fix(sandbox): make file-diff failures observable and correct (#914)

AlemTuzlak:
Follow-up to #892, closing #914. A post-merge review of the sandbox **file-diff** subsystem found a cluster of correctness + observability gaps sharing one root cause: `buildFileHookEvent` and the watcher were built on a *"never throws, fall back to `''`/empty, don't log"* contract, so a git/exec/fs failure silently became empty (or wrong) data — the worst failure mode for an observability feature.

Severity: **medium** — the feature is new and opt-in (no prod consumers yet), no data-loss/security angle beyond the git-ignored-content fix below.

## Correctness

- **Untracked files no longer stream empty diffs (#914.1).** `git diff <sha> -- <path>` shows nothing for a file git isn't tracking, so a file the agent creates — **and every later edit to it** — used to emit an empty diff. `diff()` now synthesizes an add-patch keyed on *tracked-ness at the baseline* (probed via the `git show` exit code), not on `event.type`. A tracked file identical to the baseline still diffs empty, and a transient probe failure no longer fabricates a bogus add-patch. Synthesized patches follow git's add-file shape (`diff --git` header, `new file mode`, `\ No newline at end of file` marker, header-only for empty files).
- **No more phantom `create`/`delete` storms (#914.2).** A failed `find` poll now preserves the previous snapshot instead of collapsing to empty; a failed/partial **initial** poll seeds without diffing (re-baselining on the first complete poll); a partial (permission-denied) poll is **merged**, not diffed, so a transiently-unreadable subtree isn't reported as deletions.
- **Watcher/enrichment root alignment.** The watcher now watches `definition.workspace.root` (it was defaulting to `/workspace` while enrichment relativized against `workspace.root`).
- **Teardown always destroys.** A rejecting `watcher.stop()` can no longer skip `definition.destroy()` on the abort path; the native watcher re-seeds if its initial listing fails.

## Observability (#914.3)

Every swallowed git/exec/fs failure — across the diff accessors, both watcher paths (exec-poll + native `fs.watch`), the git-baseline capture, and per-hook dispatch — now **logs before falling back**: real anomalies under `errors`/`warn` (on by default), expected-empty conditions under the `sandbox` debug category. This also aligns the middleware SKILL's hook-error claim with actual behavior (it logs under `errors`, matching the engine's run-scoped path).

## Security-adjacent

- **git-ignored files are withheld from the diff feed.** An agent-created `.env`/credentials file still fires its file event (you're notified it changed) but `diff()` returns `''`, so its contents are never surfaced in the stream.

## Tests (#914.4)

- The previously-untested `pendingDiffs` teardown drain now has deferred-promise tests for `onFinish`/`onAbort`/`onError` (verified to fail if the drain is removed).
- New coverage for: untracked create+edit synthesis, tracked-identical no-op, git-ignored withholding, no-trailing-newline/multi-line/empty synthesized patches, failed/partial/initial-throw poll handling, native re-seed, workspace.root wiring, destroy-on-rejecting-stop, and every new log path.

## Review

Ran through a 3-round automated review-fix loop (correctness / silent-failure / test-coverage / comment / type-design lenses) to convergence — every finding fixed and re-confirmed, all fixes covered by tests verified to fail on revert.

## Verification

`test:types`, `test:eslint` (src), `test:build`, and `test:docs` clean; 189/190 ai-sandbox unit tests pass (the 1 failure is a pre-existing Windows path-separator issue in `harness-cwd.test.ts`, unrelated to this change). Changeset included (`@tanstack/ai-sandbox` patch).

## Summary by CodeRabbit

* **New Features**
  * File diffs are now more reliable and easier to inspect, including clearer patch output for newly created content.
  * Sandbox file reading/diffing and watcher behavior are more observable via structured warnings/debug logs.

* **Bug Fixes**
  * Fixed empty or incorrect diffs for newly created, edited, ignored, and untracked files (including non-git workspaces and “git-ignored” cases).
  * Improved file watching to reduce misclassifying edits as creates/deletes and to stabilize polling/recovery.

* **Documentation**
  * Updated sandbox observability docs and SKILL guidance to clarify edge-case behavior.

### DIFF
--- a/.changeset/sandbox-file-diff-observability.md
+++ b/.changeset/sandbox-file-diff-observability.md
@@ -1,0 +1,12 @@
+---
+'@tanstack/ai-sandbox': patch
+---
+
+Make sandbox file-diff correct and observable (follow-up to #892):
+
+- `diff()` now synthesizes an add-patch for any file git isn't tracking (a file the agent created **and every later edit to it**), keyed on tracked-ness at the baseline rather than on the event being a `create` — so agent-created files no longer stream empty diffs. A tracked file identical to the baseline still diffs empty, and a transient git-show probe failure no longer fabricates a bogus add-patch.
+- The synthesized patch now matches `git diff`'s add-file shape (`diff --git` header, `new file mode`, repo-relative paths).
+- **git-ignored files are withheld from the diff feed**: the file event still fires (you're notified it changed) but `diff()` returns `''`, so a secret like a `.env` never has its contents surfaced.
+- The native `fs.watch` watcher re-seeds lazily if its initial workspace listing fails, so a pre-existing file is correctly reported as a `change` (not a `create`) on first edit.
+- The exec-poll watcher no longer fabricates phantom `create`/`delete` storms: a failed poll (thrown exec, or non-zero exit with no output) preserves the previous snapshot, a failed initial poll seeds without diffing, and a partial (`find` permission-denied) poll is merged rather than diffed so transiently-unreadable files aren't reported as deleted.
+- Every swallowed git/exec/fs failure — in the diff accessors, both watcher paths (exec-poll and native `fs.watch`), the git-baseline capture, and per-hook dispatch — is now logged (real anomalies under `errors`, expected-empty conditions under the `sandbox` debug category) instead of silently becoming empty data.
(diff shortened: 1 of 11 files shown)


tombeckenham (approved):
All good. Small changes only

Issue #820, opened 2026-06-24
ci: type-check is excluded from examples/** and testing/** — call-site type regressions slip through test:pr


tombeckenham:
## Context

`test:pr` (the canonical pre-PR gate, mirrored by the `PR` CI workflow) runs:

```
nx affected --targets=test:sherif,test:knip,test:docs,test:eslint,test:lib,test:types,test:build,build --exclude=examples/**,testing/**
```

The `--exclude=examples/**,testing/**` carve-out means **`test:types` never typechecks the example apps or the `testing/` packages (`testing/e2e`, `testing/panel`)**. This leaves a real hole: a type regression that only manifests at a *call site* is invisible to CI when the only call sites live in the excluded folders.

## Motivating repro (real bug, found by hand during a rebase, not by CI)

`grokSummarize(...)` is not assignable to `summarize()`'s `adapter` param for **any** current Grok model (`grok-4.3`, `grok-build-0.1`):

```
Type 'ChatStreamSummarizeAdapter<"grok-4.3", GrokTextProviderOptions>' is not assignable to
  'SummarizeAdapter<string, object>'.
  Types of property 'summarize' are incompatible.
    ... Type 'SummarizationOptions<object>' is not assignable to 'SummarizationOptions<GrokTextProviderOptions>'.
      Type 'object' is not assignable to type 'GrokTextProviderOptions'.
```

Root cause: `SummarizeAdapter.summarize` is declared as a **property** (arrow type), so under `strictFunctionTypes` its parameter is checked **contravariantly**. Assignability to the constraint `SummarizeAdapter<string, object>` therefore reduces to "is `object` assignable to `TProviderOptions`?". `GrokTextProviderOptions extends Record<string, unknown>` (it carries an index signature), and `object` is **not** assignable to `Record<string, unknown>` — so it fails. OpenAI's options are all-optional with no index signature, so `object` *is* assignable and it passes. (Separate fix tracked for the Grok options shape; this issue is about the CI gap.)

## Why `test:pr` doesn't catch it

Two factors compound:

1. **The error only exists at a `summarize({ adapter })` call site.** The constraint `TAdapter extends SummarizeAdapter<string, object>` is only instantiated when `summarize()` is actually *called* with a concrete adapter. Merely *constructing* the adapter typechecks fine — which is all `packages/ai-grok/tests/grok-adapter.test.ts` does (asserts `.kind`, never calls `summarize()`). So the `ai-grok` package's own `test:types` stays green.
[...]

PR #919, opened 2026-07-09
fix(ai-gemini): resolve GeminiClientConfig from concrete module in adapter types

tombeckenham:
## Problem

The built `.d.ts` files for the Gemini adapters import `GeminiClientConfig` from `'../utils.js'`, but the `utils` barrel builds to `utils/index.js` — no `utils.js` file exists. Under a consumer's `skipLibCheck` this import silently resolves `GeminiClientConfig` to `any`, which:

- disables client-config type-checking for **every** Gemini adapter (the inherited `GoogleGenAIOptions` fields — `httpOptions`, `apiVersion`, `vertexai`, … — vanish from the public types), and
- for the one config with an own member (`GeminiVideoConfig.allowUrlFetch`), the collapsed `Omit` becomes a non-empty weak type, surfacing a **spurious** `` `httpOptions` does not exist `` excess-property error at call sites.

This is what caused `@tanstack/ai-e2e:test:types` to fail after #820 enabled type-checking of `testing/**`.

## Fix

Import the client utilities from the concrete `'../utils/client'` module (matching the `ai-openai` convention) so the emitted declarations point at a file that actually exists and resolve to the real type. Gemini config options are fully typed for consumers again.

## Verification

- `nx run-many --targets=test:types,test:eslint,test:lib,test:build,build --projects=@tanstack/ai-gemini` — all green
- Confirmed the emitted `.d.ts` now imports `'../utils/client.js'` and that `httpOptions` (and the other `GoogleGenAIOptions` fields) are genuine, type-checked keys on the Gemini configs.

Once merged, the `test/gemini-tts-e2e` branch (#—, harshlocham's Gemini TTS matrix PR) will be rebased to pick this up.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected type resolution for Gemini adapter configurations, improving TypeScript accuracy for consumers.
  * Prevented missing-type and `any` fallback issues that could affect editor feedback and type-checking.
  * Fixed a declaration issue that could cause incorrect errors when using video generation features.

### DIFF
--- a/.changeset/gemini-client-config-dts-import.md
+++ b/.changeset/gemini-client-config-dts-import.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/ai-gemini': patch
+---
+
+fix(ai-gemini): fix `GeminiClientConfig` type import in the published adapter declarations. The emitted `.d.ts` files imported `GeminiClientConfig` from a non-existent `'../utils.js'` (the barrel builds to `utils/index.js`), so under `skipLibCheck` it silently resolved to `any` in consumers — masking client-config type-checking for every adapter and producing a spurious "`httpOptions` does not exist" error on `createGeminiVideo`. Adapters now import the type from the concrete `'../utils/client'` module so the declarations resolve to the real type.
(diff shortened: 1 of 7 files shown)


tombeckenham:
This is blocking tests. Merging

PR #921, opened 2026-07-09
feat(ai-sandbox-vercel): tag sandbox API requests with a framework user-agent

marc-vercel:
## 🎯 Changes

On the Vercel side, we would like to be able to distinguish the requests that come from `@tanstack/ai`.

This PR adds the package name `@tanstack/ai` as a token in the User-Agent HTTP header for all Vercel Sandbox API requests (`create`, `resume`, `destroy`). Existing `user-agent` values are preserved and the token is appended rather than replacing them. We have already done something similar with the Eve framework: https://github.com/vercel/eve/pull/332.

## ✅ Checklist

Testing notes: unit tests were added in `packages/ai-sandbox-vercel/tests/user-agent.test.ts` covering token appending/defaulting, header/field preservation, `Request` input handling, and fallback fetch.

## 🚀 Release Impact

## Summary by CodeRabbit

* **New Features**
  * Vercel sandbox requests now include an `@tanstack/ai` identifier in the User-Agent header.
  * The identifier is applied consistently when creating, resuming, and destroying sandboxes.

* **Bug Fixes**
  * Existing request headers and fields are preserved when the User-Agent is enhanced.


packages/ai-sandbox-vercel/src/user-agent.ts
```diff
@@ -0,0 +1,33 @@
...
+ */
+
+export const USER_AGENT_TOKEN = '@tanstack/ai'
+
+/**
+ * Wraps a `fetch` so every request's `user-agent` contains the package name
+ * (e.g. `vercel/sandbox/2.2.1 (…) @tanstack/ai`). When no
+ * `user-agent` is present, the token becomes the whole header.
+ */
+export function withSandboxUserAgent(
```
AlemTuzlak: **[suggestion]** `withSandboxUserAgent` is more code than the job. The SDK always calls `fetch(url.toString(), opts)` (`base-client.js` in `@vercel/sandbox`). The `'headers' in input` / `Request` branch never runs on the real path. `USER_AGENT_TOKEN` is only exported so tests can import it. Eve needs a separate module (version lookup + wrap of user `fetch`). This package needs neither. An inline wrapper of about 15 lines in `provider.ts` is enough.

**Suggestion:** Delete `user-agent.ts` (or keep a tiny helper with no `Request` branch and no file-level essay). Do not add a public export for the token.

packages/ai-sandbox-vercel/src/user-agent.ts
```diff
@@ -0,0 +1,33 @@
+/**
```
AlemTuzlak: **[suggestion]** The file comment and the function comment restate what the code does and how the SDK names its `user-agent`. They also still say "the package name" after the token was changed to the framework name `@tanstack/ai`. Eve’s follow-up on the same design was "remove useless comments". Comments here should be short and say why, or be omitted — the wrap + `headers.set` is obvious.

**Suggestion:** Drop both JSDoc blocks, or leave one short line: append `@tanstack/ai` after the SDK user-agent so Vercel can attribute traffic.

packages/ai-sandbox-vercel/tests/user-agent.test.ts
```diff
@@ -0,0 +1,73 @@
...
+      signal,
+    })
+
+    const [, init] = inner.mock.calls[0]!
+    expect(init?.method).toBe('POST')
+    expect(init?.body).toBe('{"ok":true}')
+    expect(init?.signal).toBe(signal)
+  })
+
+  it('reads the user-agent off a Request input when no init is given', async () => {
```
AlemTuzlak: **[suggestion]** Tests exercise the wrapper in isolation, including a `Request` input the SDK never sends and a `typeof withSandboxUserAgent() === 'function'` check that cannot fail in a useful way. Nothing asserts that `VercelProvider.create` / `resume` / `destroy` actually pass the wrapped `fetch` into `Sandbox.create` / `Sandbox.get`. Forget one of those three sites and every current test still passes.

**Suggestion:** Keep one test that the token is appended to an existing SDK user-agent, and one that other `init` fields are forwarded. Drop the `Request` and default-fetch tests. If you want a test that earns its lines, mock `Sandbox.create` / `Sandbox.get` and assert they received the wrapped `fetch`.

PR #927, opened 2026-07-11
feat: previousJobId and previousImage for follow-up media edits

jherr:
## Summary
- Add first-class `previousJobId` on `generateVideo` so callers always pass the prior job id; `'job'`-kind adapters (Sora remix, Gemini Omni) reference it server-side, `'media'`-kind adapters (Grok, fal) resolve the clip via `getVideoUrl` (fal generate→edit sibling routing included). Omni’s `previous_interaction_id` is removed from video `modelOptions`.
- Add `previousImage` on `generateImage` to prepend a prior `GeneratedImage` (or array / result) into the prompt for models that accept image inputs.
- Docs, skill, examples, unit tests, and e2e coverage for image/video follow-up edits.

## Test plan
- [x] `pnpm test:pr`
- [x] `pnpm --filter @tanstack/ai-e2e test:e2e` (353 passed; video-edit / image-edit specs green)
- [ ] Spot-check docs: `docs/media/video-generation.md` (`previousJobId`) and `docs/media/image-generation.md` (`previousImage`)

Made with [Cursor](https://cursor.com)

## Summary by CodeRabbit

* **New Features**
  * Added follow-up editing for generated images using a previous image.
  * Added follow-up editing and remix workflows for supported generated videos using a previous job.
  * Updated image and video examples with edit prompts and “Edit” actions.
* **Documentation**
  * Expanded media-generation guides with usage details, supported models, and examples.
* **Tests**
  * Added end-to-end coverage for image and video editing flows.
* **Chores**
  * Updated local ignore rules and AgentsRoom configuration to protect personal data.


tombeckenham (commented):
Hi @jherr, this is a good idea given Google's direction - which I like a lot - but... they are litterally the only provider to support this natively. OpenAI's sora model is dead... For images using type: image, as a source is equivalent. I don't think we should do this

tombeckenham (commented):
If there's a single provider that has a feature, then the way to handle it is to use model options. I'd suggest we continue to do that. Suggest closing this PR without merging

tombeckenham:
Closing as per comments

PR #931, opened 2026-07-14
fix(ai): preserve signed thinking order around provider tools

### LINKED ISSUE #910: Anthropic signed thinking blocks are reordered when a UIMessage interleaves provider-executed tools and thinking
### TanStack AI version

0.39.1

### Framework/Library version

@tanstack/ai-anthropic: 0.16.0

When an assistant `UIMessage` contains signed Anthropic `thinking` parts interleaved with provider-executed tools, `convertMessagesToModelMessages` does not preserve the original part order.

This matters for Anthropic because signed `thinking` / `redacted_thinking` blocks must be replayed exactly as they appeared in the original response. If they are moved relative to tool-use blocks, the next Anthropic request can fail with:

```text
400 {"type":"error","error":{"type":"invalid_request_error","message":"messages.N.content.M: `thinking` or `redacted_thinking` blocks in the latest assistant message cannot be modified. These blocks must remain as they were in the original response."}}
```

The shape that triggers it is:

```text
assistant UIMessage parts:
  thinking(signature A)
  provider-executed web_search tool-call
  thinking(signature B)
  local tool-call
  local tool-result
```

The original order is valid because the second `thinking` block happened after the provider tool result. But after conversion to model messages, both thinking blocks are attached to the same assistant segment and replayed before both tool calls.

## Reproduction Script

```js
import { convertMessagesToModelMessages } from "@tanstack/ai"

const messages = [
  {
    id: "assistant-message",
    role: "assistant",
    parts: [
      {
        type: "thinking",
        content: "First signed thinking block",
        signature: "signature-1",
      },
      {
        type: "tool-call",
        id: "srvtoolu_search",
        name: "web_search",
        arguments: '{"query":"top AI companies"}',
        state: "input-complete",
        metadata: {
          providerExecuted: true,
          anthropic: {
            serverToolType: "web_search",
            resultBlockType: "web_search_tool_result",
            result: [
              {
                type: "web_search_result",
                title: "Example result",
                url: "https://example.com",
                encrypted_content: "opaque-provider-payload",
              },
            ],
          },
        },
      },
      {
        type: "thinking",
        content: "Second signed thinking block",
        signature: "signature-2",
      },
      {
        type: "tool-call",
        id: "toolu_create_block",
        name: "createBlock",
        arguments: '{"block":{"type":"entity-grid"}}',
        state: "complete",
        output: { ok: true },
      },
      {
        type: "tool-result",
        toolCallId: "toolu_create_block",
        content: '{"ok":true}',
        state: "complete",
      },
    ],
  },
]

const modelMessages = convertMessagesToModelMessages(messages)

console.log(
  JSON.stringify(
    modelMessages.map((message) => ({
      role: message.role,
      thinking: message.thinking?.map((part) => part.content),
      toolCalls: message.toolCalls?.map((toolCall) => toolCall.function.name),
      toolCallId: message.toolCallId,
      content: message.content,
    })),
    null,
    2,
  ),
)
```

## Actual Output

Both signed thinking blocks are attached to the same assistant model message:

```json
[
  {
    "role": "assistant",
    "thinking": [
      "First signed thinking block",
      "Second signed thinking block"
    ],
    "toolCalls": [
      "web_search",
      "createBlock"
    ],
    "content": null
  },
  {
    "role": "tool",
    "toolCallId": "toolu_create_block",
    "content": "{\"ok\":true}"
  }
]
```

With the Anthropic adapter, that means the signed thinking blocks are appended before the provider `server_tool_use` and the local `tool_use`, changing the original signed transcript order.

## Expected Output

The conversion should split assistant segments when a later signed thinking block appears after an already accumulated tool call, so the original interleaving is preserved:

```json
[
  {
    "role": "assistant",
    "thinking": [
      "First signed thinking block"
    ],
    "toolCalls": [
      "web_search"
    ],
    "content": null
  },
  {
    "role": "assistant",
    "thinking": [
      "Second signed thinking block"
    ],
    "toolCalls": [
      "createBlock"
    ],
    "content": null
  },
  {
    "role": "tool",
    "toolCallId": "toolu_create_block",
    "content": "{\"ok\":true}"
  }
]
```

Equivalently, the Anthropic adapter needs to replay signed thinking blocks in the same relative order as the original provider response content blocks.

## Why This Is Blocking

This happens in normal multi-turn conversations that use Anthropic web search / web fetch plus thinking. The first turn succeeds and can be persisted as a `UIMessage`, but the next user turn fails because the stored assistant message is converted back into a modified Anthropic transcript.

The issue is especially easy to hit with provider-executed server tools because the provider result is stored in the tool-call metadata rather than as a normal `tool-result` part, so there is no tool-result boundary that forces a segment flush before the next thinking block.

See bug description

_No response_

No, because I do not know how

### Terms & Code of Conduct

ekkoitac:
Anthropic rejects the next turn when signed thinking is moved ahead of a provider-executed tool. This PR splits the assistant segment at that boundary, keeps trailing thinking instead of dropping it, and asserts the Anthropic request block order.

## Changes

`convertMessagesToModelMessages` used to put every thinking block on the same assistant segment as later tools. Anthropic then saw `thinking, thinking, server_tool_use, tool_use` instead of `thinking, server_tool_use, thinking, tool_use`.

This PR:

- starts a new assistant segment when thinking follows a provider-executed tool
- still keeps ordinary local tool calls in one turn
- emits a thinking-only segment on the final flush, so a trailing signed block is not dropped
- covers the issue 910 sequence in unit tests and in the Anthropic adapter request payload
- corrects docs that said thinking is UI-only

Fixes #910

## Checklist

## Validation

Ran:

- `pnpm --filter @tanstack/ai exec vitest run tests/message-converters.test.ts` — 69 passed
- `pnpm nx run @tanstack/ai-anthropic:test:lib -- anthropic-adapter.test.ts` — 32 passed
- `pnpm --filter @tanstack/ai test:types` — passed
- `pnpm --filter @tanstack/ai test:oxlint` — 0 errors

Did not run `pnpm test:pr` or the full E2E suite in this pass.

Red/green for issue 910: remove the provider-tool flush, and both thinking blocks collapse onto one assistant message (2 tests fail). Restore the flush, and those tests pass.

### Commands run

See Validation above. Skipped `pnpm test:pr` (full CI matrix) and full E2E in this pass. The E2E spec `testing/e2e/tests/anthropic-server-tool.spec.ts` is on the branch.

### Manual test

1. Convert the issue 910 UIMessage (thinking A, provider web_search, thinking B, local tool, tool result) with `convertMessagesToModelMessages`.
2. Without this PR you get one assistant message with both thinking blocks, then both tools.
3. With this PR you get two assistant messages: thinking A + web_search, then thinking B + local tool, then the tool result.
4. Replay that through the Anthropic adapter. The request content blocks are `thinking, server_tool_use, web_search_tool_result, thinking, tool_use`.

- `packages/ai/tests/message-converters.test.ts` covers the split, local-tool same-turn, and trailing thinking.
- `packages/ai-anthropic/tests/anthropic-adapter.test.ts` asserts the Anthropic wire order for issue 910.
- `testing/e2e/tests/anthropic-server-tool.spec.ts` posts the same sequence through the mock Anthropic server.

## Linked issues

Fixes #910

## Risk / rollback

Low. The extra split only runs when thinking follows a provider-executed tool. Anthropic (and Gemini) merge consecutive assistant messages, so the wire form stays one assistant turn. Revert this PR to undo.

## Public API change

No new exports. Call sites stay the same. Signed thinking now round-trips in original order.

**Before**

```ts
convertMessagesToModelMessages([assistantUiMessage])
// one assistant message: thinking [A, B], tools [web_search, createBlock]
```

**After**

```ts
convertMessagesToModelMessages([assistantUiMessage])
// assistant: thinking [A], tools [web_search]
// assistant: thinking [B], tools [createBlock]
// tool: createBlock result
```

## Release Impact

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserved signed thinking content in the correct order around provider-executed tool calls.
  * Prevented thinking blocks from being replayed before related tool activity.
  * Maintained tool-call metadata and separate tool results during message conversion.
  * Improved round-trip handling of local tool calls, results, and assistant messages.

* **Documentation**
  * Clarified that signed thinking is replayed in its original order, while unsigned thinking remains UI-only.

* **Tests**
  * Added coverage for thinking-order scenarios, metadata preservation, and Anthropic streaming flows.


tombeckenham:

tombeckenham:

tombeckenham:

AlemTuzlak:
Addressed the CodeRabbit findings that still apply:

- **Keep pending thinking on flush.** `flushSegment()` now emits when `pendingThinking` is the only leftover. A trailing signed thinking block after a provider-executed tool is no longer dropped. Covered in `should keep a trailing signed thinking block after a provider-executed tool`.
- **Round-trip thinking assertion.** The local-tool test now also asserts the thinking part survives `modelMessagesToUIMessages`.
- **Stream helper extraction.** Skipped. Two copies of the chunk collector in the e2e route is cheaper than a new helper until a third bug-repro lands.

Also fixed the CI `toEqual` misses (`id` on converted messages) and added an Anthropic adapter test that asserts the wire block order for #910: `thinking, server_tool_use, web_search_tool_result, thinking, tool_use`.

Red/green: without the provider-tool flush, both signed thinking blocks collapse onto one assistant message and those tests fail. With the flush, they pass.

PR #932, opened 2026-07-14
fix: distinguish custom and native provider tools

jan-kubica:
## What changed

- classify Anthropic, Gemini, and OpenAI native tools with stable, namespaced `metadata.__kind` values
- preserve the documented type-only core `ProviderTool` phantom brand
- keep ordinary functions custom even when their public names match provider-native tools
- strip internal runtime markers before constructing provider wire payloads
- cover Anthropic beta/skills selection and Gemini's experimental Interactions converter
- add unit invariants for every native factory/name plus final-wire E2E coverage

## Root cause

The Anthropic, Gemini, and OpenAI converters selected provider-native behavior by `tool.name`. Tool names are public application identifiers, so an ordinary function could accidentally enter a native converter, acquire provider-only request fields or beta headers, and fail or change behavior on only one provider.

Matching versioned provider metadata would fix individual payloads but couple classification to dated wire identifiers. This PR follows the existing OpenRouter and Grok convention instead: adapter-owned metadata carries a stable canonical kind, while provider API versions remain confined to wire converters.

## Adapter details

- **Anthropic:** all seven native factories use stable markers; beta and hosted-skill selection also use the marker rather than `code_execution` by name.
- **Gemini:** all seven native factories use stable markers in both Generate Content and experimental Interactions paths. Interactions now translates the public Google Search factory's camelCase `searchTypes` metadata to its snake_case wire list.
- **OpenAI:** all eleven special Responses tools, including the unbranded `customTool`, use stable markers in `openai-base`; `@tanstack/ai-openai` receives a patch release alongside the base package.

## Test plan

- every provider-native factory still selects its native converter
- every provider-native public name remains valid for an ordinary function
- runtime markers survive plain-data round trips
- runtime markers never reach provider wire payloads
- Anthropic custom `web_search`, OpenAI custom `web_search`, and Gemini custom `google_search` final-wire regressions
- Gemini Interactions reserved-name and public-factory regression coverage

## Summary by CodeRabbit

* **New Features**
  * Provider-native tools now use stable identification across Anthropic, Gemini, and OpenAI integrations.
  * Standard function tools can safely share names with provider-native tools without triggering special behavior.
  * Provider tool identity is preserved through serialization and deserialization.
  * Gemini Google Search settings now translate correctly to the Interactions API format.

* **Bug Fixes**
  * Corrected Anthropic code execution and skills handling for similarly named custom functions.
  * Improved provider tool conversion and metadata handling across supported integrations.

* **Tests**
  * Added coverage for custom-name dispatch, wire formats, and serialized provider tools.


packages/ai-anthropic/src/tools/web-search-tool.ts
```diff
@@ -54,7 +57,7 @@ const validateUserLocation = (tool: WebSearchToolConfig) => {
 export function convertWebSearchToolToAdapterFormat(
   tool: Tool,
 ): WebSearchToolConfig {
-  const metadata = tool.metadata as {
+  const metadata = getAnthropicProviderToolMetadata(tool) as {
     allowedDomains?: Array<string> | null
     blockedDomains?: Array<string> | null
     maxUses?: number | null
```
AlemTuzlak: **[bug]** `webSearchTool` stores the Anthropic SDK config as `metadata` (`allowed_domains`, `max_uses`, `user_location`, `cache_control`). `convertWebSearchToolToAdapterFormat` still reads camelCase (`allowedDomains`, `maxUses`, `userLocation`, `cacheControl`). Those keys are never present, so every search option is dropped. The new smoke test passes `max_uses: 2` and then `toEqual`s a payload with no `max_uses` and `cache_control: null`, so CI now locks in the loss. Docs and the adapter page tell users to pass `max_uses`. Bash, computer use, and text editor return the stored SDK object and do not have this mismatch.

**Suggestion:** After `getAnthropicProviderToolMetadata`, map the snake_case SDK fields (`allowed_domains`, `blocked_domains`, `max_uses`, `user_location`, `cache_control`) onto the wire object. Keep forcing `name: 'web_search'` and `type: 'web_search_20250305'`. Change the smoke test so `max_uses: 2` is present on the converted tool, and stop treating a silent drop as the expected result.

packages/ai-gemini/src/tools/provider-tool-dispatch.test.ts
```diff
@@ -0,0 +1,66 @@
...
+      computerUseTool({}),
+      fileSearchTool({ fileSearchStoreNames: [] }),
+      googleMapsTool(),
+      googleSearchTool(),
+      googleSearchRetrievalTool(),
+      urlContextTool(),
+    ])
+
+    expect(converted).toHaveLength(PROVIDER_TOOL_NAMES.length)
+    expect(JSON.stringify(converted)).not.toContain('__kind')
```
AlemTuzlak: **[suggestion]** Custom-name tests are strong. Factory tests are not. Gemini and OpenAI only check array length and that `JSON.stringify` has no `__kind`. A factory that skips branding, or that writes the wrong kind, still yields length 7 / 11: Gemini groups the miss as one `functionDeclarations` object, OpenAI emits `{ type: 'function', name }`. OpenAI `customTool({ name: 'lookup_order' })` never asserts `type: 'custom'` and `name: 'lookup_order'`. Anthropic's multi-tool test only checks that some `name`s exist. Those names also exist on the custom path, so an unbranded `bashTool` still passes. Only Anthropic `webSearchTool`, Gemini `googleSearchTool` round-trip, Gemini Interactions builtins, and OpenAI `webSearchTool` round-trip assert a real native discriminant.

**Suggestion:** For each factory, assert `metadata.__kind` and the converted native discriminant (Anthropic `type` / `name`, Gemini `googleSearch` / `codeExecution` / and the rest, OpenAI `type` plus `customTool`'s public `name`). Keep the custom-name `it.each` tables. They already prove the collision fix.

.changeset/tidy-searches-coexist.md
```diff
@@ -0,0 +1,12 @@
+---
+'@tanstack/ai-anthropic': patch
+'@tanstack/ai-gemini': patch
+'@tanstack/ai-openai': patch
+'@tanstack/openai-base': patch
+---
+
+Classify Anthropic, Gemini, and OpenAI native tools with stable runtime discriminators so ordinary functions can use the same public names without selecting provider-native behavior.
```
AlemTuzlak: **[suggestion]** After this change, a plain object named `web_search` / `google_search` / `code_execution` is a custom function. Native behavior exists only when the adapter factory sets `metadata.__kind`. That is the intended contract. The changeset is a patch and does not say that hand-built reserved-name tools stop selecting native converters. `docs/tools/provider-tools.md` and the adapter pages still only show factories. They never say that the public name is not the discriminator.

**Suggestion:** Add one sentence to the changeset: native tools must come from the factory, not from a reserved `name`. Add the same sentence to `docs/tools/provider-tools.md` so callers do not keep building `{ name: 'web_search', metadata: ... }` and expect a native payload.

AlemTuzlak (approved):

PR #937, opened 2026-07-14
fix: recover text from completed Responses events

jan-kubica:
## Summary

- recover successful OpenAI Responses text when it is only present on `response.completed`
- ignore empty `response.content_part.added` placeholders so they do not suppress the completion fallback
- cover the behavior at both the adapter and raw-wire E2E boundaries

## Root cause

The Responses adapter treated an empty `output_text` content-part placeholder as streamed content. Some successful responses then carry the populated text only in `response.completed.output`, but that terminal payload was not used as a backstop. Non-streaming callers consequently received an empty string for a successful response.

## Test plan

- `pnpm --filter @tanstack/openai-base test:lib -- responses-text.test.ts`
- `pnpm test:pr`
- `pnpm --filter @tanstack/ai-e2e test:e2e -- tests/openai-completed-response-text.spec.ts`
- `pnpm --filter @tanstack/ai-e2e test:e2e` (345 passed, 1 gated live smoke skipped; 4 unrelated tests passed on retry)

## Summary by CodeRabbit

* **Bug Fixes**
  * Recovered assistant text when OpenAI or OpenRouter provides it only in the final response event.
  * Prevented empty content placeholders from suppressing valid completion text.
  * Improved handling of response text across supported event formats.

* **Tests**
  * Added streaming and end-to-end coverage for completion-only text responses.
  * Verified correct event ordering and returned response content.

### DIFF
--- a/.changeset/quiet-dragons-write.md
+++ b/.changeset/quiet-dragons-write.md
@@ -1,5 +1,6 @@
 ---
 '@tanstack/openai-base': patch
+'@tanstack/ai-openrouter': patch
 ---
 
-Recover successful Responses API text that is present only on the completed response event.
+Recover successful OpenAI and OpenRouter Responses API text that is present only on the completed response event.
--- a/.changeset/quiet-dragons-write.md
+++ b/.changeset/quiet-dragons-write.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/openai-base': patch
+---
+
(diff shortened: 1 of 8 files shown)


AlemTuzlak:
why is this in draft?

jan-kubica:

AlemTuzlak (approved):

tombeckenham:

PR #938, opened 2026-07-14
fix(ai): reject non-streaming chat on run errors

jan-kubica:
`chat({ stream: false })` used to return partial text after a provider `RUN_ERROR`. It now rejects with the same error as the streaming path.

## 🎯 Changes

- `streamToText` throws when the event stream emits `RUN_ERROR`.
- `StreamProcessor` and `streamToText` share `runErrorEventToError`, so the consumer `Error` keeps `code` and `rawEvent`.
- A unit test and a mocked OpenAI E2E route cover the rejection.
- Rebased onto `main` after #937. The only conflict was generated `testing/e2e/src/routeTree.gen.ts`. Both E2E routes stay.

## ✅ Checklist

- [ ] Docs: I updated `docs/` for this change, or this change is not user-facing.

This is user-facing behavior, but there is no existing docs page for the non-streaming error path. The changeset is the user-visible note. I did not add a new docs page in this conflict rebase.

## 🚀 Release Impact

## Testing

**Commands run**

1. `pnpm --filter @tanstack/ai exec vitest run tests/stream-to-response.test.ts` — 30 passed
2. `pnpm --filter @tanstack/ai test:types` — passed after a workspace build
3. `pnpm --filter @tanstack/ai test:oxlint` — 0 errors
4. Skipped `pnpm test:pr` and the E2E spec in this pass (conflict was generated route tree only)

**Manual test**

1. Call `chat({ stream: false })` against a provider that returns HTTP 429.
2. Confirm the promise rejects. Old behavior: empty or partial string.
3. Confirm `error.message` and `error.code` match the provider.

- `packages/ai/tests/stream-to-response.test.ts` rejects on `RUN_ERROR`
- `testing/e2e/tests/non-streaming-run-error.spec.ts` hits a mocked 429

## Risk / rollback

Low. Revert the PR. Streaming `onError` behavior does not change.

## Public API change

**Before**

```ts
const text = await chat({ adapter, messages, stream: false })
// text can be "" or partial after a provider RUN_ERROR
```

**After**

```ts
const text = await chat({ adapter, messages, stream: false })
// throws Error with message, code, and rawEvent from RUN_ERROR
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Non-streaming chat requests now reject when a provider reports a run error instead of returning incomplete text.
  * Error responses preserve the provider’s message, error code, and underlying event details.
  * Added a fallback message when run-error details are unavailable.

* **Tests**
  * Added coverage for non-streaming provider errors, including rate-limit failures.

### DIFF
--- a/.changeset/calm-lions-throw.md
+++ b/.changeset/calm-lions-throw.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/ai': patch
+---
+
+Reject non-streaming text calls when their event stream ends with a `RUN_ERROR`.
--- a/packages/ai/src/activities/chat/stream/processor.ts
+++ b/packages/ai/src/activities/chat/stream/processor.ts
@@ -22,8 +22,9 @@ import {
   generateMessageId,
   uiMessageToModelMessages,
 } from '../messages.js'
-import { normalizeToolResult } from '../../../utilities/tool-result'
+import { runErrorEventToError } from '../../../utilities/errors'
 import { isProviderExecutedToolCall } from '../../../utilities/provider-executed'
+import { normalizeToolResult } from '../../../utilities/tool-result'
 import { defaultJSONParser } from './json-parser'
 import {
   appendStructuredOutputDelta,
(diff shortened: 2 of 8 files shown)


AlemTuzlak:
why is this in draft?

jan-kubica:

tombeckenham:

AlemTuzlak (approved):

PR #939, opened 2026-07-14
fix: undo strict null widening for tool inputs

jan-kubica:
OpenAI-compatible strict tools promote optional fields to required nullable fields. The model can then send `null` for an omitted optional. TanStack AI used to validate that `null` against the original schema, so `.optional()` failed and the tool did not run.

This PR records which nulls the converter added, then removes only those nulls before the tool runs. A genuine `.nullable()` `null` stays.

## Changes

- Record a `NullWideningMap` during the same OpenAI strict-schema conversion that adds nullability.
- Remove only provider-synthesized nulls before emitting completed tool input.
- Apply that undo on Chat Completions and Responses, including drain and backfill paths.
- Skip non-strict tools and duplicate tool names.
- Use the same converter for the inverse map as the wire schema, including Groq's override.
- Adapter tests assert the exact tool input with `toEqual`.
- Tools docs state the optional vs nullable contract.

## Checklist

## Release Impact

## Testing

Commands run on the review follow-up:

1. `pnpm --filter @tanstack/openai-base test:lib` — 203 passed
2. `pnpm --filter @tanstack/ai-groq test:lib` — 36 passed
3. `pnpm --filter @tanstack/openai-base test:types` and `pnpm --filter @tanstack/ai-groq test:types` — passed
4. `pnpm --filter @tanstack/openai-base test:oxlint` and `pnpm --filter @tanstack/ai-groq test:oxlint` — passed (existing `any` warnings only)

I did not re-run `pnpm test:pr` or the e2e spec after the follow-up commit.

Manual test:

1. Define a strict OpenAI-compatible tool with `.optional()` and `.nullable()` fields.
2. Call it so the model omits the optional field (wire value `null`) and sets the nullable field to `null`.
3. Confirm the tool handler receives the optional field as absent and the nullable field as `null`.
4. Confirm the tool executes and the next model turn runs.

How this PR makes testing easy: adapter unit tests, a normalizer unit test for a custom converter, and `testing/e2e/tests/openai-strict-tool-null-wire.spec.ts`.

## Risk / rollback

Low. If a subclass converter and the inverse map still disagree, an optional `null` can fail schema validation again. Revert the PR to undo.

## Public API change

Tool authors do not change call sites. Optional vs nullable input now matches the schema:

```ts
inputSchema: z.object({
  mode: z.enum(["canary"]).optional(),
  nullableNote: z.string().nullable(),
})
```

Adapter authors who change strict conversion should override `makeStructuredOutputCompatibleWithMap` so the inverse map matches the wire schema. `@tanstack/openai-base` also exports `makeStructuredOutputCompatibleWithMap`.


tombeckenham:

tombeckenham:

tombeckenham:

packages/openai-base/tests/chat-completions-text.test.ts
```diff
@@ -451,6 +451,69 @@ describe('OpenAIBaseChatCompletionsTextAdapter', () => {
...
+        model: 'test-model',
+        messages: [{ role: 'user', content: 'Ask me' }],
+        tools: [strictTool],
+      })) {
+        chunks.push(chunk)
+      }
+
+      expect(
+        chunks.find((chunk) => chunk.type === 'TOOL_CALL_END'),
+      ).toMatchObject({
```
AlemTuzlak: **[suggestion]** The new Chat Completions and Responses adapter tests are named as proof that strict null-widening is undone, but they assert with `toMatchObject`. That matcher allows extra keys, so `{ question, options: null, nullableNote: null }` (widening not undone) still passes. The normalizer unit tests and the e2e spec use `toEqual` and would catch this; these adapter tests would not.

**Suggestion:** Assert with `toEqual({ question: 'Which one?', nullableNote: null })` in both `packages/openai-base/tests/chat-completions-text.test.ts` and `packages/openai-base/tests/responses-text.test.ts` (same pattern at responses-text.test.ts:749).

packages/openai-base/src/utils/tool-input-normalizer.ts
```diff
@@ -0,0 +1,47 @@
...
+    seenNames.add(tool.name)
+
+    const inputSchema = (tool.inputSchema ?? {
+      type: 'object',
+      properties: {},
+      required: [],
+    }) as JSONSchema
+    if (!isStrictModeCompatible(inputSchema)) continue
+
+    const { nullWideningMap } = makeStructuredOutputCompatibleWithMap(
```
AlemTuzlak: **[suggestion]** Wire schemas are built with `this.makeStructuredOutputCompatible.bind(this)` in both adapters, but the inverse map always comes from the default `makeStructuredOutputCompatibleWithMap`. Groq already overrides the hook (`normalizeObjectSchemas` + strip empty `required`). Today that pre-pass does not add extra nulls, so the maps still match. A later subclass that widens more fields than the default converter would send a stricter wire schema than the map describes, and `undoNullWidening` would leave those extra `null`s in place.

**Suggestion:** Build the map with the same converter the request used (pass `this.makeStructuredOutputCompatible` into `createToolInputNormalizer`, or have the hook return `{ schema, nullWideningMap }`).

.changeset/calm-tools-return.md
```diff
@@ -0,0 +1,5 @@
+---
+'@tanstack/openai-base': patch
+---
+
+Undo strict-mode null widening before validating and executing OpenAI-compatible tool calls.
```
AlemTuzlak: **[suggestion]** This is user-facing: optional tool fields on OpenAI-compatible strict tools used to fail Standard Schema validation, and now they execute. `docs/structured-outputs/overview.md` already documents the same undo for `outputSchema`, but `docs/tools/tools.md` still shows `.optional()` tool fields with no note that OpenAI-compatible adapters strip provider-synthesized `null`s and keep genuine `.nullable()` `null`s.

**Suggestion:** Add a short note on the tools page (same contract as the structured-output note): optional tool inputs come back absent, nullable inputs keep `null`.

PR #940, opened 2026-07-15
feat(ai-mcp): support task-based tool execution

### LINKED ISSUE #704: feat(ai-mcp): support task-based tool execution (execution.taskSupport)
## Context

#700 ships `@tanstack/ai-mcp` with plain `callTool()` execution. The MCP spec's experimental **tasks** feature lets servers mark tools with `execution.taskSupport: 'required' | 'optional' | 'forbidden'`. Task-required tools reject plain `tools/call` with `-32600` ("requires task-based execution") — e.g. `simulate-research-query` on recent `@modelcontextprotocol/server-everything`.

**Current v1 behavior** (since the task-required-tool fix on #700):
- `tools()` auto-discovery **excludes** task-required tools (they could never succeed; offering them gives the model a guaranteed-failure tool).
- Explicitly binding one via `tools([toolDefinition(...)])` throws `MCPTaskRequiredToolError`.

## Proposal

Support task-based execution so these tools become usable instead of skipped. The SDK API is a clean fit for our execute proxy:

```ts
const stream = client.experimental.tasks.callToolStream(
  { name, arguments: args },
  undefined,
  { signal: ctx?.abortSignal },
)
for await (const message of stream) {
  // 'taskCreated' | 'taskStatus' | 'result' | 'error' — guaranteed terminal result/error
}
```

Minimal version: branch in `makeMcpExecute` — task-required tools drain `callToolStream` to the terminal message; the terminal `result` is the same `CallToolResult` shape we already normalize (`structuredContent` / `mcpContentToTanstack`). Stop excluding them from discovery.

Optional second step: surface `taskStatus` updates through `ToolExecutionContext.emitCustomEvent` so UIs can render long-running tool progress in the transcript.

## Why it was deferred from #700

The tasks API lives under `client.experimental.*` ("may change without notice"). Wiring the core execution path of a brand-new package to unstable SDK surface means a routine SDK bump could break tool execution. The v1 exclusion only depends on stable spec types (`Tool.execution.taskSupport`).

## Design questions to settle

- [ ] Should `taskSupport: 'optional'` tools also route through tasks (the SDK server does auto-polling for optional), or stay on plain `callTool`?
- [ ] On chat-run abort: drop the stream, or also call `cancelTask`?
- [ ] Expose `getTask` / `listTasks` / `cancelTask` on the `MCPClient` surface?
- [ ] Surface `taskCreated`/`taskStatus` as custom events in the chat stream (devtools/UI rendering)?
- [ ] How to pin/guard against `experimental` API drift across SDK versions (peer-dep range, runtime feature detection)?

## References

- Exclusion + `MCPTaskRequiredToolError` introduced in #700 (`packages/ai-mcp/src/tools.ts` `requiresTaskExecution()`, `packages/ai-mcp/src/client.ts` defs path)
- SDK: `@modelcontextprotocol/sdk/dist/esm/experimental/tasks/client.d.ts` (`ExperimentalClientTasks.callToolStream`)
- Repro: `npx -y @modelcontextprotocol/server-everything` → `simulate-research-query`

season179:
MCP tools with `execution.taskSupport: 'required'` now run through the SDK task stream. Auto-discovery, explicit `tools([defs])`, `MCPClient.callTool()`, and MCP Apps all use this path. Optional-task tools stay on ordinary `callTool`.

Abort stops the local wait at once and sends a best-effort `tasks/cancel`. A later `tools/list` page no longer wipes SDK metadata for earlier tools.

Closes #704

## Changes

- Run task-required MCP tools through `callToolStream`
- Keep optional-task tools on ordinary `callTool`
- Skip or throw `MCPTaskRequiredToolError` when the server has no `tools/call` tasks capability
- Paginate `tools/list` without wiping earlier-page SDK validators
- Add Tasks mode in the `ts-react-chat` MCP demo

## Checklist

## Release Impact

## Testing

**Commands run**

1. `pnpm --filter @tanstack/ai-mcp exec vitest run tests/client.test.ts tests/tools.test.ts` — 62 passed
2. `pnpm --filter @tanstack/ai-mcp test:oxlint` — 0 errors
3. `pnpm --filter @tanstack/ai-mcp test:types` and `pnpm --filter @tanstack/ai-mcp run build` — passed
4. Error-path e2e routes on port 3012 returned the expected JSON (Playwright on 3010 was blocked: port in use)
5. Did not run `pnpm test:pr`

**Manual test**

1. Run `pnpm dev:chat`
2. Open `/mcp-demo` and select **Tasks**
3. Send `Appraise my guitar collection: strat, tele, jazzmaster`
4. Confirm the tool stays pending about 4 seconds, then the total is `$4200`
5. Send again, click **Stop** while pending. The chat must stop in under a few seconds.

Unit tests cover pagination cache, cursor loops, and abort timing. E2E routes cover abort and missing tasks capability.

## Linked issues

Closes #704

## Risk / rollback

A server that ignores `tasks/cancel` can keep a remote task until TTL. Revert this PR to undo.

## Public API change

**Before**

```ts
await client.callTool('slow_research', { query })
// threw MCPTaskRequiredToolError
```

**After**

```ts
await client.callTool('slow_research', { query }, { signal })
// runs the task stream when the server declares tools/call tasks
```


packages/ai-mcp/tests/tools.test.ts
```diff
@@ -163,7 +243,7 @@ describe('makeMcpExecute', () => {
     )
     expect(callTool).toHaveBeenCalledWith(
       { name: 'x', arguments: {} },
-      undefined,
+      expect.anything(),
```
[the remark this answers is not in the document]
season179: Fixed in b38fcfdd. The wildcard was the second `callTool` argument (the result schema), so it now exactly asserts `CallToolResultSchema`. The third argument already exactly asserts `{ signal: controller.signal }`; using `{ signal: undefined }` there would contradict this test’s live-signal setup.

season179:
Review follow-up on the task-fixture nitpick: I verified the suggested callback counters against the installed `@modelcontextprotocol/sdk@1.29.0`. Its `tasks/get` and `tasks/getResult` protocol handlers read `taskStore` directly, so the `registerToolTask` `getTask`/`getTaskResult` callbacks are not invoked; a counter assertion against those callbacks fails even though the terminal result completes correctly. The handler-routing change shown in SDK PR #1335 is not present in 1.29.0.

I therefore did not add impossible callback assertions or extract a shared helper across the `packages/ai-mcp` test boundary and the E2E runtime app. The existing unit result assertion and Streamable HTTP E2E `4200` assertion both require the SDK task status/result polling path to complete successfully.

AlemTuzlak:
can you create an example in ts-react-chat so i cna manually test this out?

season179:
Done — added a **Tasks** mode to the ts-react-chat MCP demo.

**To test:**
1. `cd examples/ts-react-chat && pnpm dev` — needs an LLM key for whichever provider you pick in the UI (`OPENAI_API_KEY`, `ANTHROPIC_API_KEY`, `GOOGLE_API_KEY`, `GROQ_API_KEY`, or `OPENROUTER_API_KEY`)
2. Open http://localhost:3000/mcp-demo and switch the mode to **Tasks**
3. Ask: *"Appraise my guitar collection: strat, tele, jazzmaster"*

You should see the tool call sit pending for ~4s while the client polls, then the result lands and the model summarizes it ($1400/guitar → $4200 for that prompt).

How it works: the demo hosts an in-process Streamable HTTP MCP server at `/api/mcp-tasks-server` whose only tool declares `execution.taskSupport: 'required'`, so `@tanstack/ai-mcp` routes the call through the task flow (create task → poll every 500ms → fetch result) instead of plain `tools/call`. The chat route is a standard `chat({ mcp })` setup — nothing task-specific needed on that side.

Heads up: the first of the two new commits fixes a pre-existing dev crash on `/mcp-demo` (`standardwebhooks` CJS interop, pulled into the client bundle via `@anthropic-ai/sdk` which sits in vite's `optimizeDeps.exclude`) — the page didn't load in dev at all, so manual testing needed it. It splits adapter resolution out of `mcp-providers.ts` into a server-only module.

packages/ai-mcp/src/client.ts
```diff
@@ -212,9 +274,32 @@ class MCPClientImpl<
...
+    const definition = this.#toolDefinitions?.get(name)
+    const taskRequired =
+      definition !== undefined &&
+      requiresTaskExecution(definition) &&
+      serverSupportsTaskCalls(this.#client)
+    return callMcpTool(
+      this.#client,
+      name,
+      args ?? {},
+      taskRequired,
```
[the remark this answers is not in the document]
season179: Fixed in 18e53596 — `callTool` now throws `MCPTaskRequiredToolError` when the resolved definition requires task execution but the server doesn't declare the tasks capability for `tools/call`, mirroring the `tools([...defs])` binding guard. Added a regression test using `makeServerWithUnsupportedTaskTool`: `callTool('needs_tasks')` rejects with the error while `plain_tool` on the same server keeps working.

packages/ai-mcp/src/tools.ts
```diff
@@ -50,8 +51,57 @@ export function mcpContentToTanstack(
...
+  for await (const message of stream) {
+    if (message.type === 'taskCreated') taskId = message.task.taskId
+    if (message.type === 'result') return message.result
+    if (message.type === 'error') {
+      // On abort the SDK merely stops polling; the server-side task keeps
+      // running until its TTL. Propagate the abort as a best-effort cancel
+      // (never masking the original error with a cancel failure).
+      if (signal?.aborted && taskId !== undefined) {
+        await client.experimental.tasks.cancelTask(taskId).catch(() => {})
+      }
```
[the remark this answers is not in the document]
season179: Keeping the cancel — it's intentional, for three reasons:

1. **Consistency with every other tool call.** For ordinary (non-task) calls, aborting the signal already cancels server-side work: the SDK sends the protocol's `notifications/cancelled` for the in-flight request. Task-required tools are a transparent execution-mode detail here, so abort propagating to `tasks/cancel` gives them the same semantics instead of a surprising exception.

2. **The result is unretrievable after abort.** `callMcpTool` never exposes the `taskId` to callers — once the local call aborts, no consumer can ever fetch the task result. Leaving the task running would burn server work with no possible beneficiary (e.g. a user cancelling an expensive research task would keep paying for it until TTL).

3. **It's best-effort and non-breaking.** The cancel is guarded (`signal.aborted` + a created task) and failures are swallowed, so servers that reject `tasks/cancel` lose nothing.

If a fire-and-resume flow ever surfaces the `taskId` to callers, that path shouldn't route through this helper anyway. The behavior is documented in `docs/tools/mcp.md`, `mcp-manual.md`, and the changeset.

tombeckenham:
Thanks for the PR, @season179! 🙌 @AlemTuzlak will take a look.

- ✅ CI passing
- ⚠️ Merge conflicts with `main` — please rebase
- ✅ Changeset present
- ✅ E2E test changes included

tombeckenham:

tombeckenham:

packages/ai-mcp/src/client.ts
```diff
@@ -145,13 +161,46 @@ class MCPClientImpl<
...
+  async #listTools(options?: { raw?: boolean }): Promise<Array<McpToolDef>> {
+    const defs: Array<McpToolDef> = []
+    let cursor: string | undefined
+    do {
+      const page = options?.raw
+        ? await this.#client.request(
+            { method: 'tools/list', ...(cursor ? { params: { cursor } } : {}) },
+            ListToolsResultSchema,
+          )
+        : await this.#client.listTools(cursor ? { cursor } : undefined)
```
AlemTuzlak: **[bug]** `#listTools()` without `raw` paginates by calling `Client.listTools()` on every page. The SDK's `listTools()` runs `cacheToolMetadata()`, which **clears** then replaces the client's output-schema validators and `_cachedRequiredTaskTools` with **only the current page**. After a two-page `tools()` (the path `chat()` and MCP Apps use), tools from earlier pages lose SDK structured-content validation, and `Client.callTool`'s `isToolTaskRequired` guard no longer knows about required tools on those pages. This PR's own `Map` still routes correctly because `callToolStream` is given `{ task: {} }`, but the changeset claim that output-schema validation is unchanged does not hold for paginated servers. The new pagination test uses two pages with no `outputSchema`, so it cannot catch this.

**Suggestion:** Keep the first page on `this.#client.listTools()` if you still want the SDK validators armed for discovered tools, and fetch every later page with the raw `this.#client.request({ method: 'tools/list', params: { cursor } }, ListToolsResultSchema)` path so `cacheToolMetadata` is not called again. Alternatively, always list raw and accept that SDK-side output validation stays off (matching direct `callTool`).

packages/ai-mcp/src/tools.ts
```diff
@@ -91,8 +92,57 @@ export function mcpContentToTanstack(
...
+  let taskId: string | undefined
+  for await (const message of stream) {
+    if (message.type === 'taskCreated') taskId = message.task.taskId
+    if (message.type === 'result') return message.result
+    if (message.type === 'error') {
+      // On abort the SDK merely stops polling; the server-side task keeps
+      // running until its TTL. Propagate the abort as a best-effort cancel
+      // (never masking the original error with a cancel failure).
+      if (signal?.aborted && taskId !== undefined) {
+        await client.experimental.tasks.cancelTask(taskId).catch(() => {})
```
AlemTuzlak: **[bug]** On abort, `callMcpTool` **awaits** `client.experimental.tasks.cancelTask(taskId)` before rethrowing. `cancelTask` is a normal SDK `request()` with the default 60s timeout, and this call does not pass a short timeout (it also must not reuse the already-aborted signal, or cancel would fail immediately). If the server is slow or the session is half-dead after abort, the chat run's abort path can block for up to a minute instead of returning the abort error. That is not best-effort. The in-memory abort test cannot catch this because the test server answers `tasks/cancel` at once.

**Suggestion:** Treat cancel as fire-and-forget (`void client.experimental.tasks.cancelTask(taskId).catch(() => {})`), or pass a short `{ timeout: 1_000 }` (or similar) so abort cannot stall on the cleanup RPC. Keep swallowing cancel failures so they never mask the original abort error.

packages/ai-mcp/skills/ai-mcp/SKILL.md
```diff
@@ -449,9 +449,11 @@ await using pool = await createMCPClients({
...
 ## Abort signal — cancelling in-flight MCP calls
 
-MCP tool calls are automatically cancelled when the chat run's `AbortController`
-fires (e.g. client disconnect, server abort). The `abortSignal` is threaded
-through `ToolExecutionContext` into every `callTool` call with no extra code.
+TanStack AI stops waiting for MCP tool calls when the chat run's
+`AbortController` fires (e.g. client disconnect, server abort). The
+`abortSignal` is threaded through `ToolExecutionContext` into every tool call
+with no extra code. For a task-required tool, aborting stops the local task
+stream but does not cancel a remote task the MCP server has already created.
```
AlemTuzlak: **[suggestion]** The skill says aborting a task-required tool "stops the local task stream but does not cancel a remote task the MCP server has already created." The implementation, changeset, `docs/tools/mcp.md`, and `docs/tools/mcp-manual.md` all send a best-effort `tasks/cancel`. Agents following this skill will document the wrong abort contract.

**Suggestion:** Match the product docs: abort stops waiting on the local stream and sends best-effort `tasks/cancel` for a task that was already created. Say clearly that cancel is best-effort and that a server which ignores `tasks/cancel` may keep running until TTL.

examples/ts-react-chat/src/routes/api.mcp-tasks-server.ts
```diff
@@ -0,0 +1,152 @@
...
+          ttl: taskRequestedTtl ?? 60_000,
+          pollInterval: TASK_POLL_INTERVAL_MS,
+        })
+        console.log(
+          `[mcp-tasks] created task ${task.taskId} — appraising ${ids.length} guitar(s), completes in ${APPRAISAL_DURATION_MS}ms`,
+        )
+        // Complete the task in the background; the client keeps polling
+        // getTask until the status turns terminal. If the user aborts the
+        // chat mid-task, this timer still fires and the stored result is
+        // simply never fetched (documented ai-mcp abort semantics).
```
AlemTuzlak: **[suggestion]** The demo `createTask` comment says that if the user aborts mid-task, "this timer still fires and the stored result is simply never fetched (documented ai-mcp abort semantics)." That was true of an earlier design. The library now sends `tasks/cancel`, which marks the in-memory task cancelled; the 4s `setTimeout` still calls `storeTaskResult`, which then fails because the task is already terminal (the `.catch` logs it). The comment teaches the opposite of current client behavior.

**Suggestion:** Say that abort sends `tasks/cancel` (best-effort), and that this demo timer is not wired to that cancel — so it may still fire and `storeTaskResult` will fail on a cancelled task. Or clear/ignore the timer when the store reports `cancelled`.

packages/ai-mcp/skills/ai-mcp/SKILL.md
```diff
@@ -765,11 +767,9 @@ and do NOT appear in the library's runtime dependency graph.
   methods after `close()`.
 - `MCPToolNotFoundError` — thrown from `client.tools([defs])` when a definition's
   `name` is not exposed by the server.
-- `MCPTaskRequiredToolError` — thrown from `client.tools([defs])` when the named
-  tool declares `execution.taskSupport: 'required'` (experimental MCP tasks).
-  Such tools only run via the SDK's `tasks/callToolStream` flow, which
-  `@tanstack/ai-mcp` does not support yet; they are silently excluded from
-  `tools()` auto-discovery for the same reason.
+- `MCPTaskRequiredToolError` — deprecated compatibility export. Task-required
```
AlemTuzlak: **[suggestion]** The skill calls `MCPTaskRequiredToolError` a "deprecated compatibility export" and says `client.tools()` no longer throws it. The class is not `@deprecated`. Auto-discovery no longer throws it for a capable server, but `tools([defs])` and `callTool()` still throw it when a task-required tool is bound/called and the server does not declare `tasks.requests.tools.call`. `docs/tools/mcp.md` already describes that meaning.

**Suggestion:** Drop "deprecated compatibility export." Document the current throw sites: explicit bind or `callTool` of a task-required tool on a server without the `tools/call` tasks capability; auto-discovery skips those tools instead.

packages/ai-mcp/src/client.ts
```diff
@@ -145,13 +161,46 @@ class MCPClientImpl<
...
+      defs.push(...page.tools)
+      cursor = page.nextCursor
+    } while (cursor)
```
AlemTuzlak: **[suggestion]** The new pagination loop is `do { ... cursor = page.nextCursor } while (cursor)` with no cap and no check that the cursor changed. A server that keeps returning the same `nextCursor` (or an endless chain) will hang `tools()` and the lazy `callTool` listing forever. That hang did not exist when discovery used a single `listTools()` call.

**Suggestion:** Stop when `nextCursor` is missing, equals the cursor just used, or a page limit is exceeded, and fail that listing (callTool can still fall back to a plain `tools/call`).

AlemTuzlak (approved):
The earlier review findings are fixed in 6bd261be1. Paginated tools/list keeps earlier-page SDK metadata. Abort stops the local wait at once and sends best-effort tasks/cancel. Error-path e2e covers abort and missing tasks capability. Approving.

AlemTuzlak:
@season179 thank you, sorry for the long wait!

PR #941, opened 2026-07-15
feat(ai-openrouter): surface generation metadata on RUN_FINISHED

trevorWieland:
## 🎯  Changes

Expose OpenRouter generation IDs and selected providers on `RUN_FINISHED`.

Metadata is captured in-band from response chunks, following the cost handling in #654. The non-streaming structured-output fallback forwards the same fields.

## Test plan

- `pnpm test:pr`
- `pnpm --filter @tanstack/ai-e2e test:e2e`
- Added chat and Responses coverage for streaming and structured output.
- Live `/generation` lookup pending API-key access.

## ✅  Checklist

- [x] I have followed the contributing guide.

## 🚀  Release Impact

- [ ] This change is docs/CI/dev-only.

## Summary by CodeRabbit

* **New Features**
  * OpenRouter structured outputs now include a generation ID and serving provider when available.
  * `RUN_FINISHED` events surface generation IDs and provider details for streamed and non-streamed responses.
  * Provider and generation metadata is preserved alongside usage information, including trailing stream events.

* **Tests**
  * Added coverage across structured output, streaming, cost tracking, and end-to-end scenarios.


tombeckenham:
Thanks for the PR, @trevorWieland! 🙌 @AlemTuzlak will take a look.

- ✅ CI passing
- ✅ No merge conflicts
- ✅ Changeset present
- ✅ E2E test changes included

tombeckenham:

tombeckenham:

AlemTuzlak (commented):
I treated this diff as invasive. There is no linked GitHub issue. This is a feature, in the same shape as #654 (cost on `RUN_FINISHED`).

**Does the feature earn its keep?** Yes, in a small form. OpenRouter's `GET /api/v1/generation?id=` needs the `gen-...` id from the chat-completions `id` field. Putting that on `RUN_FINISHED` is the same pattern as `usage.cost`.

**Keep**
- `generationId` on `RunFinishedEvent` and `StructuredOutputResult`.
- Forwarding those fields in `fallbackStructuredOutputStream`.
- `extractServedProvider` in a shared helper. Two adapters need it.
- Piggybacking the cost E2E instead of a new route.
- The changeset (minor on both packages).

**Keep, but the name is a problem**
- `provider` on the core `RUN_FINISHED` event. In this repo, `provider` already means the adapter (`openai`, `openrouter`). Here it means the upstream that OpenRouter picked (`DeepInfra`). Callers will mix those up. If this field stays, the docs must say it is the served upstream, not the adapter. A name like `servedProvider` would be clearer. I did not rename it in this pass.

**Drop / do not do**
- CodeRabbit: type this with `OpenRouterMetadata` from the SDK. Chunks are `unknown` at the call site. `cost.ts` already walks the same objects with `asRecord`. The generated type does not remove that walk.
- CodeRabbit: use `!= null` so empty strings pass through. An empty generation id is not useful. Omitting it is correct.

**Unverified**
- Live `/generation` lookup. The PR says this is still pending. Chat Completions `chunk.id` is the `gen-...` id OpenRouter documents. Responses uses `response.id`, which can be a `resp_...` id. I did not prove that id works with `GET /generation`. That is the claim the feature is for.
- Public `docs/adapters/openrouter.md` still only shows `usage.cost`. This change is user-facing and has no docs page update.

**What I ran:** I did not execute the new unit tests in this pass. I read the adapter, the core types, the fallback stream, and the tests. The tests assert the adapter copies `id` and `openrouterMetadata.endpoints.available[].provider` onto `RUN_FINISHED`. They do not hit a real OpenRouter `/generation` lookup.

PR #948, opened 2026-07-15
fix(ai-isolate-quickjs): replace asyncify host functions with promise bridging 

### LINKED ISSUE #947: [ai-isolate-quickjs] Two sequential tool calls crash the WASM module (Assertion failed: p->ref_count == 0)
### TanStack AI version

@tanstack/ai@0.40.0 / @tanstack/ai-isolate-quickjs@0.1.45

### Framework/Library version

None — plain Node.js v22.21.1 (no framework; the bug reproduces in a standalone Node script using only @tanstack/ai-isolate-quickjs and @tanstack/ai-code-mode)

### Which project does this relate to?

@tanstack/ai-isolate-quickjs (Code Mode QuickJS isolate driver)

### Description

If sandboxed code awaits a host tool binding more than once, the QuickJS WASM module crashes with a refcount assertion:

RuntimeError: Aborted(Assertion failed: p->ref_count == 0, at: ../../vendor/quickjs/quickjs.c,5512,free_zero_refcount)

The error can't be caught in practice, and it breaks the shared asyncify WASM module — so every other context in the process stops working too.

Code Mode is built around programs that chain multiple tool calls, so this makes the driver unusable for anything beyond a single tool invocation.

There's a related problem: if a tool binding returns a promise that never settles, execute() hangs forever. The configured timeout never fires, because it only exists as a QuickJS interrupt handler — there's no host-side timer running while the guest sits suspended in an asyncified call.

The root cause is that tool bindings are injected with vm.newAsyncifiedFunction() on a newAsyncContext() VM. Asyncify pauses and resumes the WASM stack on every host await, and doing that repeatedly corrupts QuickJS refcounts — a known upstream bug: justjake/quickjs-emscripten#258.

### Platform

- OS: Windows 11 (also happens on Linux/macOS and in StackBlitz WebContainers — the bug is in the WASM module, not the platform)
- Node.js: v22.21.1
- @tanstack/ai-isolate-quickjs: 0.1.45
- quickjs-emscripten: 0.31.0

### Suggested fix

Stop using asyncify. Instead of newAsyncContext() + newAsyncifiedFunction(), use the plain sync WASM build   (getQuickJS().newContext()) and turn ea host function that returns a QuickJSpromise (vm.newPromise()). The host resolves that promise when the tool finishes, then calls                 runtime.executePendingJobs() so the gue. The WASM stack never gets suspended, so the upstream bug can't happen. This also makes two other things possible: the global cross-context execution queue can go away (it only exists becau one suspension at a time), andevalCodeAsync becomes plain evalCode + resolvePromise.

I have a working patch for this (it als the context, since with promise bridging an interrupted guest could otherwise be left waiting forever) and am happy to open a PR. The patch is published as a fork so it's easy to verify: @silviomollov/ai-isolate-quickjs (https://www.npmjs.com/package/@silviomollov/ai-isolate-quickjs) — swap only the import in the repro and it passes.

https://stackblitz.com/edit/node-qwj94pnm?file=repro.mjs

_No response_

Yes, I am also opening a PR that solves the problem along side this issue

### Terms & Code of Conduct

SilvioYMollov:
Changes

Replaces the Asyncify-based host tool bridge in @tanstack/ai-isolate-quickjs with native QuickJS promises, and makes execution timeouts actually work under the new model.

Problem

Two bugs with the same root cause:

1. Awaiting a tool binding more than once crashes the process with Assertion failed: p->ref_count == 0. The abort kills the shared asyncify WASM module, so every other context dies too. Root cause is a known upstream asyncify bug: justjake/quickjs-emscripten#258.
2. A tool that never settles hangs execute() forever. The timeout is only a QuickJS interrupt handler, and no host-side timer runs while the guest waits on a host call — so it never fires.

// crashes on the second await
const a = await get({ path: "/a" });
const b = await get({ path: "/b" });

Fix

- Use the sync QuickJS build instead of newAsyncContext(). Tool bindings are now plain host functions that return a vm.newPromise(); the host resolves it when the tool finishes and pumps executePendingJobs() to resume the guest. The WASM stack never suspends, so the crash can't happen.
- Race the program promise against a real host-side timer, so stuck executions return TimeoutError instead of hanging. A timeout is terminal for the context: interrupted jobs stay queued in the VM and must never run in a later execution, so outstanding tool calls are cancelled and the VM disposed safely.
- Surface executePendingJobs() errors instead of swallowing them.
- The global cross-context queue becomes per-context (it only existed for asyncify), and tool calls in one execution can run concurrently.

Tests

Regression tests for sequential awaits, Promise.all tool calls, TimeoutError on timeout, and context disposal after timeout.

Related

- Fixes #947
- Published as @silviomollov/ai-isolate-quickjs (https://www.npmjs.com/package/@silviomollov/ai-isolate-quickjs) for verification - swap the import and the repro passes.

## Summary by CodeRabbit

* **New Features**
  * Improved QuickJS tool execution with promise-based host bridging and enhanced support for concurrent tool calls.
* **Bug Fixes**
  * Refined timeout handling: timeouts now emit a dedicated `TimeoutError`, cancel pending tool calls, and make the execution context terminal (subsequent executions fail with `DisposedError`).
  * Improved handling of in-flight tool calls, including proper cleanup and safer disposal during mid-execution timeouts.
* **Documentation**
  * Updated migration guidance for the required QuickJS WASM variant and clarified timeout terminality.
* **Tests**
  * Added/expanded coverage for timeout normalization, concurrent calls, post-timeout disposal, and resilience after OOM/late settlements.


packages/ai-isolate-quickjs/src/isolate-context.ts
```diff
@@ -76,10 +158,37 @@ export class QuickJSIsolateContext implements IsolateContext {
       this.vm.dispose()
```
tombeckenham, line 170: **Critical — FreeRuntime hole on fatal limits**

`releaseAfterTimeout` and `dispose()` both cancel `pendingCancels` before `vm.dispose()` because unsettled QuickJS deferreds abort the **shared** WASM module (`list_empty(&rt->gc_obj_list)`). This path does not.

Repro shape:

```js
const p = hangTool({})              // registers deferred + pendingCancel
return 'x'.repeat(8 * 1024 * 1024)  // MemoryLimitError while deferred still live
```

Existing OOM tests never leave a tool in flight, so this is unguarded.

**Suggested fix:** best-effort cancel + grace tick (same as timeout/dispose) before free; if the sick heap cannot cancel, catch and prefer intentional leak over aborting the shared module. See also #946’s `requestVmDispose()`, which already cancels on both timeout and fatal limits.

**Test to add:** start a never-resolving tool, then OOM (or stack overflow); assert a **new** context from the same driver still executes successfully.

packages/ai-isolate-quickjs/src/isolate-driver.ts
```diff
@@ -34,6 +37,112 @@ export interface QuickJSIsolateDriverConfig {
...
+      )
+    execState.pendingCancels.add(cancel)
+
+    void invokeBinding(binding, argsJson).then(resolveWithPayload)
+
+    // Resume guest code waiting on the promise. Defense in depth: outside an
+    // active execute() the interrupt deadline is 0, so a stray job from an
+    // abandoned execution is interrupted instead of running unbounded.
+    void promise.settled.then(() => {
+      if (!vm.runtime.alive) return
```
tombeckenham: **Important — early return skips `promise.dispose()`**

If `settled` runs after the runtime is already dead, this returns **before** the `try/finally` that calls `promise.dispose()`. Paired with the fatal path that frees without cancelling deferreds, host-side deferred lifetimes can outlive FreeRuntime.

Please dispose unconditionally, e.g.:

```ts
void promise.settled.then(() => {
  try {
    if (vm.runtime.alive) {
      const jobs = vm.runtime.executePendingJobs()
      // ...
    }
  } finally {
    if (promise.alive) promise.dispose()
  }
})
```

packages/ai-isolate-quickjs/src/isolate-context.ts
```diff
@@ -92,24 +201,39 @@ export class QuickJSIsolateContext implements IsolateContext {
...
+            },
+            () => {
+              guestSettled = true
+            },
+          )
+          const jobs = this.vm.runtime.executePendingJobs()
+          if (jobs.error) {
+            const dumped: unknown = this.vm.dump(jobs.error)
+            jobs.error.dispose()
+            return await fail(dumped)
```
tombeckenham: **Important — dirty context reuse after early `jobs.error`**

Here `guestSettled` is already `false` and `nativePromise` is never awaited. `fail()` only terminal-disposes for `TimeoutError` / memory / stack.

For any other `jobs.error`, the context stays reusable while the prior program promise / jobs may still be pending — next `execute()` can interleave leftover work under a fresh deadline.

Prefer one of: await/cancel the program promise, treat as terminal (dispose), or fully drain/cancel outstanding work before returning a non-terminal error.

packages/ai-isolate-quickjs/src/isolate-context.ts
```diff
@@ -65,7 +143,11 @@ export class QuickJSIsolateContext implements IsolateContext {
     this.executing = true
     this.logs.length = 0
 
-    const releaseVmAfterFatalLimit = () => {
+    // True until a program promise is in flight; sync failure paths (parse
+    // errors, interrupted straight-line code) leave no unsettled guest state.
```
tombeckenham: **Nit — comment contradicts busy-loop timeout path**

This says interrupted straight-line code leaves no unsettled guest state, but the busy-loop timeout sets `guestSettled = false` before `executePendingJobs()` and relies on the intentional VM **leak** when it stays false (`releaseAfterTimeout`).

Suggest rewriting along the lines of: starts true; cleared once `resolvePromise` is armed; stays false if the program promise never settles (e.g. interrupt during a spin); only pre-arm failures leave it true so timeout release can dispose.

packages/ai-isolate-quickjs/tests/isolate-driver.test.ts
```diff
@@ -148,7 +240,53 @@ describe('createQuickJSIsolateDriver', () => {
...
+        return 1;
+      `)
+      expect(first.success).toBe(false)
+      expect(first.error?.name).toBe('TimeoutError')
+
+      const second = await context.execute('return 42')
+      expect(second.success).toBe(false)
+      expect(second.error?.name).toBe('DisposedError')
+
+      await expect(context.dispose()).resolves.toBeUndefined()
```
tombeckenham, line 267: **Important — assert process-level WASM health after timeout cancel**

This correctly checks context terminality (`DisposedError` on reuse). The dispose + abandoned-tool test below goes further and proves a **new** context still works after FreeRuntime-sensitive cleanup.

Please extend this path similarly:

1. After hanging-tool `TimeoutError`, `createContext` + `execute` on the same driver succeeds.
2. Optionally: host tool resolves **after** timeout cancel; second settle must no-op without poisoning the module.

That locks the same invariant the dispose path already guards.

tombeckenham (changes requested):
## Review summary

Strong root-cause fix for [#947](https://github.com/TanStack/ai/issues/947) / [quickjs-emscripten#258](https://github.com/justjake/quickjs-emscripten/issues/258). Promise bridging, host-side deadlines, terminal timeouts, and abandoned-tool dispose are the right direction — and the regression suite is substantially better than the competing PR.

**Requesting changes** for one FreeRuntime hole and a few important follow-ups before merge.

### Freshness
- Behind `main` by ~6 commits; **MERGEABLE**, no real conflict with `packages/ai-isolate-quickjs` (only a version bump on main).
- Still fully applicable; soft rebase is fine, not blocking.
- Prefer this PR over [#946](https://github.com/TanStack/ai/pull/946) as the landing vehicle (more complete lifecycle, migration notes, broader tests). One idea worth porting from #946: cancel pending host deferreds on **fatal** OOM/stack, not only timeout.

### Must fix
1. **Fatal OOM/stack path does not settle tool deferreds** before `vm.dispose()` (inline).
2. **`promise.settled` early-return can skip deferred dispose** (inline).

### Should fix
3. Timeout path should assert **shared WASM still healthy** (new context after hanging-tool timeout / late host settle) — dispose path already does this.
4. Stale product docs still describe asyncify + global queue (`docs/code-mode/code-mode-isolates.md`, package README, code-mode skill).
5. Early `jobs.error` after `guestSettled = false` can leave a reusable dirty context (inline).
6. `guestSettled` intro comment is inaccurate for busy-loop interrupt (inline).

### Strengths
- Correct asyncify removal; dual timeout (interrupt + host deadline).
- Intentional leak when guest cannot settle after timeout.
- Excellent `Promise.race` abandon + dispose regression (process-level WASM health).
- Changeset migration notes (sync WASM + terminal timeouts) match code-mode per-execution context usage.
- Review follow-ups (dispose race recheck, deferred settle on dispose) look solid.

### Process
- Full TanStack CI has not run on this fork head (only CodeRabbit/Socket) — please ensure workflows are approved/green before merge.
- After the fatal-path cancel lands, unit tests covering “OOM with in-flight tool → new context still works” would lock the invariant.

packages/ai-isolate-quickjs/src/isolate-context.ts
SilvioYMollov: Thank you for pointing this out! Fatal limit cleanup now settles pending tool deferred before disposing the VM. I also added an OOM regression test that verifies a new context still works afterward, using your repro suggestion.

packages/ai-isolate-quickjs/src/isolate-context.ts
SilvioYMollov: Fixed by treating an early pending-job error as terminal and cancelling any outstanding work before safely releasing the context.

packages/ai-isolate-quickjs/tests/isolate-driver.test.ts
SilvioYMollov: Extended the test coverage, it now resolves the host tool after timeout cancellation and verifies that a new context still executes successfully, confirming the late settlement is a no-op and the shared WASM module remains healthy.

tombeckenham:
Thanks for the PR, @SilvioYMollov! 🙌 @tombeckenham will take a look.

- ✅ CI passing
- ✅ No merge conflicts
- ✅ Changeset present
- ⚠️ No E2E test changes detected — behavior changes need coverage under `testing/e2e/` (see CONTRIBUTING)

jherr (approved):

jherr:
@tombeckenham — I went through this branch and mapped each of your still-open review points to where it now lands, so a re-review is easier. The fixes themselves are @SilvioYMollov's; my part was verifying them against the code, bringing the branch up to date with `main` (trivial merge, `ed35b4f0`), and confirming the full suite is green (36 tests pass under Vitest/Node).

Quick status on your five unresolved threads (GitHub has collapsed a few as "outdated" since the code around them was restructured):

### 1. Critical — FreeRuntime hole on fatal limits  *(still open)*
`packages/ai-isolate-quickjs/src/isolate-context.ts` — `releaseVmAfterFatalError()`

Addressed exactly as you suggested (best-effort cancel + grace tick before free). The fatal memory/stack path now settles every in-flight tool deferred **before** `vm.dispose()`:

```ts
// Settle in-flight tool deferreds before freeing the runtime so their native
// handles cannot abort the shared WASM module during JS_FreeRuntime.
for (const cancel of [...this.execState.pendingCancels]) cancel()
this.execState.pendingCancels.clear()
await new Promise((r) => setTimeout(r, 0))   // grace tick
this.vm.dispose()
```

`fail()` routes `MemoryLimitError`/`StackOverflowError` into this path. **Test added** (your repro): *"keeps shared WASM healthy after OOM with an in-flight tool"* — never-resolving tool → OOM → asserts a **new** context from the same driver still executes.

### 2. Important — early return skips `promise.dispose()`  *(outdated/collapsed)*
`packages/ai-isolate-quickjs/src/isolate-driver.ts`

Restructured to your suggestion — the `vm.runtime.alive` check now guards only the job pump, and `promise.dispose()` runs unconditionally in `finally`:

```ts
void promise.settled.then(() => {
  try {
    if (vm.runtime.alive) {
      const jobs = vm.runtime.executePendingJobs()
      if (jobs.error) { /* log */ jobs.error.dispose() }
    }
  } finally {
    promise.dispose()   // no early return; always disposes
  }
})
```

### 3. Important — dirty context reuse after early `jobs.error`  *(outdated/collapsed)*
`packages/ai-isolate-quickjs/src/isolate-context.ts`

An early pending-job pump error is now **terminal**: it cancels outstanding work and releases the context rather than leaving it reusable with a pending program promise:

```ts
const jobs = this.vm.runtime.executePendingJobs()
if (jobs.error) {
  const dumped: unknown = this.vm.dump(jobs.error)
  jobs.error.dispose()
  return await fail(dumped, 'terminal')   // -> releaseAfterUnsettledExecution()
}
```

### 4. Nit — comment contradicts busy-loop timeout path  *(outdated/collapsed)*
The `guestSettled` comment was rewritten to match the actual states:

```ts
// Starts true.
// Becomes false once the guest program promise is being awaited, and returns to true when it settles.
// Stays false if the program promise never settles (e.g. interrupt during a spin).
// Only pre-arm failures leave it true so timeout release can dispose.
let guestSettled = true
```

### 5. Important — assert process-level WASM health after timeout cancel  *(still open)*
`packages/ai-isolate-quickjs/tests/isolate-driver.test.ts`

The timeout path now proves the same invariant as the dispose path. The test hangs a tool → `TimeoutError`, **resolves the tool late** (asserting the second settle is a harmless no-op), then creates a **new** context from the same driver and executes successfully:

```ts
const first = await context.execute(`await never({}); return 1;`)
expect(first.error?.name).toBe('TimeoutError')
// reuse is terminal
expect((await context.execute('return 42')).error?.name).toBe('DisposedError')
await context.dispose()
resolveNever('late')                       // late settle must not poison the module
const next = await driver.createContext({ bindings: {} })
expect((await next.execute('return 42')).value).toBe(42)
```

---

Net: all five are addressed in code + covered by tests, and the branch is current with `main` and mergeable. The only thing still holding it is this review's `changes-requested` state — if the above looks right to you, it just needs a re-review to clear.

> **Attribution:** this analysis and write-up was done by **Claude (Claude Code)**, working in Jack Herrington's (@jherr) environment. Claude read the current code and tests, mapped each review point to its fix, and synced the branch to `main`. Credit for the code fixes themselves goes to @SilvioYMollov; Jack has not personally reviewed this line-by-line.

tombeckenham (approved):

PR #956, opened 2026-07-17
fix(ai-mistral): normalize strict-schema null inputs

jan-kubica:
Mistral strict mode makes optional fields required and nullable. This PR strips only those synthesized `null`s after the model replies. Genuine `.nullable()` values stay.

`chat({ outputSchema })` fills `required` before the adapter runs. This branch now also adds `null` to `enum` and `const` on that path, so the wire schema stays valid JSON Schema.

## 🎯 Changes

- Track the exact nullability added when Mistral strict schemas make optional fields required and nullable.
- Remove only provider-synthesized null fields from streamed tool calls and structured output. Keep genuine `.nullable()` values.
- Widen optional `enum` and `const` constraints so they admit `null`, including after the engine has already filled `required`.
- Recurse into nested objects and arrays whose `type` is a union such as `['object', 'null']`.
- Leave nodes that already accept null as-is. Do not wrap them again.
- Keep `oneOf`, `allOf`, `not`, and `$ref` schemas on the non-strict fallback. `anyOf` stays strict only when no branch-dependent widening is required.
- Do not normalize ambiguous duplicate public tool names.

## ✅ Checklist

## 🚀 Release Impact

## Testing

This branch is merged with current `main`. One conflict was in `packages/ai-mistral/src/utils/schema-converter.ts`. The PR keeps `undoNullWidening` from `@tanstack/ai-utils` and does not keep the old `transformNullsToUndefined` copy.

CI `test:knip` failed on unused `makeMistralStructuredOutputCompatible`. That wrapper is now removed. All call sites use `makeMistralStructuredOutputCompatibleWithMap`.

Commands run after the knip fix (`2921af688`):

1. `pnpm test:knip` — passed
2. `pnpm --filter @tanstack/ai-mistral test:lib` — 46 passed
3. `pnpm --filter @tanstack/ai-mistral test:oxlint` — 0 errors

Not re-run after this commit: `pnpm test:pr` and the E2E matrix.

Manual check:

1. Call `chat({ adapter: mistralText(...), outputSchema: z.object({ mode: z.enum(['canary']).optional(), note: z.string().nullable() }) })`.
2. Confirm the wire schema has `enum: ['canary', null]` for `mode`.
3. Confirm an omitted `mode` comes back missing, and `note: null` stays.

The easy test path is `packages/ai-mistral/tests/mistral-output-schema-chat.test.ts` plus the converter tests in `packages/ai-mistral/src/utils/schema-converter.test.ts`.

## Risk / rollback

Low. The change is inside `@tanstack/ai-mistral` schema conversion. Revert the PR to undo it.


jan-kubica:
@coderabbitai review

packages/ai-mistral/src/utils/schema-converter.ts
```diff
@@ -49,69 +13,142 @@ export function makeMistralStructuredOutputCompatible(
...
+  if (Array.isArray(schema.type) && !schema.type.includes('null')) return false
+
+  if (
+    Array.isArray(schema.anyOf) &&
+    !schema.anyOf.some((variant: Record<string, any>) => acceptsNull(variant))
+  ) {
+    return false
+  }
+
+  return true
```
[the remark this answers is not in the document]
jan-kubica: Addressed with a conservative strict-mode fallback. oneOf, allOf, not, and referenced schemas now remain unchanged and are sent with strict false, avoiding both semantic changes from recursive rewriting and branch-dependent inverse maps. anyOf stays strict only when its branches require no untrackable widening. Added converter, function-tool, and structured-output boundary regressions; the focused Mistral suite passes 30 tests. CC on behalf of @jan-kubica

jan-kubica:
@coderabbitai review

AlemTuzlak:
why is this in draft?

jan-kubica:

packages/ai-mistral/src/utils/schema-converter.ts
```diff
@@ -49,69 +13,188 @@ export function makeMistralStructuredOutputCompatible(
...
+        hasUntrackableAnyOfWidening ||=
+          convertedItems.hasUntrackableAnyOfWidening
+      } else if (isSchemaObject(prop) && Array.isArray(prop.anyOf)) {
+        const converted = coerceMistralStrictSchema(prop, prop.required || [])
+        prop = converted.schema
+        childMap = converted.nullWideningMap
+        hasUntrackableAnyOfWidening ||= converted.hasUntrackableAnyOfWidening
+      }
+
+      if (wasOptional) {
```
AlemTuzlak: **[bug]** Enum and const get `null` only when `wasOptional` is true. `chat({ outputSchema })` first runs `convertSchemaForStructuredOutput`, which puts every key in `required` and rewrites optional types to `['string', 'null']` without adding `null` to `enum`/`const`. The adapter then calls `makeMistralStructuredOutputCompatibleWithMap(outputSchema, outputSchema.required || [])` (`text.ts:254`), so `wasOptional` is false for every field. Mistral still receives `{ type: ['string', 'null'], enum: ['canary'] }`. JSON Schema requires both constraints, so `null` is illegal and an omitted optional enum cannot be expressed. Nested optional objects make this worse: the engine rewrites them to `type: ['object', 'null']`, and line 134 only recurses when `prop.type === 'object'`, so inner enums are not revisited either. Direct `adapter.structuredOutput` with `required: []` still works; that is the only structured-output case the new test hits.

**Suggestion:** After type widening (or whenever `type` already includes `null`), add `null` to `enum`/`const` even if the field is already required. Recurse when `type` is an array that includes `'object'` or `'array'`, not only when it is the string `'object'`/`'array'`. Add a `chat({ outputSchema })` test with an optional enum plus a `.nullable()` field, and assert both the wire schema (`enum` includes `null`) and the value after engine undo (enum key omitted, nullable key kept).

packages/ai-mistral/src/utils/schema-converter.ts
```diff
@@ -49,69 +13,188 @@ export function makeMistralStructuredOutputCompatible(
...
+
+        if (isSchemaObject(prop) && prop.type && !Array.isArray(prop.type)) {
+          prop = { ...prop, type: [prop.type, 'null'] }
+        } else if (
+          isSchemaObject(prop) &&
+          Array.isArray(prop.type) &&
+          !prop.type.includes('null')
+        ) {
+          prop = { ...prop, type: [...prop.type, 'null'] }
+        } else if (!isSchemaObject(prop) || !prop.type) {
```
AlemTuzlak: **[suggestion]** When a field is optional, the converter always rewrites the node even if `acceptsNull(prop)` is already true. For a typical Zod `.nullish()` / optional `.nullable()` shape (`{ anyOf: [{ type: 'string' }, { type: 'null' }] }`, no `type` key), line 186 wraps it again as `{ anyOf: [original, { type: 'null' }] }` and then sets `widenedHere = false`. The extra wrap is not marked synthesized, so inverse undo still keeps `null` (behavior is ok), but the wire schema is nested `anyOf` that the original schema did not need. Required `.nullable()` fields skip this block and stay as a single `anyOf`, which is why the tool-execution test still sees `nullableNote: { anyOf: [{ type: 'string' }, { type: 'null' }] }`.

**Suggestion:** If `originallyAcceptedNull` is true, skip the const/enum/type/`anyOf` rewrite and leave the node as-is. Only widen (and mark `widened`) when the original node did not already admit `null`.

AlemTuzlak (approved):

PR #958, opened 2026-07-19
fix(ai-bedrock): use string-literal specifiers for AWS SDK dynamic imports

### LINKED ISSUE #929: `@tanstack/ai-bedrock`: the AWS SDK cannot be bundled — variable-specifier dynamic imports defeat static bundlers (incl. esbuild)
### TanStack AI version

0.40.0

### Framework/Library version

esbuild

## Summary

`@tanstack/ai-bedrock` loads its AWS SDK dependencies through **dynamic imports whose specifier is a variable, not a string literal**:

`adapters/converse-text.js`
```js
importBedrockRuntime() {
  const mod = "@aws-sdk/client-bedrock-runtime";
  return import(/* @vite-ignore */ mod);
}
```

`utils/auth.js`
```js
const mod = "@aws-sdk/credential-providers";
const { fromNodeProviderChain } = await import(/* @vite-ignore */ mod);
```

Static bundlers cannot resolve `import()` when the argument is a variable — they leave it as a runtime import and **do not include the target in the output**. This is not bun-specific; it is documented behaviour for esbuild too (esbuild only bundles `import()` with a string-literal specifier; a non-analyzable specifier is left external). As a result, **`@aws-sdk/client-bedrock-runtime`, `@aws-sdk/credential-providers`, and their deep/subpath dependencies (`@aws-sdk/token-providers`, `@smithy/core/schema`, …) are excluded from the bundle.** A build that uses this adapter therefore cannot be a self-contained bundle: the AWS SDK must still be resolved from `node_modules` at runtime.

## Contradiction with the documented bundler support

The tree-shaking docs — https://tanstack.com/ai/latest/docs/advanced/tree-shaking#bundle-size-benefits — state:

> Modern bundlers (Vite, Webpack, Rollup, esbuild) can easily eliminate unused code

and list, among the reasons this works:

> No dynamic imports of unused code
> Functions are statically analyzable
> Clear dependency graphs

For the `@tanstack/ai-bedrock` adapter specifically, none of that holds for its AWS SDK edge: the imports are **deliberately** non-analyzable (the source comment says the variable specifier exists "so bundler dep scanners (e.g. Vite/esbuild optimizeDeps) cannot statically discover the AWS SDK"). So the very bundlers the docs advertise — esbuild included — cannot bundle the SDK that this adapter needs to run.

## Reproduction

Bundle a server entry that uses the bedrock adapter with an esbuild-class bundler targeting a self-contained output, e.g.:

```
bun build src/server.ts --target bun --outdir dist
```

Observed:
- The output contains the verbatim `const mod = "@aws-sdk/client-bedrock-runtime"; import(mod)` — i.e. the SDK is **left external**, not bundled.
- Running the output in an environment with **no `node_modules`** fails at runtime, e.g. `Cannot find module '@aws-sdk/token-providers'` (Converse path) or `Cannot find module '@smithy/core/schema'` (auto-install path).

Making the specifiers string literals is the only way we found to get the bundler to include the SDK — confirming the variable specifier is precisely what blocks bundling.

## Impact

Any consumer that ships `@tanstack/ai-bedrock` in a **fully-bundled, `node_modules`-free artifact** (single-file server bundle, some serverless/edge packaging, etc.) cannot do so: the AWS SDK silently drops out of the bundle and only fails at runtime. The adapter effectively requires `node_modules` to be present at runtime, which is at odds with the package's advertised bundler-friendliness.

see above

NA

None

### Terms & Code of Conduct

feiiiiii5:
## 🎯 Changes

`@tanstack/ai-bedrock` loaded its AWS SDK dependencies (`@aws-sdk/client-bedrock-runtime` and `@aws-sdk/credential-providers`) through dynamic imports whose specifier was a **variable**:

\`\`\`ts
const mod = '@aws-sdk/client-bedrock-runtime'
return import(/* @vite-ignore */ mod)
\`\`\`

Static bundlers — esbuild, bun build, Rollup — cannot resolve \`import()\` when the argument is not a string literal, so the SDK was left external and runtime resolution failed in \`node_modules\`-free server bundles. The reported failure in #929: a Bun build of a Bedrock-backed server threw \`Cannot find package @aws-sdk/client-bedrock-runtime\` on the first request.

This PR switches both anti-pattern sites to **string-literal specifiers**:

- \`packages/ai-bedrock/src/utils/auth.ts\` — credential-providers import
- \`packages/ai-bedrock/src/adapters/converse-text.ts\` — client-bedrock-runtime import

The dynamic import still defers the Node-only SDK until first use, so the module-load graph is unchanged — but bundlers can now statically resolve and include the SDK in self-contained server artifacts.

Also adds a source-level regression test (\`packages/ai-bedrock/tests/bundler-static-analysis.test.ts\`) that pins the literal-specifier contract for both sites — if a future refactor brings the variable-specifier pattern back, the test fails before the regression ships. A source-level regex check was chosen over a real bundler run because the contract is purely about static analysability; a regex over the source is the smallest test that faithfully pins it, with no new dev dependency (esbuild/acorn/etc.) and no slow build step in the unit suite.

Vite's dev-time \`optimizeDeps\` pre-bundler may try to scan the now-static SDK import for the browser and fail on the SDK's Node-only exports. Browser-side use of this server-only adapter is unsupported; the comments in \`auth.ts\` and \`converse-text.ts\` document the \`optimizeDeps.exclude\` workaround and point to the matching pattern already in \`examples/ts-react-chat/vite.config.ts\`.

## ✅ Checklist

- [x] I have tested this code locally with \`pnpm run test:pr\`.

## 🚀 Release Impact

### Test Plan

Ran locally (skipping \`test:react-native\` which needs a RN simulator — RN packages are not touched by this PR):

\`\`\`
pnpm exec nx affected --targets=test:sherif,test:knip,test:docs,test:kiira,test:eslint,test:lib,test:types,test:build,build --exclude=examples/**,testing/**
pnpm test:dts
\`\`\`

All green (12/12 Nx tasks, \`test:dts\` clean — 127 .d.ts files, no dangling references). The new \`bundler-static-analysis.test.ts\` passes 4/4. The \`@tanstack/ai-bedrock:test:eslint\` initially flagged a redundant \`as Promise<typeof BedrockRuntime>\` cast that the string-literal specifier made unnecessary; that cast has been removed.

Fixes #929.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved AWS SDK dynamic imports so static bundlers can resolve them reliably by using string-literal import specifiers.
  * Continued deferred loading behavior for AWS services and credential providers.

* **Tests**
  * Added a regression test that scans the codebase to ensure the bundler-compatible import pattern is maintained and to prevent reintroducing the prior indirection approach.

### DIFF
--- a/.changeset/ai-bedrock-static-bundler-specifiers.md
+++ b/.changeset/ai-bedrock-static-bundler-specifiers.md
@@ -1,0 +1,16 @@
+---
+"@tanstack/ai-bedrock": patch
+---
+
+Use string-literal specifiers for the AWS SDK dynamic imports in
+`@tanstack/ai-bedrock` so static bundlers — esbuild, bun build, Rollup — can
+resolve them. Previously the adapter loaded `@aws-sdk/client-bedrock-runtime`
+and `@aws-sdk/credential-providers` through `import(/* @vite-ignore */ mod)`
+where `mod` was a variable; static analysers cannot resolve a non-literal
+specifier, so the SDK was left external and runtime resolution failed in
+`node_modules`-free server bundles. Switching to `import('@aws-sdk/...')`
+keeps the dynamic import (so the Node-only SDK still defers until first use)
+while letting bundlers include it in self-contained server artifacts. Adds a
+source-level regression test pinning the literal-specifier contract.
+
+Fixes #929.
--- a/packages/ai-bedrock/src/adapters/converse-text.ts
+++ b/packages/ai-bedrock/src/adapters/converse-text.ts
@@ -97,8 +97,9 @@ export class BedrockConverseTextAdapter<
    * include the AWS SDK in self-contained server bundles (#929). The SDK is
    * Node/server-only and is only reached on a real request, so the dynamic
    * import also keeps it out of the module-load graph until first use.
-   * `typeof import(...)` is a type-only reference (erased at emit) so the
-   * imported members keep full typing — no cast is needed on the return.
+   * `typeof BedrockRuntime` (from the type-only namespace import above) is
+   * erased at emit, so the imported members keep full typing — no value-level
+   * `as` cast is needed on the return.
    *
    * Vite's dev-time optimizeDeps pre-bundler may try to scan this import for
    * the browser and fail on the SDK's Node-only exports. Browser-side use of
--- a/packages/ai-bedrock/src/adapters/converse-text.ts
+++ b/packages/ai-bedrock/src/adapters/converse-text.ts
@@ -92,17 +92,24 @@ export class BedrockConverseTextAdapter<
   }
 
   /**
-   * Dynamically import `@aws-sdk/client-bedrock-runtime`. The specifier is held
-   * in a variable (not a string literal) so bundler dep scanners (e.g. Vite/
-   * esbuild optimizeDeps) cannot statically discover the AWS SDK and try to
-   * pre-bundle it for the browser — it would fail on the SDK's Node-only
-   * `fromTokenFile` export chain. The SDK is Node/server-only and is only
-   * reached on a real request. `typeof import(...)` is a type-only reference
-   * (erased at emit) so the imported members keep full typing.
+   * Dynamically import `@aws-sdk/client-bedrock-runtime`. The string-literal
+   * specifier lets static bundlers — esbuild, bun build, Rollup — resolve and
+   * include the AWS SDK in self-contained server bundles (#929). The SDK is
+   * Node/server-only and is only reached on a real request, so the dynamic
+   * import also keeps it out of the module-load graph until first use.
+   * `typeof import(...)` is a type-only reference (erased at emit) so the
+   * imported members keep full typing — no cast is needed on the return.
+   *
+   * Vite's dev-time optimizeDeps pre-bundler may try to scan this import for
+   * the browser and fail on the SDK's Node-only exports. Browser-side use of
+   * this server-only adapter is unsupported; if Vite flags the SDK during a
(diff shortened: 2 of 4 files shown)


feiiiiii5:
@coderabbitai thanks for the review. Both nitpicks addressed in `649ab2d`:

**1. `bundler-static-analysis.test.ts` import order** — applied as suggested. Built-in `node:fs` / `node:url` now precede `vitest` (built-in-first, matching the repo's other test files).

**2. `converse-text.ts` return type** — tested inlining `Promise<typeof import('@aws-sdk/client-bedrock-runtime')>` and dropping the top-level `import type * as BedrockRuntime` namespace. It works at runtime but trips the repo's `@typescript-eslint/consistent-type-imports` rule, which bans `import()` type annotations in this codebase (`error  \`import()\` type annotations are forbidden`). Reverted to the namespace import and aligned the JSDoc wording to match (`typeof BedrockRuntime` from the type-only namespace import, not `typeof import(...)`). The key fix for #929 — using a string-literal specifier — is unchanged.

Local pre-PR gate re-run after the refactor: `nx affected --targets=test:lib,test:types,test:eslint,test:build,build` 9/9 ✅, `pnpm test:dts` clean (127 .d.ts files).

feiiiiii5:
Closing this to consolidate my open PRs and refocus my contributions on LLM inference/infrastructure. It hasn't entered maintainer review yet, so I'd rather not leave it lingering. The branch stays on my fork — happy to reopen if a maintainer would like to take it forward. Thanks!

PR #959, opened 2026-07-19
fix(ai-bedrock): pick the bedrock-mantle URL path per model family

### LINKED ISSUE #925: @tanstack/ai-bedrock: `bedrock-mantle` base URL is hardcoded to `/v1`, but the path is per-model (`/openai/v1`, `/anthropic/v1/messages`, `/v1`) — Gemma/Claude 401
### TanStack AI version

0.40.0

### Framework/Library version

Tanstack Start

First and most importantly: I really love this initiative and project. Amazing like anything that gets the Tanstack touch!

Second: I realise patching the lib is probably not what you had in mind and that the @tanstack/ai-bedrock is a very new package still under development. What i hope here is to bring forward some of the findings I've done with using your package as well as some specifics about the bedrock api that I have tried to wrap my head around (but in typical aws fashion, that is probably not possible).

Third, the issue:

## Summary

`buildBaseURL` returns a single path per endpoint:

```ts
endpoint === 'mantle'
  ? `https://bedrock-mantle.${region}.api.aws/v1`
  : `https://bedrock-runtime.${region}.amazonaws.com/openai/v1`
```

On the `bedrock-mantle` endpoint, though, AWS serves **different models on different paths**. The hardcoded `/v1` is correct only for a subset of models. For Gemma the path is `/openai/v1`; for Claude it's `/anthropic/v1/messages`. Sending a request to the wrong path returns a confusing, misleading `401 ... is not enabled for this account (access_denied)` rather than a 404/"wrong path" error, which makes this very hard to diagnose.

## Evidence (AWS model cards → "Programmatic Access")

| Model ID | Endpoint | In-Region endpoint URL (verbatim from the model card) |
|---|---|---|
| `google.gemma-4-31b` | `bedrock-mantle` | `https://bedrock-mantle.{region}.api.aws/openai/v1` |
| `anthropic.claude-haiku-4-5` | `bedrock-mantle` | `https://bedrock-mantle.{region}.api.aws/anthropic/v1/messages` |
| `deepseek.*` | `bedrock-mantle` | `https://bedrock-mantle.{region}.api.aws/v1` |

The Gemma card explicitly calls this out: *"This model is available on the `openai/v1/...` path on the `bedrock-mantle` endpoint. This is different from the `v1/...` path used by other models."*

- Gemma: https://docs.aws.amazon.com/bedrock/latest/userguide/model-card-google-gemma-4-31b.html
- Claude Haiku 4.5: https://docs.aws.amazon.com/bedrock/latest/userguide/model-card-anthropic-claude-haiku-4-5.html
- DeepSeek: https://docs.aws.amazon.com/bedrock/latest/userguide/model-card-deepseek-deepseek-v3-2.html

## Reproduction

```ts
const adapter = bedrockText('google.gemma-4-31b', {
  api: 'chat',
  endpoint: 'mantle',
  region: 'eu-central-1',
})
// → POSTs to https://bedrock-mantle.eu-central-1.api.aws/v1/chat/completions
// → 401 "... is not enabled for this account (access_denied)"
```

The SigV4 signing service (`bedrock-mantle`) is correct — only the path is wrong. Overriding `baseURL` to `https://bedrock-mantle.eu-central-1.api.aws/openai/v1` makes the exact same call succeed.

## Broader design concerns surfaced while adding our own models

We're pinning + patching `model-catalog.generated.ts` to add models that ship faster than the catalog updates. Doing that exposed a few gaps beyond the path bug:

1. **The catalog encodes `apis` (`converse`/`chat`/`responses`) but not the endpoint or the URL path.** There's nothing to derive the correct base URL from, so `buildBaseURL` can only be a per-endpoint constant — which is wrong for any model whose mantle path isn't `/v1`. The catalog seems like the right place to carry, per model, *which endpoint(s), which API(s), and which path*.

2. **`api: 'chat'` can't represent Claude-on-mantle.** Claude's mantle path is `/anthropic/v1/messages` — the Anthropic **Messages** API, not OpenAI Chat Completions. So "chat over mantle" differs by model in both path *and* wire format, and there's no `api` value that captures the Messages shape.

3. **`region` is decoupled from the model id, but AWS couples them.** The catalog lists geo-prefixed ids (e.g. `us.anthropic.claude-...`) as first-class models, yet `bedrockText(model, { region })` accepts any region. `bedrockText('us.anthropic.claude-haiku-4-5-20251001-v1:0', { region: 'eu-central-1' })` typechecks but is contradictory (a US geo-inference profile invoked from an EU region). Per-model region/geo availability (`us.` / `eu.` / `global.` / in-region) isn't captured, so invalid combinations only fail as a remote error.

### 1. Make the generated catalog carry endpoint / path / region

Have the generated catalog compile, per model, everything needed to build a correct request without the caller guessing:

- supported `endpoint`(s): `runtime` / `mantle`
- supported `api`(s) per endpoint (ideally including wire format, since mantle-Claude is Anthropic Messages, not OpenAI chat)
- the URL **path** per (endpoint, api): `/v1`, `/openai/v1`, `/anthropic/v1/messages`
- region / geo availability

Then `buildBaseURL` and adapter selection can be catalog-driven rather than per-endpoint constants, and invalid `(model, endpoint, api, region)` combinations can fail at the type level or with a clear local error instead of a remote `401 access_denied`.

### 2. A first-class way to declare a custom model (so users don't patch the lib)

Today the only way to use a model the catalog doesn't ship yet is to edit `model-catalog.generated.ts` — which for a pinned dependency means patching the package (`bun patch` / `patch-package`). That's brittle: the patch is version-pinned and silently stops applying on upgrade, and it can't be caught by typecheck. The catalog is both the **type source** (the model-id union via `IdsWhere<'chat'>`) and the **runtime behavior source** (apis/modalities, and — per this issue — ideally the path), so a custom model has to supply both, and there's currently no public seam to do it. New Bedrock models ship faster than a catalog release (this is how we hit Gemma and Kimi before they were in the generated catalog), so an escape hatch matters.

Options that would remove the need to patch:

**(a) Accept an inline model descriptor at the call site.** Let `bedrockText` take either a catalog key (as today, with full inference) or a fully-specified object for anything not in the catalog:

```ts
bedrockText(
  {
    id: 'google.gemma-4-31b',
    endpoint: 'mantle',
    api: 'chat',
    path: '/openai/v1',          // per-model, from the model card
    input: ['text'],
    output: ['text'],
  },
  { region: 'eu-central-1' },
)
```

When a descriptor is passed, skip the catalog lookup and use its fields directly. This also self-serves the path bug above — users set the correct path without waiting for a catalog release.

**(b) A `defineBedrockModel({...})` helper** that returns a typed, reusable model object to pass to `bedrockText` (nice for sharing a custom model across a codebase).

**(c) A registry + module augmentation:** `registerBedrockModels([...])` that merges entries into the runtime catalog, plus an exported interface consumers can `declare module`-augment to extend the id union — so custom ids typecheck without editing generated files.

Any of these lets teams adopt just-shipped models without a lockfile-pinned patch that breaks on the next upgrade, while keeping the fully-inferred experience for catalog models.

## Workaround (today)

```ts
bedrockText('google.gemma-4-31b', {
  api: 'chat',
  endpoint: 'mantle',
  region: 'eu-central-1',
  baseURL: `https://bedrock-mantle.${region}.api.aws/openai/v1`,
})
```

see above

not applicable

None

### Terms & Code of Conduct

feiiiiii5:
## 🎯 Changes

`@tanstack/ai-bedrock` hardcoded the `bedrock-mantle` base URL to `/v1`, but AWS serves different model families on different mantle paths. From the AWS model cards' "Programmatic Access → In-Region endpoint URL":

| Model ID | Mantle path | Source |
|---|---|---|
| `google.gemma-*` | `/openai/v1` | [Gemma card](https://docs.aws.amazon.com/bedrock/latest/userguide/model-card-google-gemma-4-31b.html) — explicitly: *"This model is available on the `openai/v1/...` path on the `bedrock-mantle` endpoint. This is different from the `v1/...` path used by other models."* |
| `deepseek.*` | `/v1` | [DeepSeek card](https://docs.aws.amazon.com/bedrock/latest/userguide/model-card-deepseek-deepseek-v3-2.html) |
| `openai.gpt-oss-*` | `/v1` | (default) |

The hardcoded `/v1` was correct only for the latter two. Sending a Gemma request to `/v1/chat/completions` returned a misleading `401 ... is not enabled for this account (access_denied)` instead of a 404 / "wrong path" error, which made the bug hard to diagnose (#925). The SigV4 signing service (`bedrock-mantle`) was correct — only the path was wrong.

This PR threads the model id through `withBedrockDefaults` → `buildBaseURL` → `mantlePathForModel`, which routes `google.gemma-*` to `/openai/v1` and falls back to `/v1` for every other id. The chat and responses adapter constructors pass the model id through; an explicit `baseURL` override still wins (preserves the existing escape hatch and the E2E → aimock wiring).

The runtime (`bedrock-runtime`) endpoint is unchanged — every chat-capable model is served at `/openai/v1` there, so the model parameter is unused on that branch (covered by a separate test group).

### Out of scope (called out as "Suggested directions" in #925)

- **Catalog-driven path lookup (Direction 1)** — carry the path per `(endpoint, api)` on the generated catalog entry so this becomes data-driven instead of a prefix switch. That's a larger follow-up; the prefix match here is the minimal bug fix.
- **Claude-on-mantle at `/anthropic/v1/messages`** — that path serves the Anthropic Messages API, which has a different wire format from OpenAI Chat Completions, so the chat adapter can't drive it regardless of the path. The catalog already marks Claude models as `chat: false`, so `bedrockText('anthropic.claude-…', { endpoint: 'mantle' })` typechecks as an error today. A native Messages-API adapter would be a separate piece of work.

## ✅ Checklist

- [x] I have tested this code locally with \`pnpm run test:pr\`.

## 🚀 Release Impact

### Test Plan

Ran locally (skipping `test:react-native` which needs a RN simulator — RN packages are not touched by this PR):

\`\`\`
pnpm exec nx affected --targets=test:sherif,test:knip,test:docs,test:kiira,test:eslint,test:lib,test:types,test:build,build --exclude=examples/**,testing/**
pnpm test:dts
\`\`\`

All green (9/9 Nx tasks, `test:dts` clean — 127 .d.ts files). The new `tests/mantle-path.test.ts` adds 14 cases pinning:
- Gemma (3 variants incl. versioned and hypothetical future) → `/openai/v1` on mantle
- DeepSeek / gpt-oss / unknown id → `/v1` default on mantle
- Backward compat: model omitted / undefined / empty string → `/v1` default
- `acme.google.gemma-fake-v1:0` does **not** match (uses `startsWith`, not `includes` — Bedrock model ids always start with the provider prefix)
- Explicit `baseURL` override wins even when model would route elsewhere
- Runtime endpoint: model has no effect on path (3 cases, pins the runtime-branch invariant)

The signature change to `withBedrockDefaults(config, forced?, model?)` is backward compatible — `model` is optional and defaults to undefined → `/v1` (the pre-#925 behavior). Existing `tests/client.test.ts` calls without `model` still pass unchanged.

Fixes #925.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Amazon Bedrock model routing for Gemma models by using the correct `/openai/v1` path.
  * Preserved the default `/v1` path for other model families.
  * Continued honoring explicit base URL overrides.
  * Kept runtime endpoint behavior unchanged.

### DIFF
--- a/.changeset/ai-bedrock-mantle-per-model-path.md
+++ b/.changeset/ai-bedrock-mantle-per-model-path.md
@@ -1,0 +1,33 @@
+---
+"@tanstack/ai-bedrock": patch
+---
+
+Pick the `bedrock-mantle` URL path per model family. The mantle endpoint was
+hardcoded to `/v1`, but AWS serves different model families on different
+paths: Gemma needs `/openai/v1` (per its model card's "In-Region endpoint
+URL"), while gpt-oss and DeepSeek use the `/v1` default. The wrong path
+returned a misleading `401 ... is not enabled for this account (access_denied)`
+instead of a 404 / "wrong path" error, which made the bug hard to diagnose.
+
+`withBedrockDefaults` now takes the model id as a third argument and threads
+it into `buildBaseURL` → `mantlePathForModel`, which routes `google.gemma-*`
+to `/openai/v1` and falls back to `/v1` for every other id. The chat and
+responses adapter constructors pass the model id through. An explicit
+`baseURL` override still wins (preserves the existing escape hatch and the
+E2E → aimock wiring).
+
+The runtime (`bedrock-runtime`) endpoint is unchanged — every chat-capable
+model is served at `/openai/v1` there, so the model parameter is unused on
+that branch.
+
+Out of scope: a catalog-driven refactor (Direction 1 in #925) that carries
+the path per `(endpoint, api)` on the generated catalog entry would replace
+this prefix switch with a data-driven lookup. That's a larger follow-up;
+this is the minimal bug fix. Claude-on-mantle at `/anthropic/v1/messages`
+is also out of scope — that path serves the Anthropic Messages API, which
+has a different wire format from OpenAI Chat Completions, so the chat
+adapter can't drive it regardless of the path. The catalog already marks
+Claude models as `chat: false`, so that combination typechecks as an error
+today.
+
+Fixes #925.
--- a/packages/ai-bedrock/src/adapters/responses-text.ts
+++ b/packages/ai-bedrock/src/adapters/responses-text.ts
@@ -52,11 +52,13 @@ export class BedrockResponsesTextAdapter<
 
   constructor(config: BedrockResponsesConfig, model: TModel) {
     // Responses is mantle-only — force the mantle base URL (an explicit
-    // config.baseURL still wins, e.g. E2E pointing at aimock).
+    // config.baseURL still wins, e.g. E2E pointing at aimock). `model` reaches
+    // `buildBaseURL` so the mantle path matches the model family (Gemma needs
+    // `/openai/v1` instead of the `/v1` default) — see #925.
     super(
       model,
       'bedrock-responses',
-      new OpenAI(withBedrockDefaults(config, 'mantle')),
+      new OpenAI(withBedrockDefaults(config, 'mantle', model)),
     )
   }
 }
--- a/packages/ai-bedrock/src/adapters/text.ts
+++ b/packages/ai-bedrock/src/adapters/text.ts
@@ -52,7 +52,9 @@ export class BedrockTextAdapter<
 
   constructor(config: BedrockTextConfig, model: TModel) {
     // No `forced` -> honors config.endpoint ('runtime' default, 'mantle' allowed).
-    super(model, 'bedrock', new OpenAI(withBedrockDefaults(config)))
+    // `model` reaches `buildBaseURL` so the mantle path matches the model family
+    // (e.g. Gemma needs `/openai/v1` instead of the `/v1` default) — see #925.
+    super(model, 'bedrock', new OpenAI(withBedrockDefaults(config, undefined, model)))
   }
 
   /**
--- a/packages/ai-bedrock/src/utils/client.ts
+++ b/packages/ai-bedrock/src/utils/client.ts
@@ -27,16 +27,58 @@ const DEFAULT_REGION = 'us-east-1'
 /** OpenAI SDK requires a non-empty apiKey even when a signed fetch overrides Authorization. */
 const SIGV4_PLACEHOLDER_KEY = 'bedrock-sigv4'
 
-function buildBaseURL(region: string, endpoint: BedrockEndpoint): string {
+/**
+ * Per-model URL path on the `bedrock-mantle` endpoint (#925).
+ *
+ * AWS serves different model families on different mantle paths:
+ *   - `google.gemma-*`     → `/openai/v1`   (OpenAI-compatible translation layer)
+ *   - `deepseek.*`         → `/v1`          (default OpenAI-compatible path)
+ *   - `openai.gpt-oss-*`   → `/v1`          (default)
+ *   - `anthropic.claude-*` → `/anthropic/v1/messages` — *NOT* supported by the
+ *     chat adapter: that path serves the Anthropic Messages API, which has a
+ *     different wire format from OpenAI Chat Completions. Calling
+ *     `bedrockText('anthropic.claude-…', { endpoint: 'mantle' })` will not work
+ *     regardless of the path; use the Converse adapter or wait for a
+ *     Messages-API adapter. The catalog already marks Claude models as
+ *     `chat: false`, so this combination typechecks as an error today.
+ *
+ * Evidence: AWS model cards → "Programmatic Access" → "In-Region endpoint URL".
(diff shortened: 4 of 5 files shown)


feiiiiii5:
Same reasoning as my other TanStack PRs — I'm trimming my open-PR queue and refocusing elsewhere. The work is preserved on my fork branch, so this can be reopened anytime if it's useful. Thanks for the CodeRabbit pass.

PR #970, opened 2026-07-20
feat(interrupts): AG-UI interrupt lifecycle overhaul (ephemeral)

AlemTuzlak:
Extracts the AG-UI interrupts overhaul (from #935, which was stacked on the persistence PR #785) into its own PR targeting `main`, **excluding the persistence story**. Interrupts run fully ephemeral here: resume reconstructs pending tool calls from the client's message history in a fresh child run. A follow-up stacked PR will add durable persistence back on top.

## What's in scope
- AG-UI interrupt lifecycle: `RUN_FINISHED` with `outcome.type === 'interrupt'`, interrupt descriptors/bindings, ephemeral resume via `parentRunId` + `resume` batch.
- Tool-approval and client-tool interrupt kinds.
- Client-side `interruptManager` (hydrate / resolve / batch submit), `useChat` integration across frameworks.
- `ts-react-chat` interrupt-lab example for manual testing.

## Explicitly out of scope (deferred to the stacked persistence PR)
- `@tanstack/ai-persistence`, `-drizzle`, `-prisma`, `-cloudflare`.
- Durable interrupts (recovery via `joinRun`, server-side interrupt state fetchers).

## Notable fixes
- Forward `parentRunId` so ephemeral resume correlates the interrupted run to its continuation child run.
- Resolve the approval tool-call's owning message from history when a `MESSAGES_SNAPSHOT` reset the live maps, so the approval UI state still updates.
- Removed durable (`interrupts-v2`) e2e routes/fixtures/specs and persistence-era recovery tests; these return with persistence.

## Testing
- `tool-approval.spec.ts` e2e green across all providers (approve continuation, denial, issue #532 follow-up).
- Unit tests, typecheck, lint, build green for affected packages.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Added AG-UI interrupt lifecycle with support for tool approvals, generic responses, and client-tool execution, including typed payloads/edited args, cancellation, retry, and atomic multi-interrupt resolution.
  * Introduced resumable interrupt state via resume/continuation handling, plus persistence with durable interrupt recovery (ephemeral enabled by default) and new storage-adapter options.
  * Expanded interrupt/resume state and controls across React, Preact, Vue, Solid, Svelte, and Angular integrations.
  * Added an “Interrupts Lab” example and new `/interrupts` UI flow.

* **Documentation**
[...]

PR #973, opened 2026-07-21
chore: non-interrupt residue archived from #970 slim

tombeckenham:
## Summary

Archive of the **pre-slim** tip of #970 (`feat/interrupts-overhaul` before Option A cleanup).

#970 was force-pushed to a slim surface (~132 files) that keeps the AG-UI interrupt lifecycle. This branch freezes the full pre-slim tree (**267 files**) so nothing was lost:

- Generation resume / identity-replay paths
- Framework `useGeneration*` / media generation wiring churn
- Example + e2e media/devtools touch-ups
- Import-only / autofix noise
- Other persistence-stack residue that rode along with the interrupt extract

**Not intended to merge as-is.** Use it to cherry-pick or re-land non-interrupt work later (e.g. client generation resume, storage helpers that are not required for ephemeral interrupts).

Related: https://github.com/TanStack/ai/pull/970

## Test plan

- [ ] Do not merge until intentionally rebased/split
- [ ] When reclaiming work: diff against `feat/interrupts-overhaul` and extract only the desired paths


tombeckenham:
Closing — this was only ever an archive of the pre-slim #970 tree, never intended to merge, and it's just noise in the PR queue at this point.

The branch `chore/non-interrupt-residue-from-970` stays put, so nothing is lost. If we want the non-interrupt residue later (generation resume / identity replay, `useGeneration*` wiring), it's all still there to cherry-pick from.

PR #542, opened 2026-05-10
feat(ai-orchestration): generator-based workflows + orchestrators

AlemTuzlak:
## Summary

- New `@tanstack/ai-orchestration` package: define agents, workflows, and orchestrators using async generators. `yield* agents.x(...)` for typed agent calls, `yield* approve(...)` for pause/resume on user decision, plain JS for everything else (`if`, `for`, `await`, `Promise.all`).
- `defineOrchestrator` is sugar over `defineWorkflow` (same runtime, different vocabulary). `defineRouter(config, fn)` lets users extract the orchestrator router as a named function with full type inference.
- All workflow lifecycle is emitted as native AG-UI events (`RunStarted`, `StepStarted/Finished`, `StateSnapshot`, `StateDelta` via JSON Patch RFC 6902, `RunFinished` carrying typed output, `RunError`). Approvals reuse the existing `approval-requested` custom event.
- `WorkflowClient` added to `@tanstack/ai-client` (mirrors `ChatClient`'s connection-adapter pattern), `useWorkflow` / `useOrchestration` hooks added to `@tanstack/ai-react`.
- Server-side `parseWorkflowRequest(request)` extractor; consumers use `runWorkflow({ workflow, runStore, ...params })` + `toServerSentEventsResponse(stream)` — symmetric with how `chat()` is called.
- Pluggable `RunStore` interface with default `inMemoryRunStore` (1h TTL).
- Two demo pages on `ts-react-chat`: `/workflow` (article writing pipeline with writer → legal → skeptic → editor → approve/revise loop, dramatic fullscreen ArticleModal on publish) and `/orchestration` (Claude Code-style spec → approve → implement [nested workflow] → review).
- Editorial-brutalist UI: Fraunces variable serif + JetBrains Mono, warm ink/cream palette with citron accent, paper-grain textures, hazard-tape approval bands, live `DraftPreview` rendering the article-in-progress as the editor mutates state.

## Architecture decisions captured during design

- Workflow body is an async generator (`function*`), not a node-array DSL — preserves "feels like JavaScript".
- State lives in a declared schema, mutated as a plain object inside the generator; engine snapshots between yields and emits JSON Patches.
- `succeed({...})` / `fail(reason)` helpers replace `as const` discriminator casts at return sites.
- v1 `runStore` is in-memory only; engine uses live generator handles (no replay) for resume. Pluggable interface in place for future durable stores.

## Out of scope (post-prototype)

[...]

PR #979, opened 2026-07-21
Add Workflow-backed AI orchestration

tannerlinsley:
## RFC

**[Read the Workflow-backed AI orchestration RFC](https://github.com/TanStack/ai/blob/taren/workflow-orchestration-rfc/rfcs/0001-workflow-backed-orchestration.md)**

This PR is the RFC plus a small experimental implementation that proves the proposed boundary: TanStack Workflow owns durable execution, while TanStack AI owns agent ergonomics, AI persistence, and AG-UI projection.

**Authoring model decision:** async handlers and the context API are the sole workflow syntax. A generator/`yield*` DSL is rejected and is not planned as optional sugar.

## Decisions requested

- AI persistence ownership and public contract
- interrupt and approval envelope
- AI run ID correlation
- token attempt/reset and publisher failure semantics
- durable boundaries inside long model/tool loops
- child workflow semantics

## Implementation included

- experimental `@tanstack/ai-orchestration` package
- `defineAgent()` and Workflow middleware exposing `ctx.ai.agent()`
- Workflow-to-AG-UI event projection and publisher adapter
- replay, approval resume, deadline yield, publishing, type, and provider-free E2E coverage

This supersedes #542: no second workflow engine, store contract, generator DSL, or parallel authoring model.

Depends on the merged TanStack/workflow#15 for runtime event publishing and stale-lease recovery.

## Validation

- `pnpm test:pr`
- `pnpm --filter @tanstack/ai-e2e test:e2e`
- RFC formatting and repository link validation
- all GitHub checks passing before this RFC-only update


jherr (approved):

rfcs/0001-workflow-backed-orchestration.md
```diff
@@ -0,0 +1,266 @@
...
+    return { draft, approved: review.approved }
+  })
+```
+
+`ctx.ai` is the package-owned middleware namespace. Workflow core does not need
+to reserve every official integration name. A conflicting `ai` extension is a
+composition error visible in TypeScript, while Workflow's own fields remain
+reserved by core.
+
+The first argument to `ctx.ai.agent()` is the stable Workflow step ID. It is not
```
tombeckenham, line 112: Why does a workflow step id need to be explicit?

rfcs/0001-workflow-backed-orchestration.md
```diff
@@ -0,0 +1,266 @@
...
+an arbitrary batch position. Reordering code across deployed workflow versions
+still follows Workflow's normal versioning rules.
+
+## Agent execution semantics
+
+An agent call is one Workflow step.
```
tombeckenham, line 118: Is this one turn? Or all the turns

tombeckenham: The reason I ask is that an agent loop might have 5 interations. If it gets through 3, then fails, if the chat is inside the durable step, then those 3 steps are repeated. There's no durability inside the agent loop. e.g. an agent could call a tool in one loop which adds a todo item to a list. The user would have to make all tool calls indempotent to ensure the tool wasn't called a second time. If we could resume the loop, that could be avoided. That'd mean making the agent loop itself durable. You've said below that "Long autonomous loops need a later integration that exposes model turns and side-effecting tools as Workflow operations". Any idea how that might work?

rfcs/0001-workflow-backed-orchestration.md
```diff
@@ -0,0 +1,266 @@
...
+produce duplicate attempts.
+
+## Child workflows
+
+True child workflows are deferred. Wrapping a child run inside `ctx.step()` is
+not sufficient: cancellation, parent-child status, signals, leases, and replay
+all need first-class runtime semantics.
+
+Composition in the first release uses normal helper functions and agent calls
+inside one Workflow definition.
```
tombeckenham, line 223: Cloudflare's event model is a nice way to do child worfkows. If you have a way to send events, and wait for events, then it's easy to create a child model.

sedghi:
Hi guys, here is the C for the RF:
This is great and really happy you kept the plain async handler and didn't go down the mastra style graph builder route (.branch,.map,.commit stuff). One thing that isn't clear to me though: what does `agentMiddleware()` actually buy us? A workflow shouldn't need a middleware registered just to call an agent, if I'm not mistaken it exists only so `ctx.ai` typechecks, and the method itself just delegates to `ctx.step()`. I'd rather the agent be callable directly with the ctx, something like:


```ts
const draft = await writer.step(ctx, { topic: ctx.input.topic })
```

instead of `.middleware([agentMiddleware()])` + `ctx.ai.agent('draft', writer, ...)`, which makes it no registration and one less name at the call site (the step ID could default to the agent name and only be passed explicitly when the same agent runs twice), and it matches how `workflow-core` already handles shared helpers. Is there a reason the middleware/namespace approach was chosen that I'm missing?

So overall will look like 


```ts
import { defineAgent } from '@tanstack/ai-orchestration'
import { openaiText } from '@tanstack/ai-openai'
import { createWorkflow } from '@tanstack/workflow-core'
import { z } from 'zod'

const writer = defineAgent({
  name: 'writer',
  input: z.object({ topic: z.string() }),
  output: z.object({ article: z.string() }),
  run: ({ input, abortController }) =>
    chat({
      adapter: openaiText('gpt-5.5'),
      messages: [{ role: 'user', content: `Write about ${input.topic}` }],
      outputSchema: z.object({ article: z.string() }),
      stream: true,
      abortController,
    }),
})

const editor = defineAgent({
  name: 'editor',
  input: z.object({ article: z.string(), feedback: z.string() }),
  output: z.object({ article: z.string() }),
  run: ({ input, abortController }) =>
    chat({
      adapter: openaiText('gpt-5.5'),
      messages: [
        {
          role: 'user',
          content: `Revise this article based on the feedback.\n\nArticle: ${input.article}\n\nFeedback: ${input.feedback}`,
        },
      ],
      outputSchema: z.object({ article: z.string() }),
      stream: true,
      abortController,
    }),
})

export const article = createWorkflow({
  id: 'article',
  input: z.object({ topic: z.string() }),
}).handler(async (ctx) => {
  const draft = await writer.step(ctx, { topic: ctx.input.topic })

  const review = await ctx.approve({ title: 'Publish this article?' })

  if (review.approved) {
    return { article: draft.article, published: true }
  }

  if (!review.feedback) {
    return { article: draft.article, published: false }
  }

  const revised = await editor.step(ctx, {
    article: draft.article,
    feedback: review.feedback,
  })

  const second = await ctx.approve({ title: 'Publish the revised article?' })
  return { article: revised.article, published: second.approved }
})
```

PR #984, opened 2026-07-22
feat(persistence): server persistence + client browser-refresh durability

AlemTuzlak:
## 🎯 Changes

Adds durable chat state on the server and full-page-reload durability on the client, on top of the shipped ephemeral interrupts (#970), resumable streams (#955), and the shared `Scope` type (#980). Uses #785 as reference; #987 (generation persistence) and #988 (sandbox persistence) stack on this branch.

### Server: one new package, `@tanstack/ai-persistence`

It ships the **contract, not backends** — four store interfaces (`MessageStore`, `RunStore`, `InterruptStore`, `MetadataStore`), the `withPersistence` / `withGenerationPersistence` middleware, `reconstructChat` for server-side hydrate, an in-memory reference backend, and a conformance testkit.

An earlier revision of this branch also published `-drizzle`, `-prisma`, and `-cloudflare` packages. Those were removed in 62c9975: a backend package has to own DDL, which puts it in permanent conflict with the migration tooling the consuming app already runs. You implement the stores against your own database instead — the guide and the agent skills walk through it per stack, and the conformance suite is what proves an adapter correct.

### Locks live in `@tanstack/ai`, not in the persistence package

Multi-instance coordination started out inside `@tanstack/ai-persistence` and moved to core in cd8b731. `LockStore` / `InMemoryLockStore` / `LocksCapability` / `withLocks` are exported from `@tanstack/ai`.

The split is deliberate: state stores answer *"what is durable chat data?"*, locks answer *"who may run this critical section right now?"* Locks are **not** a `stores` key, are **not** composable via `composePersistence`, and are not covered by the conformance testkit. `withPersistence` does not lock a turn for you. Keeping them in core also means the consumer that actually needs them today — `@tanstack/ai-sandbox`, which reads `LocksCapability` off the middleware chain — doesn't drag in a persistence dependency to get mutual exclusion.

`InMemoryLockStore` is a per-key promise chain, correct within a single process only; multi-instance deployments implement the one-method `LockStore` interface themselves, and the Cloudflare Durable Object recipe walks through a lease-based one. New guide at `docs/advanced/locks.md`.

### Client: browser-refresh durability

The `persistence` option now stores one combined `{ messages, resume? }` record per chat id, so a full page reload restores the transcript, rehydrates pending interrupts, and rejoins an in-flight run via `joinRun` on a durability-backed connection. Adds `localStoragePersistence` / `sessionStoragePersistence` / `indexedDBPersistence`, and an object form `{ store, messages: false }` that keeps large transcripts off the client while still persisting the resume pointer.

Because it rides the existing option, **all six framework integrations get it with no framework-specific code**.

### Breaking

- **The chat hooks no longer accept `id`** — a hook's identity is its `threadId` (`usechat-threadid-identity`). `ChatClient` keeps `id` as a lower-level escape hatch.
- **Hand-rolled `persistence` adapters must update their write path.** `setItem` now receives the combined `{ messages, resume? }` record instead of a bare `UIMessage[]`. `getItem` still reads legacy arrays, but an adapter that assumed an array will write the new shape and fail to parse it back — and because adapter reads are best-effort, the conversation silently does not restore. This bit the e2e harness in this very PR; `docs/persistence/client-persistence.md` now documents the round-trip.

### Docs, skills, example, e2e

- New `docs/persistence/` section: overview, chat persistence, client persistence, controls, build-your-own-adapter, migrations, internals. Replaces `docs/chat/persistence.md`. Locks get their own page under advanced, `docs/advanced/locks.md`.
- `@tanstack/ai-persistence` ships 7 agent skills, nested `ai-core`-style under `skills/ai-persistence/` — an entry point, `server`, `stores`, and four `build-*-adapter` recipes that read the app's existing ORM config and schema and write a single `chat-persistence.ts` against it (Drizzle, Prisma, Cloudflare D1, and a custom recipe for raw `pg` / Kysely / Mongo / Supabase).
- Two skills ship in `@tanstack/ai` instead, because that's where the code they teach lives: `ai-core/client-persistence` (browser persistence is in the framework packages — an app persisting only in the browser never installs the server package) and `ai-core/locks`. Also fixes three packages (`ai-mcp`, `ai-memory`, `ai-sandbox`) whose skills TanStack Intent could never load.
- `examples/ts-react-chat` `/persistent-chat` route (SQLite server + localStorage client), with `sqlite-persistence.ts` as the runnable reference the docs point at.
- E2E: `persistence-durability.spec.ts` (message restore after reload, interrupt survives reload) plus localStorage persistence cases in `chat.spec.ts`.

## ✅ Checklist

`test:pr` green on the current head. Full Playwright suite run locally: 385 passed, 1 skipped. Three tools-test specs failed only under unbounded local parallelism and pass at CI's `--workers=4`; Nx separately flagged `@tanstack/ai-sandbox-local-process:test:lib` as a flaky task and it passes on re-run.

## 🚀 Release Impact

Ten changesets on this branch: `persistence-packages` and `fresh-client-tail` (minor, `@tanstack/ai-persistence`), `client-browser-refresh-durability` and `usechat-threadid-identity` (minor, `ai-client` + all six frameworks), `seamless-reload-resume` and `fast-fail-rejoin` (`@tanstack/ai`), `locks-to-core` (minor, `ai` / `ai-persistence` / `ai-sandbox`), `memorystream-agent-loop-delivery` and `rejoin-not-aborted-on-mount` (patch), and `skill-discoverability` (patch, `ai` / `ai-mcp` / `ai-memory` / `ai-sandbox`). No majors — the two breaking changes above land in pre-1.0 minors.


docs/persistence/custom-stores.md
```diff
@@ -0,0 +1,161 @@
...
+  const base = cloudflarePersistence({ d1: state })
+  return composePersistence(base, {
+    overrides: { interrupts, runs },
+  })
+}
+```
+
+Only those two stores move to the custom database; D1 still owns messages and
+metadata. Composition does not create a transaction across those systems;
+design related writes accordingly.
```
jherr: It would be good to have specifics for each of the events that trigger persistence and what the flow is through the persistence store at that point. (e.g. this is when/how we load the thread, this is when/how we store messages, this is what happens when there is an interrupt, etc.) Maybe as swim lanes.

Basically I don't think folks are going to have a different store for chats. They are going to want to integrate it in with the rest of the customer data. So we should invest pretty heavily in this doc.

docs/persistence/drizzle.md
```diff
@@ -0,0 +1,164 @@
+---
+title: Persistence with Drizzle
+id: drizzle
+---
+
+# Persistence with Drizzle
+
+`@tanstack/ai-persistence-drizzle` supports SQLite-family Drizzle databases.
```
tombeckenham: Drizzle should support every type of database shouldn't it?

packages/ai-persistence/src/types.ts
```diff
@@ -0,0 +1,422 @@
...
+   */
+  delete: (namespace: string, key: string) => Promise<void>
+}
+
+export interface AIPersistenceStores {
+  messages?: MessageStore
+  runs?: RunStore
+  interrupts?: InterruptStore
+  metadata?: MetadataStore
+  locks?: LockStore
```
tombeckenham: This feels off. Locks is not really chat state. It shoudn't be on this type

tombeckenham: It also feels off that all other fields are optional. Looking further

tombeckenham (approved):
I'm happy with this now. I've reworked drizzle a bit. I've added skills and I'm happy with the type structure.

packages/ai-persistence-drizzle/src/sqlite.ts
```diff
@@ -0,0 +1,82 @@
...
+export function sqlitePersistence(options: SqlitePersistenceOptions) {
+  const filename = normalizeSqliteUrl(options.url)
+  ensureParentDirectory(filename)
+  const sqlite = new DatabaseSync(filename)
+  try {
+    if (options.migrate) applySqliteMigrations(sqlite, sqliteMigrations)
+  } catch (error) {
+    sqlite.close()
+    throw error
+  }
```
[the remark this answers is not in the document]
tombeckenham: Declining an `engines.node` floor on the package root.

`@tanstack/ai-persistence-drizzle` has two surfaces:

1. **Root `drizzlePersistence(db, { provider, schema })`** — BYO Drizzle DB. Works with SQLite *or* Postgres drivers, and is imported from edge/Cloudflare (D1) paths that are not Node. An `engines.node: ">=22.13"` on the package would incorrectly advertise that the whole package needs modern Node.
2. **`/sqlite` convenience factory** — the only entry that imports `node:sqlite`. That path is already documented as Node-only (`docs/persistence/drizzle.md`), and failure mode on older Node is a clear missing-module / experimental-API error at import time.

If we document a floor, it belongs on the `/sqlite` docs/README callout (Node ≥22.13 / ≥23.4 for unflagged `node:sqlite`), not as a package-wide `engines` constraint that rejects legitimate Postgres/edge consumers. Happy to add that callout if you want it explicit in the package README.

packages/ai-persistence-drizzle/tests/migration-cli.test.ts
```diff
@@ -0,0 +1,65 @@
+import { mkdtemp, readFile, rm, writeFile } from 'node:fs/promises'
+import { tmpdir } from 'node:os'
+import { join } from 'node:path'
+import { afterEach, describe, expect, it } from 'vitest'
+import { sqliteMigrations } from '../src/migrations'
+import { runDrizzleMigrationsCli } from '../src/migration-cli'
```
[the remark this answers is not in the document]
tombeckenham: Leaving the `tests/` layout as-is.

Across this monorepo the established convention for library packages is `include: ['tests/**/*.test.ts']` (see `@tanstack/ai-client`, `@tanstack/ai-openai`, `@tanstack/ai-anthropic`, and most other packages). The CLAUDE.md line about colocated `*.test.ts` describes the general unit-test pattern; the packages that ship conformance suites + CLI tests deliberately keep them under `tests/` so fixtures and multi-file harnesses do not mix with the published `src/` tree that `tanstackViteConfig` builds from.

Moving only this package would diverge from the rest of the workspace without a product benefit. Happy to revisit if the repo standardizes on colocated tests workspace-wide.

packages/ai-persistence-prisma/vite.config.ts
```diff
@@ -0,0 +1,37 @@
...
+
+const config = defineConfig({
+  test: {
+    name: packageJson.name,
+    dir: './',
+    watch: false,
+
+    globals: true,
+    environment: 'node',
+    include: ['tests/**/*.test.ts'],
```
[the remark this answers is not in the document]
tombeckenham: Same answer as the drizzle twin: `tests/**/*.test.ts` is the monorepo standard for library packages (conformance + CLI tests under `tests/`, published modules under `src/`). Not changing this package alone.

testing/e2e/tests/persistence-durability.spec.ts
```diff
@@ -0,0 +1,109 @@
...
+ *
+ * Provider-free: the `/api/persistence-durability` harness route streams a fixed
+ * AG-UI sequence through a `memoryStream` delivery sink, so there is no LLM in
+ * the loop and nothing to mock (exempt from the aimock policy).
+ *
+ * The mid-stream "rejoin an in-flight run via joinRun after reload" path is NOT
+ * covered here: the harness stream completes in a single tick, so there is no
+ * deterministic window to reload while a run is still producing. That resume
+ * cursor is covered at the transport layer by `delivery-durability.spec.ts` and
+ * in `@tanstack/ai-client` unit tests.
```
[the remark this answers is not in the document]
tombeckenham: Acknowledged — mid-stream reload/`joinRun` is intentionally out of this E2E file (see the header comment in `persistence-durability.spec.ts`).

What is covered today:

- **Unit** (`@tanstack/ai-client`): combined `{ messages, resume }` record, async hydrate rejoin, `resumeInFlightRun` / `applyResumeSnapshot`, and the explicit-`undefined` `joinRun` guard on the connection adapter.
- **E2E**: transcript restore after clean finish, and interrupt survival across reload (local resume snapshot + server-authoritative reconstruction).

A deterministic mid-stream pause + reload harness needs a fixture that freezes the stream mid-token, reloads, asserts the client issues the resume/join request, and checks single continuation — that is real work (aimock fixture + transport coordination) and is tracked as follow-up rather than blocking this PR. The durable-stream e2e (`e87998db8`) already guards full agent-loop delivery through durability-backed streams; the missing piece is specifically the *browser reload while streaming* path.

docs/persistence/overview.md
```diff
@@ -0,0 +1,152 @@
...
+
+```tsx
+import { fetchServerSentEvents, localStoragePersistence } from '@tanstack/ai-client'
+import { useChat } from '@tanstack/ai-react'
+import type { ChatPersistedState } from '@tanstack/ai-client'
+
+const store = localStoragePersistence<ChatPersistedState>({
+  serialize: (value) => JSON.stringify(value),
+  deserialize: (value) => JSON.parse(value),
+})
```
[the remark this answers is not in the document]
tombeckenham: Outdated against the current docs. The shipped client adapters (`localStoragePersistence` / `sessionStoragePersistence` / `indexedDBPersistence`) already own serialize/deserialize and tolerate corrupt records (they fall back rather than throwing into `useChat`). Overview samples that still show a hand-rolled adapter are the escape hatch; for those, a Standard Schema/`typeof` guard is good practice and can be added when that snippet is next edited — not blocking, and not worth a drive-by docs-only change on an outdated thread.

docs/persistence/overview.md
```diff
@@ -0,0 +1,241 @@
...
+  fetchServerSentEvents,
+  localStoragePersistence,
+  useChat,
+} from '@tanstack/ai-react'
+
+function Chat() {
+  const { messages, sendMessage } = useChat({
+    threadId: 'support-chat',
+    connection: fetchServerSentEvents('/api/chat'),
+    persistence: localStoragePersistence(),
```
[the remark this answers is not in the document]
tombeckenham: The docs and example already treat thread identity as **app-owned and authorized server-side**:

- `reconstructChat({ authorize })` is the documented gate — store methods take bare `threadId`s; ownership is checked at the route boundary (see `docs/persistence/custom-stores.md` invariants and the persistent-chat example).
- The multi-thread example scopes threads per session/user rather than a single global `support-chat` id.
- Literal ids in overview snippets are minimal demos of the *API shape*, not a multi-tenant recipe.

We will not put a full auth system into every sample, but the security boundary is explicit: authorize before load/save. If a remaining snippet still reads as “share one global id across users,” point at the line and we will reword it to `threadId` from session + `authorize`.

docs/persistence/custom-stores.md
tombeckenham: Agreed — most apps fold chat into the existing customer DB rather than standing up a separate store.

That is now spelled out in `docs/persistence/custom-stores.md` under **When each store is called**:

1. **Chat middleware lifecycle** table — which store methods fire on `onConfig` / `onStart` / stream `onChunk` / interrupt boundary / `onFinish` / `onError` / `onAbort`, and hard-fail vs best-effort.
2. **Swimlanes (request → store)** — ASCII flow from client POST through load/createOrResume/listPending → saveThread/update/commit.
3. **Invariants** — full-overwrite `saveThread`, idempotent `createOrResume`, insert-if-absent interrupts, authorize-at-boundary, locks via `withLocks` not the state bag.

If anything is still thin for the integrate-with-my-tables path (e.g. mapping each method onto a concrete `threads`/`messages`/`approvals` schema, or transaction boundaries across stores), say which event you want expanded and we can add a worked example in a follow-up.

docs/persistence/drizzle.md
tombeckenham: Not every Drizzle dialect is first-class in this package (yet).

Shipped providers:

- **`provider: 'sqlite'`** — including the Node `node:sqlite` convenience factory and D1 (via the Cloudflare package, which delegates to the drizzle SQLite path).
- **`provider: 'pg'`** — BYO Postgres Drizzle DB (node-postgres, postgres.js, Neon, etc.) + `/pg-schema`.

MySQL / libsql / other Drizzle dialects are not typed or schema-codegen’d here. The runtime needs dialect-specific table definitions and type maps; we intentionally narrowed to the two backends we test and document rather than claiming “any Drizzle DB.” Adding another dialect is mostly schema + provider discriminant + conformance — happy to take MySQL as a follow-up if that is the missing one for you.

packages/ai-persistence/src/types.ts
tombeckenham: Agreed — and this is already landed on the branch (see `43cd409d5` and the current `packages/ai-persistence/src/types.ts`).

- **`AIPersistenceStores` / named chat shapes** hold only **state** stores (`messages` / `runs` / `interrupts` / `metadata`). Locks are not on these types.
- **Locks** are a separate capability: `LockStore` + `withLocks(...)` / `LocksCapability` (and the shared core token used by sandbox in the stacked PR).
- **Optional vs required:** the sparse bag stays optional for composition, but the **named product shapes** require what they claim — e.g. `ChatTranscriptStores.messages` is required, `ChatPersistenceStores` requires all four state stores. Prefer those named types at call sites over the sparse bag.

On the “all other fields optional” feel: that was the sparse internal bag; packaged backends return `ChatPersistenceStores` (all present). Custom apps that only need a transcript use `ChatTranscriptStores`.

jherr (approved):

PR #987, opened 2026-07-23
feat(persistence): generation persistence — client snapshot + durable media bytes

AlemTuzlak:
## Generation Persistence

Stacked on **#984** (`feat/persistence-core`) — please review/merge that first. Rebased onto its current tip.

Generation persistence in two halves:

### 1. Client resume snapshot (read-only)
As a media generation streams, the client builds a lightweight `GenerationResumeSnapshot` (run identity, status, errors, result metadata + artifact refs — **never** the bytes) and writes it to a `persistence` storage adapter. The option reuses the shared `ChatStorageAdapter` contract, so `localStoragePersistence` / `sessionStoragePersistence` / `indexedDBPersistence` work with no type argument:

```tsx
const snapshots = localStoragePersistence({ keyPrefix: 'my-app:generation:' })
useGenerateImage({ connection, persistence: snapshots })
```

Hooks expose `resumeSnapshot` / `resumeState` / `pendingArtifacts` / `resultArtifacts` across react/solid/vue/svelte/angular. No `resume()` action; reconnect to an in-flight stream is the delivery layer's job (resumable streams, #955, merged), wired via a `durability` adapter + `GET` handler exactly as for chat.

### 2. Durable media-byte storage (server, opt-in)
When the persistence backend provides both an `artifacts` (`ArtifactStore`) and a `blobs` (`BlobStore`) store, `withGenerationPersistence` writes each generated file's bytes to the blob store (key `artifacts/<runId>/<artifactId>`), records an `ArtifactRecord`, attaches `PersistedArtifactRef`s to the result, and emits `generation:artifacts` so the client records them. `memoryPersistence()` ships both stores; any backend implementing the two contracts works. Extraction is customizable via `extractArtifacts` / `nameArtifact`.

### Changes
- **`@tanstack/ai`** — result-transform machinery (`resultTransforms` / `artifactInputs` on `GenerationMiddlewareContext`, `applyGenerationResultTransforms`), `threadId`/`runId` options on the image/audio/speech/transcription activities, and `generation:artifacts` emission from `streamGenerationResult`.
- **`@tanstack/ai-utils`** — `base64ToUint8Array`.
- **`@tanstack/ai-persistence`** — `ArtifactStore` + `BlobStore` contracts, in-memory impls in `memoryPersistence()`, and byte-persistence in `withGenerationPersistence` (`extractArtifacts`/`nameArtifact`).
- **`@tanstack/ai-client`** + **framework hooks** (react/solid/vue/svelte/angular) — the client snapshot + `persistence` option + artifact refs.
- **`@tanstack/ai-event-client`** — optional `threadId`/`runId` on generation events.
- **Docs** — `docs/persistence/generation-persistence.md` (client snapshot, byte storage + serve route, reconnect), kiira-verified.
- **Changeset** — minor across the affected packages.

### Verification
- Builds green across ai-utils, ai, ai-event-client, ai-client, ai-persistence.
- `test:types` — **0 errors** in ai / ai-client / ai-persistence.
- Tests — ai-persistence **78 pass** (incl. 12 artifact + the chat resume suites), ai-client generation **50 pass**, ai core stream/middleware/activity **92 pass**.
- kiira — doc passes (3 snippets).

### Scope note
Durable SQL/R2 artifact+blob backends (Drizzle/Prisma/Cloudflare R2) are **not** in this PR: `memoryPersistence` plus any custom `ArtifactStore`+`BlobStore` cover the contract today, and the durable backends (which add SQL schema + migrations per package) are a clean follow-up. `generateVideo`'s own job-polling artifact path is likewise deferred (image/audio/speech/transcription persist bytes).


tombeckenham:
Superseded by a two-PR stack — closing this one, but **the `feat/generation-persistence` branch is staying put**, so nothing here is lost.

- #997 — client-side generation resume snapshot (base `feat/persistence-core`)
- #998 — durable generation media-byte storage (stacked on #997)

### Why the split

Two things happened to the base branch after this PR was opened.

`feat/persistence-core` was rebased, so 25 of this PR's 37 commits ended up duplicated on the base under different hashes. That inflated the diff here to 212 files / +19k when the actual new work is 12 commits / 72 files / +5.2k.

More importantly, `62c99754d refactor(persistence)!: ship the contract, not the backends` dropped `@tanstack/ai-persistence-drizzle`, `-prisma`, and `-cloudflare`. This branch still carries `3e0e477b4`, which adds them — that's the whole source of the `CONFLICTING` state, and merging as-is would resurrect the deleted packages. `978996e3b` (Prisma 7 support) is dead for the same reason.

Rebasing would have meant replaying those 25 duplicates against a force-pushed base, so the stack was rebuilt by cherry-picking only the 12 generation-persistence commits onto the current base instead. The history split cleanly at `c7a901f46` with zero package overlap between the two halves, which is what made two PRs the natural shape rather than one.

Both new PRs are `MERGEABLE`.

### What changed in the rebuild

The client half (#997) cherry-picked with no manual intervention.

The server half (#998) needed real reconciliation with the reworked `ai-persistence`, isolated in a single trailing `fix(persistence): reconcile …` commit so it's reviewable separately from the cherry-picks. In short: `memoryPersistence()` keeps the base's no-locks decision and gains only `artifacts` + `blobs`; the generation validator now carries both invariants (`runs` required, and `artifacts`/`blobs` as a pair); and the base's "⚠️ TEMPORARY / WRONG SHAPE" warning on `withGenerationPersistence` is preserved rather than dropped, since the byte path doesn't actually resolve the `requestId`-vs-`runId` keying it warns about. Full details in #998's description.

### Heads up on review findings

Both PRs are **drafts** and each carries a "Known issues" checklist — the code is as authored here, so the review findings are documented rather than fixed. The two blockers:

- **#997** — `dispose()` sets `disposed = true` and nothing ever clears it, so under React StrictMode's mount→cleanup→mount replay every subsequent `generate()` call is a silent no-op. Reproduced under `{ wrapper: StrictMode }`; StrictMode is on by default in Vite/Next React templates. The fix pattern already exists at `devtools.ts:571`.
- **#998** — `ArtifactStore` / `BlobStore` ship with documented invariants but no conformance-testkit suites, which `types.ts` calls "the authoritative compatibility gate". Now that backends are app-owned, that testkit is the only thing holding an implementation to the rules.

Thanks @AlemTuzlak — the work carried over intact, it just needed a different base.

PR #997, opened 2026-07-26
feat(persistence): client-side generation resume snapshot

tombeckenham:
## 🎯 Changes

Generation persistence — the media-generation parallel of chat persistence. A media run (image, video, audio, TTS, transcription, summarize) now survives a page reload or a dropped connection, restoring **transparently into the normal hook fields**, with optional durable storage of the generated bytes.

Split out of #987 and rebuilt on the current `feat/persistence-core`. The previously-stacked byte-storage half has been folded in, so this is now the whole feature in one PR.

### The shape

A generation run is recorded in its own store, keyed by its `runId` (the same AG-UI run id the client sends), with `threadId` an optional link — it no longer overloads the chat `RunStore`. On the client, the hook picks a mode exactly the way `useChat` does:

```tsx
// Client-driven: the browser holds a lightweight snapshot under `generation:<id>`.
useGenerateImage({ id: 'hero', connection, persistence: localStoragePersistence() })

// Server-driven: nothing is cached; the last run for `threadId` is fetched on mount.
useGenerateImage({ threadId, connection, persistence: true })
```

Restore is invisible. There is no `resumeSnapshot` / `pendingArtifacts` / `resultArtifacts` hook field — a reload repaints `result` / `status` / `error` as if the run had just finished. `resumeState` carries the in-flight run identity only.

Crucially, this restores a **record**, not provider work: generation still only begins when `generate(...)` is called, and nothing is ever restarted. A run that is still streaming *is* re-attached and finished in place, via the connection's `joinRun` durability replay — the same mechanism `useChat` uses.

### Server

- **`withGenerationPersistence`** records each run in a dedicated `generationRuns` (`GenerationRunStore`) store: activity/provider/model, lifecycle status, result metadata, and — when byte storage is on — the durable artifact refs.
- **`reconstructGeneration(persistence, request, options?)`** is the generation parallel of `reconstructChat`: it reads `?runId=` (preferred) or the latest run linked to `?threadId=`, authorizes via `authorize`, and returns `{ resumeSnapshot, activeRun }` JSON for a server-authoritative client to hydrate from.
[...]

PR #998, opened 2026-07-26
feat(persistence): durable generation media-byte storage

tombeckenham:
## 🎯 Changes

**Stack 2 of 2.** Stacked on #997 — please review/merge that first; this PR's base is its branch, so the diff here is only the server half. Split out of #987 and rebuilt on the current `feat/persistence-core`.

Durable media-byte storage for generations (server, opt-in).

When the persistence backend provides **both** an `artifacts` (`ArtifactStore`) and a `blobs` (`BlobStore`) store, `withGenerationPersistence` writes each generated file's bytes to the blob store (key `artifacts/<runId>/<artifactId>`), records an `ArtifactRecord`, attaches `PersistedArtifactRef`s to the result, and emits `generation:artifacts` so the client records them. Extraction is customizable via `extractArtifacts` / `nameArtifact`.

`memoryPersistence()` now ships both stores; any backend implementing the two contracts works. To serve a stored artifact, `@tanstack/ai-persistence` exports `retrieveArtifact(persistence, id)` and `retrieveBlob(persistence, idOrRecord)` (plus `artifactBlobKey`).

### Packages

- **`@tanstack/ai`** — result-transform machinery (`resultTransforms` / `artifactInputs` on `GenerationMiddlewareContext`, `applyGenerationResultTransforms`), `threadId` / `runId` options on the image/audio/speech/transcription activities, and `generation:artifacts` emission from `streamGenerationResult`.
- **`@tanstack/ai-utils`** — `base64ToUint8Array`.
- **`@tanstack/ai-persistence`** — `ArtifactStore` + `BlobStore` contracts, in-memory impls in `memoryPersistence()`, byte persistence in `withGenerationPersistence`, `retrieveArtifact` / `retrieveBlob`.
- **Docs** — the byte-storage + serve-route half of `docs/persistence/generation-persistence.md`.

### Reconciling with the contract-only core

The original commits predate `62c99754d refactor(persistence)!: ship the contract, not the backends`, so the rebase onto the reworked `ai-persistence` needed real decisions. These are isolated in the final `fix(persistence): reconcile …` commit, separate from the cherry-picks:

- `memoryPersistence()` keeps the base's **no-locks** decision and gains only `artifacts` + `blobs`; locks stay separate via `InMemoryLockStore` + `withLocks`. Same for `AIPersistenceStores` / `storeKeys`.
- `validateGenerationPersistenceStores` and `InvalidGenerationPersistence` now carry **both** invariants: `runs` is required (base), and `artifacts` / `blobs` come as a pair (this PR).
- The base's "⚠️ TEMPORARY / WRONG SHAPE" warning on `withGenerationPersistence` is **preserved**. The original patch deleted it without fixing the underlying problem, so the note now also records the `requestId`-vs-`runId` mismatch below.
- Two tests adjusted: the type test no longer pins `memoryPersistence()` to exact `ChatPersistence` equality, and the artifact-pairing test gains a `runs` store so it exercises the pairing rule rather than tripping the required-runs rule first.

### ⚠️ Known issues — not yet fixed on this branch

- [ ] **Blocker: the new store contracts skip the conformance testkit.** `types.ts` states the testkit is "the authoritative compatibility gate" and that new invariants must be promoted into it. `ArtifactStore` / `BlobStore` add documented invariants (notably `BlobStore.list` cursor/prefix semantics) with zero suites. Now that backends are app-owned, this is the only thing holding an implementation to the rules.
- [ ] **`RunRecord.runId` and `ArtifactRecord.runId` disagree.** The run record is keyed on `ctx.requestId`; artifacts use `ctx.runId ?? ctx.requestId`. With a caller-supplied `runId`, `runs.get(runId)` is empty while `artifacts.list(runId)` returns rows — the stores can't be joined. This is exactly what the preserved TEMPORARY/WRONG-SHAPE note warns about.
- [ ] **Video artifacts are never persisted.** `generateVideo` is the only media activity with no `artifactInputs` / `applyGenerationResultTransforms`, yet `middleware.ts` carries an unreachable `activity === 'video'` branch and the docs lead with video as the motivating case. Either wire it or drop the dead branch and scope it out explicitly.
- [ ] **An artifact-persistence failure destroys a successful generation.** The `fetch(descriptor.url)` throw isn't isolated by `applyGenerationResultTransforms`, so a blip fetching an expiring provider URL discards the paid-for result, fails the run, and orphans already-written blobs.
- [ ] Streamed artifacts record `size: 0`, and the documented serve route sends `content-length: 0` alongside a real body.
- [ ] The documented serve route has no authorization — `GET ?id=<artifactId>` returns any artifact's bytes, and `artifactId` is `Date.now()` + 5 chars of `Math.random()`.
- [ ] Minor: `sourcePartDescriptors` defaults mime to `${type}/mpeg` (yielding invalid `image/mpeg`); `parseDataUrl` doesn't match parameterized data URLs and its fallback base64-decodes the whole URL.
- [ ] No E2E coverage yet.

## ✅ Checklist

- [ ] I have tested this code locally with `pnpm run test:pr`. — ran the affected subset: `build`, `test:types`, `test:lib` (ai-persistence at 99 tests), `test:oxlint` all green.

## 🚀 Release Impact


tombeckenham:
Note: the durable media-byte storage from this stacked PR has been folded into #1011, which now ships the whole generation-persistence feature (client + server) in one PR.

Issue #1005, opened 2026-07-28
Structured output: support response_format: json_object for providers without strict json_schema


daveycodez:
## Summary

`outputSchema` always sends `response_format: { type: "json_schema", strict: true }`. Providers that accept `response_format: { type: "json_object" }` but do **not** support strict structured outputs reject the request, so those models are unreachable through the library — even though the OpenRouter SDK, and TanStack AI's own core types, already model `json_object`.

There is no opt-out: setting `modelOptions.responseFormat = { type: "json_object" }` alongside `outputSchema` does not take effect (see [Why the override does not work](#why-the-override-does-not-work) — it is a spread-ordering issue, not a validation one).

Versions: `@tanstack/ai@0.42.0`, `@tanstack/ai-openrouter@0.15.10`, bundled `@openrouter/sdk@0.13.20`.

## Reproduction

```ts
import { chat } from '@tanstack/ai'
import { openRouterText } from '@tanstack/ai-openrouter'
import { z } from 'zod'

// qwen3.7-flash declares response_format + tools, but NOT structured_outputs
await chat({
  adapter: openRouterText('qwen/qwen3.7-flash'),
  messages: [{ role: 'user', content: 'Name three colors as json.' }],
  outputSchema: z.object({ colors: z.array(z.string()) }),
})
```

Result:

```
Error: Provider returned error
    at runAgenticStructuredOutput (@tanstack/ai/dist/esm/activities/chat/index.js:1671:17)
```

The upstream detail never reaches the caller. The actual provider response is:

```
400 invalid_parameter_error
'messages' must contain the word 'json' in some form, to use 'response_format' of type 'json_object'
```

(That message is the provider's reply to OpenRouter's own downgrade attempt; the point is that the caller sees none of it.)

## Scope

Against the live OpenRouter model list (`GET /api/v1/models`, 341 models at time of writing):

| `supported_parameters` | count |
| --- | ---: |
| `response_format` **and** `structured_outputs` | 258 |
| `response_format` but **not** `structured_outputs` | **28** |
| neither | 55 |

Those 28 are usable today only if the caller gives up `outputSchema` entirely. Verified individually via the API: `qwen/qwen3.7-flash`, `qwen/qwen3-30b-a3b-thinking-2507`, `inclusionai/ring-2.6-1t`, `aion-labs/aion-3.0`, `z-ai/glm-5-turbo`, `minimax/minimax-m2.1`, `ibm-granite/granite-4.0-h-micro`, and others.

[...]

PR #1014, opened 2026-07-29
fix(ai-openrouter): honor json_object structured output

kinKingen:
## Changes

Partially addresses #1005.

- Honor the existing typed `modelOptions.responseFormat: { type: 'json_object' }` option during OpenRouter structured-output calls.
- Do not let native combined-mode `json_schema` overwrite that opt-in.
- Preserve strict `json_schema` as the default. Other caller response formats are ignored so the activity-layer schema cannot drift from the provider request.
- Cover non-streaming and streaming paths with unit tests and add a real OpenRouter SDK-to-HTTP wire regression test.
- Add a patch changeset for `@tanstack/ai-openrouter`.

### Scope

This is the OpenRouter compatibility fix for the existing typed `modelOptions.responseFormat: { type: 'json_object' }` path. It does not add a core `structuredOutputMode` API, generated capability tables, prompt mutation, or retry.

Callers targeting models that reject strict `json_schema` opt in:

```ts
await chat({
  adapter: openRouterText('qwen/qwen3.7-flash'),
  messages: [{ role: 'user', content: 'Name three colors as json.' }],
  outputSchema: z.object({ colors: z.array(z.string()) }),
  modelOptions: { responseFormat: { type: 'json_object' } },
})
```

### Root cause

Both OpenRouter structured-output methods rebuilt `responseFormat` as strict `json_schema` after spreading the mapped request, silently overwriting the caller's declared `json_object` option. Combined mode (`mapOptionsToRequest`) had the same overwrite after #836. This patch resolves the format once: explicit `json_object` is preserved, everything else keeps the current strict schema from `outputSchema`.

## Checklist

## Release Impact

## AI assistance disclosure

This contribution was developed with AI assistance. I reviewed the implementation, tests, and validation results and take responsibility for the submitted changes.

## Summary by CodeRabbit

* **New Features**
  * OpenRouter structured output now honors explicitly selected JSON-object response formats for streaming and non-streaming requests.
  * Strict JSON-schema output remains the default.
* **Bug Fixes**
  * Combined tool and schema requests now correctly handle explicitly selected JSON-object output.
* **Tests**
  * Added coverage for JSON-object responses across standard and end-to-end request flows.
  * Added validation to ensure response format settings are preserved through OpenRouter requests.

### DIFF
--- a/.changeset/openrouter-json-object-structured-output.md
+++ b/.changeset/openrouter-json-object-structured-output.md
@@ -1,0 +1,6 @@
+---
+'@tanstack/ai-openrouter': patch
+---
+
+Honor `modelOptions.responseFormat: { type: 'json_object' }` during structured
+output generation while preserving strict `json_schema` as the default.
--- a/packages/ai-openrouter/src/adapters/text.ts
+++ b/packages/ai-openrouter/src/adapters/text.ts
@@ -204,28 +204,32 @@ export class OpenRouterTextAdapter<
   }
 
   /**
-   * Generate structured output via OpenRouter's `responseFormat: { type:
-   * 'json_schema', jsonSchema: ... }` (camelCase). Uses stream: false to get
-   * the complete response in one call.
-   *
-   * The outputSchema is already JSON Schema (converted in the ai layer).
-   * We apply OpenAI-strict transformations for cross-provider compatibility.
+   * Generate structured output via OpenRouter's `responseFormat`. Uses
+   * `stream: false`. Default is strict `json_schema` from `outputSchema`.
+   * Callers can opt into JSON mode with
+   * `modelOptions.responseFormat: { type: 'json_object' }`.
    */
   async structuredOutput(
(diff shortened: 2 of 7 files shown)


kinKingen:
Resolved the conflict against the current main while retaining both the upstream #1076 structured-output usage coverage and this PR's json_object regression coverage. I also verified that remote merge commit `802f95e` has the same tree as the locally tested rebased head.

Validation on the resolved tree:
- OpenRouter package: 158 tests passed
- focused SDK-to-wire E2E: 1 passed
- `pnpm test:pr`: passed
- full E2E: 453 passed, 1 skipped, 1 flaky; one durable-takeover test failed consistently
- the exact durable-takeover test also fails on current main with the same `detachedSince` assertion, so it is unrelated to this PR

The refreshed GitHub Actions workflows are now awaiting maintainer approval.

AI assistance was used for the conflict resolution and verification. I reviewed the exact merged tree, diff, test output, and conclusions.

tombeckenham (approved):

PR #1011, opened 2026-07-29
feat(persistence): generation run persistence (client + server)

tombeckenham:
## 🎯 Changes

Generation persistence — the media-generation parallel of chat persistence. A media run (image, video, audio, TTS, transcription, summarize) now survives a page reload or a dropped connection, restoring **transparently into the normal hook fields**, with optional durable storage of the generated bytes.

Split out of #987 and rebuilt on the current `feat/persistence-core`. The previously-stacked byte-storage half has been folded in, so this is now the whole feature in one PR.

Supersedes #997 (same change, renamed from `feat/generation-persistence-client` — the branch always contained the server half too) and folds in #998.

### The shape

A generation run is recorded in its own store, keyed by its `runId` (the same AG-UI run id the client sends), with `threadId` an optional link — it no longer overloads the chat `RunStore`. On the client, the hook picks a mode exactly the way `useChat` does:

```tsx
// Client-driven: the browser holds a lightweight snapshot under `generation:<id>`.
useGenerateImage({ id: 'hero', connection, persistence: localStoragePersistence() })

// Server-driven: nothing is cached; the last run for `threadId` is fetched on mount.
useGenerateImage({ threadId, connection, persistence: true })
```

Restore is invisible. There is no `resumeSnapshot` / `pendingArtifacts` / `resultArtifacts` hook field — a reload repaints `result` / `status` / `error` as if the run had just finished. `resumeState` carries the in-flight run identity only.

Crucially, this restores a **record**, not provider work: generation still only begins when `generate(...)` is called, and nothing is ever restarted. A run that is still streaming *is* re-attached and finished in place, via the connection's `joinRun` durability replay — the same mechanism `useChat` uses.

### Server

- **`withGenerationPersistence`** records each run in a dedicated `generationRuns` (`GenerationRunStore`) store: activity/provider/model, lifecycle status, result metadata, and — when byte storage is on — the durable artifact refs.
- **`reconstructGeneration(persistence, request, options?)`** is the generation parallel of `reconstructChat`: it reads `?runId=` (preferred) or the latest run linked to `?threadId=`, authorizes via `authorize`, and returns `{ resumeSnapshot, activeRun }` JSON for a server-authoritative client to hydrate from.
[...]

PR #988, opened 2026-07-23
feat(ai-sandbox): SandboxInstanceStore instance durability (BYO)

AlemTuzlak:
## Summary

Durable **sandbox instance** resume for multi-process / multi-replica deploy, owned by `@tanstack/ai-sandbox`.

Based on **main** (after #984). Locks already live in `@tanstack/ai/locks` — this PR does **not** re-ship them. Chat persistence is independent: no `stores` bag key for sandbox instances.

### API

- **`SandboxInstanceStore` / `SandboxInstanceRecord` / `InMemorySandboxInstanceStore` / `SandboxInstanceStoreCapability`**
- **`withSandboxInstanceStore(store)`** — provide before `withSandbox`
- **`defineSandboxInstanceStore`** — inline BYO typing (same pattern as `defineLock` / `defineMessageStore`)
- **`withSandbox`** consumes the capability in `ensure` (in-memory fallback when absent)
- Multi-instance: pair with **`withLocks`** from `@tanstack/ai/locks`
- Conformance: `runSandboxInstanceStoreConformance` from `@tanstack/ai-sandbox/testkit`

### Docs / e2e

- `docs/sandbox/durability.md` (Instance Durability)
- Locks guide links to durability
- E2E: `sandbox-durability` (store shared across independent `chat()` runs)

### Test plan

- [x] Unit: in-memory store, conformance suite, ensure algorithm, resume-from-store middleware
- [x] E2E durability route + spec
- [ ] CI green on PR

## Summary by CodeRabbit

* **New Features**
  * Added durable sandbox instance resume across processes and replicas using shared storage.
  * Added configurable instance storage and distributed locking, with in-memory fallback for single-process use.
  * Added tools for validating custom storage implementations.
* **Breaking Changes**
  * Replaced legacy sandbox store integrations with instance store APIs.
* **Documentation**
  * Added a sandbox instance durability guide and updated related guidance.
* **Tests**
  * Added coverage for durable resume, locking, and storage behavior.

PR #1015, opened 2026-07-29
feat: durable, reconnectable sandboxed agent runs

AlemTuzlak:
Durable, reconnectable sandboxed agent runs. A client disconnect **detaches** the run instead of destroying the sandbox; a later request **takes it over** and keeps streaming; a **reaper** finalizes or expires runs nobody comes back for.

Verified end to end by `testing/e2e/tests/durable-takeover.spec.ts` — a real mid-stream disconnect, an attach, and the stream continues with the transcript intact and no duplicated prefix.

> **Note on the base.** This PR was opened against `feat/persistence-sandbox`, which was later deleted upstream, so GitHub retargeted it to `main`. `main` has since squash-merged that foundation itself (#988 sandbox instance durability, #1011 generation run persistence), and `main` is now merged into this branch — so the PR no longer carries the foundation and there is nothing to split out. One resolution worth flagging: #1004 made `RunStore.findActiveRun` **required**, but this branch had relocated `RunStore` into `@tanstack/ai`, and that file merged *cleanly* — so the merge had to re-apply the requirement by hand or it would have silently reverted #1004. It is now required in core, gone from the conformance suite's `skipMethods`, and forwarded unconditionally by `fenceRunStore`.

## The problem

You are building something like ChatGPT, except the assistant is a coding agent that works inside a sandbox and can take ten minutes to finish a task.

Ten minutes is a long time for a browser tab. The user refreshes. Closes the laptop. Loses wifi. Or their next request lands on a different replica than the one running the job.

## What used to happen

The connection dropping killed everything. The sandbox was destroyed, the work was thrown away, and the user came back to nothing.

That was not a bug — it was the least-bad option available. When you close the pipe to an agent running in a sandbox, **the agent does not stop.** It keeps working and keeps spending money on tokens. So destroying the sandbox was the only reliable way to be sure a disconnected job stopped burning cash.

Which is exactly right if the user pressed **Stop**. And completely wrong if they just refreshed the page.

## What this PR changes

It teaches the system to tell those two situations apart, and to handle the refresh case properly. Four pieces:

**1. A disconnect no longer kills anything.** It *detaches*: the agent keeps working, the sandbox stays up, and the run record notes "nobody is watching this, as of 3:42pm."

**2. The agent writes its output to a file instead of down the wire.** This is the key trick. If the agent talks directly to the browser, its words vanish the moment the browser leaves. If it writes to a file inside the sandbox, the words are still sitting there when someone comes back.

**3. Someone comes back, and we pick up mid-sentence.** A new request — possibly on a different replica — reads that file, works out how much the user already saw, and streams only the part they missed. No repeated paragraphs, no gaps.

**4. Something has to clean up after the people who never come back.** Otherwise a sandbox runs forever on a job nobody will ever read. So there is a sweeper: it finds abandoned runs, checks whether the agent finished on its own, and either wraps them up or shuts them down.

## The two parts that sound over-engineered but are not

**Making sure two replicas never both drive one run.** If a user opens the same thread in two tabs, or a load balancer sends a retry elsewhere, two replicas could both try to continue one run — and the user would see doubled text and contradictory "finished" messages. So a replica has to take a numbered ticket to drive a run, and if a newer replica takes a higher number, the older one is locked out of writing anything at all. Not merely discouraged — unable to append to the log or mark the run finished.

**Making "the agent finished" impossible to fake.** The way we know a run ended is that a special line appears at the end of that file. But the *agent itself* writes that file, and agents write whatever the model says. If a model happened to print that line, the sweeper would believe a running job had ended and would shut down a live sandbox. So the line includes a secret value derived from the run's own id. The agent cannot produce it, so it cannot get its own sandbox destroyed.

## What you have to do to use it

Two things, and the second is easy to forget:

1. **Turn it on** — give `withSandbox` both a run store and a durability backend. Passing only one leaves you with exactly today's destroy-on-disconnect behavior, silently, because you have not asked for durability.
2. **Actually schedule the sweeper** — cron, a queue, a Durable Object `alarm()`, whatever the platform offers.

Do the first and not the second and everything *looks* fine, then sandboxes bill indefinitely and disconnected readers wait forever on logs nothing will ever close. A real distributed `LockStore` is also required; the in-memory one cannot coordinate across hosts.

This explanation ships as [`docs/sandbox/durable-runs.md`](docs/sandbox/durable-runs.md), ordered ahead of the three wiring pages.

---

## What ships

**Run lifecycle** (`@tanstack/ai`) — `RunRecord` gains `sandboxKey`, `detachedSince`, `driverEpoch`, `cancelRequested`; `RunStore.listReclaimable`; `isTerminalRunStatus` / `isRunStatus`; out-of-band cancel via `requestRunCancel` / `wasCancelRequested`.

**Detach on disconnect** (`@tanstack/ai-sandbox`) — `withSandbox({ runs, durability })` records `detachedSince` + `sandboxKey` and leaves the sandbox up. A detached run's delivery log stays **open** (`RunDetachedCapability`) so a successor can continue it.

**Takeover** — `sandboxRunDriver` claims a run under a lease, fences it by epoch, waits for quiescence, replays from byte 0 and aligns against the stored log so only the remainder is appended. A superseded driver can write **neither** the event log **nor** a terminal run status.

**Reaping** — `reapDetachedRuns`, `pruneJournals`, `reclaimSandbox` / `sandboxReclaimer`. The reaper never drives a run to discover whether it finished: an injected `hasFinished` probe reads the in-sandbox journal out of band, because entering the drive writes a terminal status and closes the log on every path.

**Journal** — a nonced, unforgeable exit sentinel; injective filename encoding; a fail-closed decoder; bounded attach preflight (`JournalAttachUnavailableError`).

## Breaking changes

All pre-1.0, and the durability surface has never been published, so these break no released consumer.

- `onAbort` writes `aborted` (terminal, with `finishedAt`); `interrupted` is no longer terminal-shaped.
- `RunDeps.durability` is a per-run factory `(runId) => StreamDurability<TOffset>`.
- `SandboxDurabilityOptions.detachedRunTtl` is **removed** — it was validated, parsed, and read by nothing. `ReapOptions.detachedRunTtlMs` is the only TTL.
- `SandboxCapabilities.killableProcesses` is now required.

## Read this before enabling durability

**The reaper ships as a function, not a scheduler.** An app that wires `durability` and never calls `reapDetachedRuns` has nothing closing detached delivery logs: tailers park forever, the TTL is enforced by nothing, and sandboxes bill indefinitely. Wiring durability and scheduling the sweep are two separate integration steps. See `docs/sandbox/reaping.md`.

A real `LockStore` is required — `InMemoryLockStore` cannot coordinate across hosts, and `withSandbox` warns when it is used with durability.

## Review

34 fix commits across four review rounds (8 reviewers on the feature, then re-reviews of the fixes themselves). Notable findings, all fixed:

- The detach guard included `!terminalPersisted`, and an agent-loop run emits one `RUN_FINISHED` per iteration — so detach was defeated for **every tool-calling run**.
- `durableStream` restarted its sequence at 1 on takeover, so a takeover's appends were silently discarded by the reader's dedup.
- Three of four `killableProcesses: true` declarations were **false** when measured: Docker's `stream.destroy()` detached only the client, local-process POSIX killed a forking shell, and Vercel's `kill()` was a no-op that never called the SDK's real kill. Vercel and Daytona are now `false`; Docker's is fixed and falsifiable.
- The exit sentinel was forgeable from ordinary agent stdout, which could get a live sandbox destroyed by the reaper.
- An attach against a run with no journal hung forever, and self-perpetuated once triggered.
- `isTerminalRunStatus` used `in`, so `'toString'` was terminal — reachable from any user-implemented store, and a consumer deletes journals.
- One middleware's failing teardown cancelled every later middleware's teardown.
- The conformance suite asserted the *reader* stopped, never that the remote process died — which is how the false capability claims survived it. It now asserts the process is gone.

## Known limitations

- **Sprites `killableProcesses: true` is unmeasured** — the one remaining `true`. Its server-side kill endpoint is confirmed to be *issued*; what it signals (process group vs. pid) is not. Needs `SPRITES_API_KEY`.
- Vercel, Daytona, and Sprites journal-conformance suites are registered but render **named skips** without credentials. A secrets-gated nightly job would close this.
- `packages/ai-sandbox/tests/harness-cwd.test.ts › maps nested virtual paths under /workspace on local-process` fails on Windows (path separators). Pre-existing and unrelated.
- Event-log retention stays the application's job, inside its own `StreamDurability` backend.

## Verification

`ai` 1390 · `ai-sandbox` 602/603 (the Windows case above) · `ai-persistence` 95 · `ai-durable-stream` 45 · adapters 261 · providers 214 including **Docker 37/37 under `REQUIRE_DOCKER=1`** and BusyBox conformance · E2E 390 passed / 0 failed · kiira 894/894 · publint 15/15 · sherif, knip, format and 17 typechecks clean.

Run directly per package rather than through `nx`, which is unreliable on this machine.

## Summary by CodeRabbit

* **New Features**
  * Added durable sandbox runs that survive disconnects, support replay and takeover, and enforce single-writer safety.
  * Added persistent run journals, deterministic replay alignment, cancellation tracking, snapshots, and detached-run reaping.
  * Added unified run statuses, structured errors, resumable stream improvements, and provider capability reporting.
* **Bug Fixes**
  * Improved process cleanup, Windows and Docker termination, stream handling, timeouts, teardown, and middleware error isolation.
* **Documentation**
  * Added guides covering durable runs, journals, takeover, persistence, reaping, and provider behavior.


tombeckenham:
## Cloudflare feedback

The CF path ("DO owns the run + clients tail a DO-backed event log") shouldn't be documented as a Cloudflare specialization — it's the **log-first tier of the portable protocol**, and most of it already exists: `StreamDurability` is the required half of the durability wiring, `ai-durable-stream` is a multi-host backend with cursor replay, and `align.ts` already reconciles journal → stored log on attach. The only genuinely CF-specific piece is the DO's supervised lifetime (driver outlives requests, `alarm()`); off CF, journal + takeover **is** the substitute for that, so it stays mandatory — just demoted to driver-recovery duty.

### Asks

**Docs**
- [ ] Name the two tiers explicitly: **journal-only** (zero infra; log durability = sandbox lifetime) vs **log-first** (clients only tail the log; reconnect never touches the sandbox; journal = warm reattach / pump-resume only). CF is the log-first tier with DO-backed implementations, not a parallel architecture.
- [ ] One-sentence precedence rule when both exist: **log wins for clients, journal wins for the pump's resume position** (what `align.ts` implements).
- [ ] CF non-goals: `/tmp` journal is not durable (`durableFilesystem: false`); refresh safety ≠ sleep survival; document the **`sleepAfter`** footgun (a detached-but-quiet run can sleep the container and look like durability failed). Prefer named sandbox (`threadId`) in CF examples.
- [ ] Keep the three layers separate in CF docs: run/events → DO log; workspace → re-bootstrap / `mountBucket`; artifacts → R2.

**Runtime / API**
- [ ] Ship non-CF log backends so log-first is real off Workers: a Node/SQLite or Postgres durable-streams server, or `StreamDurability` + `RunStore` + `LockStore` recipes for Redis/Postgres.
- [ ] Converge the CF package onto the portable seams (the `run-log.ts` header already marks its legacy vocabulary as deferred migration): DO log becomes a `StreamDurability` backend, coordinator a platform binding of the portable driver, `alarm()` a binding of the reaper. Ship the **DO `alarm()` reaper example**, not only generic cron.
- [ ] Keep **disconnect = detach tail; Stop = explicit cancel** on the CF path, and when `killableProcesses: false`, document what cancel actually does (destroy vs "marked cancelled, still running").

**Verification**
- [ ] CF journal/kill conformance should be **measured or named-skip**, never a silent pass.

### Keep as-is
Detach vs Stop, both-or-neither `runs`+`durability`, unforgeable exit sentinel, measured `killableProcesses`, non-driving reaper.

**Summary:** promote what CF proved — clients tail a log, the sandbox is never touched on reconnect, cancel is explicit — to the documented production tier of the one portable protocol, with the journal officially demoted to driver recovery whenever a real log backend is wired. Cost worth stating: log-first double-writes every chunk (journal + log), and journal-only must remain the zero-infra default.

docs/sandbox/journal.md
```diff
@@ -0,0 +1,476 @@
...
+```
+/tmp/tanstack-runs/<runId>.ndjson    every NDJSON event the agent emitted
+/tmp/tanstack-runs/<runId>.err       the agent's stderr, kept separate
+```
+
+Because there is no pipe, there is no reader whose disappearance can signal the
+agent. And because the journal is a file, a reader can start at byte 0 whenever
+it likes.
+
+## You have to ask for a journal
```
tombeckenham, line 36: I'm starting to think we should be a opinionated. Is there any reason why the user would not want to ask for a journal? Should it be on by default?

docs/sandbox/journal.md
```diff
@@ -0,0 +1,476 @@
...
+    middleware: [withSandbox(sandbox, { runs, durability: { adapter } })],
+  })
+  return toServerSentEventsResponse(stream, { durability: { adapter } })
+}
+```
+
+`@tanstack/ai-client` mints a fresh `runId` for every run and puts it in the
+AG-UI request body, which is what `chatParamsFromRequest` hands back. So the
+client half of this is nothing at all: `useChat` already sends a unique id per
+run, and reconnecting behaviour is unchanged.
```
tombeckenham, line 110: I totally miss this section. When would they use journals without using ai-client...

tombeckenham: We should be leading with. tanstack-ai already mints the runId, so don't worry about it

docs/sandbox/journal.md
```diff
@@ -0,0 +1,476 @@
...
+cannot be found, claimed, or reaped. There is no useful half.
+
+## Give every run an id you can recompute
+
+The journal path is derived from `runId` alone, so a `runId` you cannot
+reproduce is a journal nobody can find. Adapters fall back to a random internal
+id on a **non-durable** run; on a durable one there is no fallback at all —
+`chatStream` throws `DurableRunIdRequiredError`, deliberately, because an id no
+successor host can recompute produces a run that streams normally and is
+silently unrecoverable.
```
tombeckenham, line 61: This runId section makes it sound more complicated than it is. It reads like the user has to create runIds manually and store them somewhere.

docs/sandbox/journal.md
```diff
@@ -0,0 +1,476 @@
...
+})
+```
+
+## The exit sentinel carries a nonce
+
+The line a run's shell appends after the agent exits is not the bare
+`{"__exit":N}` it once was. It now carries a second field too:
+
+```json
+{"__exit":0,"__nonce":"3f9c1a7b..."}
```
tombeckenham, line 213: This is maybe getting into detail that probably should live in comments rather than user facing docs.

Me333-jjj:
Noticed the focus on durable, reconnectable sandbox runs. I specialize in maintaining agent state across connection drops.

tombeckenham (commented):
> Noticed the focus on durable, reconnectable sandbox runs. I specialize in maintaining agent state across connection drops. 

Hey @Me333-jjj what are you thoughts on the PR?

tombeckenham (approved):

PR #1019, opened 2026-07-30
fix(ai): re-open force-completed tool calls when more args arrive

### LINKED ISSUE #1017: Interleaved TEXT_MESSAGE_CONTENT permanently corrupts tool-call `input` with a partial-JSON parse
### TanStack AI version

v0.42.0

### Framework/Library version

React v	18.3.1

# Summary
A single TEXT_MESSAGE_CONTENT event arriving between two TOOL_CALL_ARGS events causes the tool call's input to be set from a lenient partial-JSON parse of the arguments received so far. The value is never corrected, because the subsequent authoritative TOOL_CALL_END is skipped for a call already marked input-complete.

The result is a tool-call part in state input-complete whose input contains silently truncated values, while arguments holds the correct, complete JSON string. Consumers have no way to detect this from state.

This reproduces in the framework agnostic core with no React involved.

# Reproduction
Run inside a project with @tanstack/ai installed:

```
import { StreamProcessor } from '@tanstack/ai/client';

const ARGS_HEAD = '{"templateIds":["mock-gsk-e';
const ARGS_TAIL = 'fimosfermin"]}';

const toolCallPartFrom = async (chunks) => {
  let messages = [];
  const processor = new StreamProcessor({
    events: { onMessagesChange: (next) => { messages = next; } },
  });
  await processor.process((async function* () { for (const c of chunks) yield c; })());
  return messages.flatMap((m) => m.parts ?? []).find((p) => p.type === 'tool-call');
};

const withoutInterleavedText = [
  { type: 'RUN_STARTED' },
  { type: 'TOOL_CALL_START', toolCallId: 't1', toolCallName: 'offerTemplates', messageId: 'm1' },
  { type: 'TOOL_CALL_ARGS', toolCallId: 't1', delta: ARGS_HEAD },
  { type: 'TOOL_CALL_ARGS', toolCallId: 't1', delta: ARGS_TAIL },
  { type: 'TOOL_CALL_END', toolCallId: 't1' },
  { type: 'RUN_FINISHED' },
];

const withInterleavedText = [
  { type: 'RUN_STARTED' },
  { type: 'TOOL_CALL_START', toolCallId: 't1', toolCallName: 'offerTemplates', messageId: 'm1' },
  { type: 'TOOL_CALL_ARGS', toolCallId: 't1', delta: ARGS_HEAD },
  // The only difference: one text delta between two TOOL_CALL_ARGS events.
  { type: 'TEXT_MESSAGE_CONTENT', messageId: 'm1', delta: 'Let me look those up. ' },
  { type: 'TOOL_CALL_ARGS', toolCallId: 't1', delta: ARGS_TAIL },
  { type: 'TOOL_CALL_END', toolCallId: 't1' },
  { type: 'RUN_FINISHED' },
];

for (const [label, chunks] of [
  ['without interleaved text', withoutInterleavedText],
  ['with interleaved text', withInterleavedText],
]) {
  const part = await toolCallPartFrom(chunks);
  console.log(`\n${label}`);
  console.log('  state    :', part?.state);
  console.log('  arguments:', part?.arguments);
  console.log('  input    :', JSON.stringify(part?.input));
}
```

outputs:

```
without interleaved text
  state    : input-complete
  arguments: {"templateIds":["mock-gsk-efimosfermin"]}
  input    : {"templateIds":["mock-gsk-efimosfermin"]}

with interleaved text
  state    : input-complete
  arguments: {"templateIds":["mock-gsk-efimosfermin"]}
  input    : {"templateIds":["mock-gsk-e"]}
```

input should equal {"templateIds":["mock-gsk-efimosfermin"]} in both cases, matching arguments. Failing that, input should be left undefined rather than populated with a truncated value, since ToolCallPart.input is already optional and its documentation states the raw arguments string is always available as a fallback.

# Root Cause (claude's best guess)
Three behaviours in activities/chat/stream/processor.js combine:

1. handleTextMessageContentEvent calls completeAllToolCallsForMessage(messageId) on every text delta, inferring that any in-flight tool call must be finished because text has started.
2. completeToolCall then populates the authoritative input via this.jsonParser.parse(toolCall.arguments). That is the lenient partial-json parser, which closes an unterminated string instead of rejecting it, so the truncated "mock-gsk-e becomes a plausible looking but wrong value.
3. handleToolCallEndEvent begins with if (existingToolCall && existingToolCall.state !== "input-complete"). The call is already input-complete from step 1, so the handler early-returns and cannot repair input.
Step 1 on its own would be a recoverable heuristic. Step 3 is what makes the corruption permanent.

Two pieces of in-repo documentation are contradicted by this behaviour. handleToolCallEndEvent is documented as the "authoritative signal that a tool call's input is finalized", and the ToolCallPart.input type comment says it is "Set from the parsed arguments once they are complete."

Any consumer reading part.input for a client side or human in the loop tool can act on truncated arguments while state reports input-complete. In our case an id arrived truncated and the UI issued a request for a resource that does not exist. Depending on the field, the failure mode may be a visible error or silently wrong content, for example a clipped question string or a dropped final array element.

# Claude's suggested fix
Any one of these resolves it; the first two seem most targeted:

1. In completeToolCall, populate input only when a strict JSON.parse(toolCall.arguments) succeeds, and leave it undefined otherwise.
2. Distinguish inferred completion from authoritative completion, and let TOOL_CALL_END overwrite an inferred one.
3. Stop force-completing tool calls from handleTextMessageContentEvent. The missed TOOL_CALL_END safety net is already covered by RUN_FINISHED and finalizeStream, which call completeAllToolCalls at stream termination.

# Other notes
In this case, I am using a client-side tool without an execution (empty `.client()), allowing a tool to "pause" the convo so the user can make a selection.

See above for reproduction notes!

_No response_

No, because I do not have time to dig into it

### Terms & Code of Conduct

Mohith26:
Fixes #1017, reported by @andrewmcgivery, matching the diagnosis in the issue exactly.

`handleTextMessageContentEvent` calls `completeAllToolCallsForMessage()` on every text delta, which finalizes in-flight tool calls with a lenient partial-JSON parse into `input`. `handleToolCallEndEvent` then early-skips calls already in `input-complete`, so when a TEXT_MESSAGE_CONTENT delta interleaves between two TOOL_CALL_ARGS deltas, the truncated parse becomes permanent even though `arguments` holds the full correct JSON.

The fix is 8 lines in `handleToolCallArgsEvent`: a further args delta on an `input-complete` call re-opens it to `input-streaming`, letting the authoritative TOOL_CALL_END (or the RUN_FINISHED safety net) re-finalize `input` from the complete arguments string. Self-correcting across interleaving orders; the existing inferred-completion heuristic is preserved.

Patch changeset included. Five regression tests (END-finalized ordering, END-input-canonical, RUN_FINISHED safety net, multiple parallel tool calls, the internal re-open transition) were written fail-first and all fail without the fix, reproducing the issue's exact truncated output. Package suite: 1291 passed, zero new failures; oxlint and tsc clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed tool-call inputs being truncated when text content appears between streamed argument updates.
  * Ensured completed tool calls use the full, authoritative arguments and input.
  * Improved recovery when tool-call completion events are missing.
  * Corrected handling for multiple tool calls processed in parallel.

* **Tests**
  * Added regression coverage for interleaved text and tool-call streaming scenarios.

### DIFF
--- a/.changeset/interleaved-text-tool-input.md
+++ b/.changeset/interleaved-text-tool-input.md
@@ -1,0 +1,13 @@
+---
+'@tanstack/ai': patch
+---
+
+Fix permanent tool-call `input` corruption when a `TEXT_MESSAGE_CONTENT` event
+arrives between two `TOOL_CALL_ARGS` events. Interleaved text force-completes
+in-flight tool calls with a lenient partial-JSON parse, and the later
+authoritative `TOOL_CALL_END` was skipped because the call was already
+`input-complete`, leaving silently truncated values in `input` while
+`arguments` held the correct JSON. The processor now re-opens a
+force-completed call when further args deltas arrive, so `TOOL_CALL_END` (or
+the `RUN_FINISHED` safety net) re-finalizes `input` from the complete
+arguments string.
--- a/packages/ai/src/activities/chat/stream/processor.ts
+++ b/packages/ai/src/activities/chat/stream/processor.ts
@@ -1347,6 +1347,14 @@ export class StreamProcessor {
     // Update state
     if (wasAwaitingInput && chunk.delta) {
       existingToolCall.state = 'input-streaming'
+    } else if (existingToolCall.state === 'input-complete' && chunk.delta) {
+      // The call was force-completed by an inferred completion (an interleaved
+      // TEXT_MESSAGE_CONTENT force-completes in-flight tool calls). A further
+      // args delta proves that inference was premature: re-open the call so
+      // the authoritative TOOL_CALL_END (or the RUN_FINISHED safety net)
+      // re-finalizes `input` from the complete arguments string instead of
+      // leaving the partial-JSON parse in place. See issue #1017.
+      existingToolCall.state = 'input-streaming'
     }
 
     // Try to parse the updated arguments
--- a/packages/ai/tests/stream-processor.test.ts
+++ b/packages/ai/tests/stream-processor.test.ts
@@ -773,6 +773,146 @@ describe('StreamProcessor', () => {
       expect((textParts[0] as any).content).toBe('Before')
       expect((textParts[1] as any).content).toBe('After')
     })
+
+    // Regression tests for https://github.com/TanStack/ai/issues/1017 —
+    // a TEXT_MESSAGE_CONTENT delta between two TOOL_CALL_ARGS deltas
+    // force-completes the in-flight tool call with a lenient partial-JSON
+    // parse, and the later authoritative TOOL_CALL_END was skipped because
+    // the call was already input-complete, permanently corrupting `input`.
+    describe('text delta interleaved between TOOL_CALL_ARGS deltas (#1017)', () => {
+      const ARGS_HEAD = '{"templateIds":["mock-gsk-e'
+      const ARGS_TAIL = 'fimosfermin"]}'
+      const FULL_ARGS = ARGS_HEAD + ARGS_TAIL
+
+      it('re-opens the inferred-complete call so state returns to input-streaming', () => {
+        const processor = new StreamProcessor()
+        processor.prepareAssistantMessage()
+
+        processor.processChunk(ev.runStarted())
+        processor.processChunk(ev.toolStart('tc-1', 'offerTemplates'))
+        processor.processChunk(ev.toolArgs('tc-1', ARGS_HEAD))
+        // Interleaved text force-completes the in-flight call (inferred).
+        processor.processChunk(ev.textContent('Let me look those up. '))
+        expect(processor.getState().toolCalls.get('tc-1')?.state).toBe(
+          'input-complete',
+        )
+
+        // A further args delta proves the inference was premature.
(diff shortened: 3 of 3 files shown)


tombeckenham:

tombeckenham:

tombeckenham:

AlemTuzlak:
This PR and #1183 target the same bug as #1020 (issue #1017). #1020 now has the smaller fix plus tests for this PR's cases (END.input, RUN_FINISHED safety net, parallel calls, streaming-through-text). See https://github.com/TanStack/ai/issues/1017#issuecomment-5369156432 and https://github.com/TanStack/ai/pull/1020. This can close when #1020 merges.

AlemTuzlak:
Closing as a duplicate of #1020.

#1020 stops text events from completing in-flight tool calls, so the call stays input-streaming and TOOL_CALL_END still runs. The reopen-on-later-args path is not needed.

Issue: #1017

PR #1021, opened 2026-07-30
fix(ai-anthropic): document metadata

jan-kubica:
## Problem

The document adapter spreads `part.metadata` into Anthropic's request. This forwards `mediaType`, which is declared by `AnthropicDocumentMetadata` but invalid on Anthropic's document block, as well as application metadata such as `filename` or `placeholder`. Anthropic rejects these as extra inputs.

## Fix

Serialize only supported document fields. Preserve `cache_control`, `citations`, `context`, and `title`, and use `filename` as a fallback title.

Includes a request-shape regression test and patch changeset.

Tests: CI

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved document and image metadata serialization for Anthropic requests.
  * Document titles now use the specified title, or fall back to the filename.
  * Supported metadata, including caching, citations, and context, is forwarded correctly while unsupported fields are excluded.

* **Documentation**
  * Clarified filename fallback behavior and deprecated legacy MIME-type metadata fields.

### DIFF
--- a/.changeset/tidy-pdfs.md
+++ b/.changeset/tidy-pdfs.md
@@ -2,4 +2,4 @@
 '@tanstack/ai-anthropic': patch
 ---
 
-Fix document metadata serialization.
+Serialize only Anthropic-supported fields from document and image metadata; use `filename` as a fallback document title.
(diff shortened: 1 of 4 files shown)


AlemTuzlak:
why is this in draft?

jan-kubica:

tombeckenham:

tombeckenham (commented):
Thanks!

jan-kubica:
@tombeckenham thank you for review & merging

PR #1090, opened 2026-08-12
fix(ai-persistence): preserve cancelled tool resumes

### LINKED ISSUE #1088: `withPersistence` does not apply cancelled client-tool resume entries
### TanStack AI version

0.44.0

### Framework/Library version

@tanstack/ai-client: 0.23.2 | @tanstack/ai-persistence: 0.1.2 | @tanstack/ai-svelte: 0.16.2

#1086 covers the client-side gap: recognizing that server hydration installed hidden client-tool interrupts so an application can call `cancelInterrupts()`. That issue stops once the cancelled resume batch is submitted.

The server-authoritative resume path then drops those cancellations.

1. The client submits cancelled resume entries for the pending client-tool interrupts.
2. `withPersistence` `onConfig` translates persisted pending interrupts into `resumeToolState` and clears `config.resume`, so the engine skips ephemeral reconstruction [middleware.ts#L1475-L1486](https://github.com/TanStack/ai/blob/79d8812742b1c0358a83ab1cbe7ae724c6a0d8e0/packages/ai-persistence/src/middleware.ts#L1475-L1486).
3. `resumeToolStateFromPending()` handles approvals and **resolved** client-tool results, but never maps a cancelled client-tool entry into `cancelledToolCallIds` [middleware.ts#L393-L423](https://github.com/TanStack/ai/blob/79d8812742b1c0358a83ab1cbe7ae724c6a0d8e0/packages/ai-persistence/src/middleware.ts#L393-L423).

The ephemeral path already does the right thing: cancelled bindings are added to `cancelledToolCallIds` in [interrupt-resume.ts#L631-L633](https://github.com/TanStack/ai/blob/79d8812742b1c0358a83ab1cbe7ae724c6a0d8e0/packages/ai/src/interrupt-resume.ts#L631-L633). The tool executor then emits an `output-error` result (`Tool execution cancelled`) instead of re-interrupting [tool-calls.ts#L806-L813](https://github.com/TanStack/ai/blob/79d8812742b1c0358a83ab1cbe7ae724c6a0d8e0/packages/ai/src/activities/chat/tools/tool-calls.ts#L806-L813).

Because persistence clears `config.resume` and supplies an incomplete `resumeToolState`, the engine still sees the stored tool call as unhandled and emits another client-tool interrupt instead of an output-error. Cancelling a hydrated orphaned client tool therefore cannot complete the turn under `withPersistence`.

### Expected behavior

`resumeToolStateFromPending()` should mirror the ephemeral resume translator:

- cancelled client-tool (and approval) entries populate `cancelledToolCallIds`
- resolved client-tool entries continue to populate `clientToolResults`
- a resume batch that is only cancellations still produces a non-empty `resumeToolState` so the engine does not fall through to a fresh interrupt

https://stackblitz.com/edit/vitejs-vite-ewvk7m7x?file=package.json,index.html,src%2Fmain.ts

_No response_

None

### Terms & Code of Conduct

mikemikimike:
## Summary
- Preserve cancelled approval and client-tool IDs in persisted resume state.
- Keep cancellation-only resumes non-empty so the chat engine consumes them.
- Add regression coverage for both cancellation paths.

Fixes #1088

## Validation
- vitest run packages/ai-persistence/tests: 172 passed
- tsc -p packages/ai-persistence/tsconfig.json --noEmit: passed
- oxlint packages/ai-persistence/src --type-aware: passed
- vite build in packages/ai-persistence: passed
- publint --strict: passed
- git diff --check: passed

The repository Playwright/Chromium install was blocked during Windows extraction, so the E2E suite was not run.

## Summary by CodeRabbit

* **Bug Fixes**
  * Cancelled tool interruptions are now preserved when conversations are resumed.
  * Cancelled approval and client-tool calls are accurately tracked during resume.
  * Cancelled client-tool operations now return a cancellation error, complete the run, and clear pending interruptions.
  * Cancellation-only resume requests no longer trigger another client-tool interruption.
  * Approval-denial behavior continues to work as expected.

### DIFF
--- a/packages/ai-persistence/tests/interrupts.test.ts
+++ b/packages/ai-persistence/tests/interrupts.test.ts
@@ -87,6 +87,39 @@ const runFinished = (runId = 'r1'): StreamChunk => ({
   timestamp: 1,
 })
 
+const toolCallFinished = (runId = 'r1'): StreamChunk => ({
+  type: EventType.RUN_FINISHED,
+  runId,
+  threadId: 't1',
+  finishReason: 'tool_calls',
+  timestamp: 1,
+})
+
+const toolCallChunks = () => [
+  runStarted(),
+  toolStart(),
+  toolArgs(),
+  toolCallFinished(),
+]
+
+async function persistClientToolTurn(
+  persistence: ReturnType<typeof memoryPersistence>,
+  tools: Array<Tool>,
+) {
+  const first = mockAdapter([toolCallChunks()])
+  await collect(
+    chat({
+      adapter: first.adapter,
+      messages: [{ role: 'user', content: 'hi' }],
+      tools,
+      runId: 'r1',
+      threadId: 't1',
+      middleware: [withPersistence(persistence)],
+    }) as AsyncIterable<StreamChunk>,
+  )
+  return first
+}
+
 const clientTool = (name: string): Tool => ({
   name,
   description: `${name} client tool`,
@@ -144,20 +177,7 @@ describe('interrupt persistence', () => {
   it('does not persist duplicate records before terminal interrupt outcome', async () => {
     const persistence = memoryPersistence()
     const create = vi.spyOn(persistence.stores.interrupts!, 'create')
-    const { adapter } = mockAdapter([
-      [
-        runStarted(),
-        toolStart(),
-        toolArgs(),
-        {
-          type: EventType.RUN_FINISHED,
-          runId: 'r1',
-          threadId: 't1',
-          finishReason: 'tool_calls',
-          timestamp: 1,
-        },
-      ],
-    ])
+    const { adapter } = mockAdapter([toolCallChunks()])
 
     await collect(
       chat({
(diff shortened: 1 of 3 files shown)


packages/ai-persistence/tests/interrupts.test.ts
```diff
@@ -829,6 +844,7 @@ describe('interrupt persistence', () => {
       run.calls[0] as { clientToolResults?: ReadonlyMap<string, unknown> }
     ).clientToolResults
```
AlemTuzlak, line 904: **[suggestion]** The new assertions only prove the translator wrote `cancelledToolCallIds`. Both cancel tests still use empty `messages`, no `tools`, and no stored thread with a pending assistant tool call. The engine therefore never reaches `checkForPendingToolCalls()` / `executeToolCalls()`, so these tests cannot fail if a cancelled hydrated client tool still re-interrupts. The file already has the right shape for that path in `applies persisted approval and client-tool resume decisions with empty client messages` (persist tool-call turn, resume with empty client messages, assert chunks).

**Suggestion:** Add a regression that mirrors that hydrate path for `status: 'cancelled'`: persist a client-tool interrupt plus the assistant tool-call message, resume with empty `messages` and the same tools, and assert a `TOOL_CALL_RESULT` of `Tool execution cancelled` plus no new `client_tool_*` interrupt. Keep the current `onConfig` Set check if you want, but do not treat it as coverage for #1088.

AlemTuzlak (approved):

PR #1103, opened 2026-08-14
fix(ai-gemini): stop dropping modelOptions on the native image path

L4Ph:
## 🎯 Changes

`GeminiImageModelProviderOptionsByName` maps **every** Gemini image model to the Imagen-shaped `GeminiImageProviderOptions`:

```ts
export type GeminiImageModelProviderOptionsByName = {
  [K in GeminiImageModels]: GeminiImageProviderOptions
}
```

But the four Gemini-native image models are served by `generateContent`, not `generateImages`, so they take `GenerateContentConfig` — a different shape. The consequences:

- **Type level.** `modelOptions: { safetySettings }` / `{ thinkingConfig }` / `{ imageConfig }` is a compile error on `gemini-3.1-flash-image-preview`, `gemini-3.1-flash-lite-image`, `gemini-3-pro-image-preview` and `gemini-2.5-flash-image`, even though the API accepts all three.
- **Runtime.** `generateWithGeminiApi` forwarded only `modelOptions.seed` and dropped everything else. That was the correct response to a wrong type, and the code said so:

  > `GeminiImageProviderOptions` is Imagen-shaped — most fields … are only valid on `GenerateImagesConfig` and would be rejected by the Gemini-native `generateContent` path. Pick only the fields that are valid on `GenerateContentConfig` instead of spreading the whole options object.

So there is currently no way to set safety thresholds or a thinking budget on a Nano Banana call.

### The fix

Make the provider-options map follow the native/Imagen split that **the two sibling maps in the same file already use**:

```ts
export type GeminiImageModelSizeByName = {
  [K in GeminiNativeImageModels]: GeminiNativeImageSize
} & { [K in Exclude<GeminiImageModels, GeminiNativeImageModels>]: GeminiImageSize }

export type GeminiImageModelInputModalitiesByName = {
  [K in GeminiNativeImageModels]: readonly ['image']
} & { [K in Exclude<GeminiImageModels, GeminiNativeImageModels>]: readonly [] }
```

This PR makes the provider-options map the third one to do it. Native models get a new `GeminiNativeImageProviderOptions` (`seed`, `safetySettings`, `thinkingConfig`, `imageConfig`, `systemInstruction`); Imagen models keep `GeminiImageProviderOptions`.

Notes on the details:

- **Both** paths now pick their config fields **by name** rather than spreading `modelOptions` wholesale, so neither endpoint can receive a field shaped for the other. (Spreading was previously safe on the Imagen side only because the type made the native fields unreachable; widening the base type removed that accident, so `buildImagenConfig` was hardened to match.)
- `responseModalities` is deliberately **absent** from the new type. The adapter always requests `['TEXT', 'IMAGE']`, and the existing `does not let modelOptions override responseModalities` test still guards it.
- `modelOptions.imageConfig` merges **over** the `imageConfig` derived from the portable `size` option, per field — passing only `imageConfig.imageSize` keeps the `aspectRatio` that `size` implied.
- Runtime routing moves off a `model.startsWith('gemini-')` heuristic onto membership in `GEMINI_NATIVE_IMAGE_MODELS`, so the route and the type-level split cannot drift. An unknown id now reaches the Imagen endpoint and fails there instead of taking the native path with Imagen-shaped option types — the same class of mismatch this PR exists to remove.
- `HarmCategory` / `HarmBlockThreshold` are re-exported so `safetySettings` is writable without adding `@google/genai` to your own dependencies.

**BREAKING (types only)**, spelled out in the changeset: the 14 Imagen-only fields no longer compile on the four native models. They previously type-checked but were already discarded at runtime, so no request behaviour changes — the compiler now reports what was already happening.

## Test plan

- `packages/ai-gemini`: `pnpm test:lib` → 16 files / 275 tests pass; `pnpm test:types` → clean.
- Root `pnpm run test:pr` → green (sherif, knip, docs, kiira, maintainer, oxlint, lib, types, build across 77 projects; `scan-dangling-dts` clean).
- New runtime tests assert the config object actually handed to `generateContent` / `generateImages`: `safetySettings`, `thinkingConfig` and `systemInstruction` forwarded; `imageConfig` merge asserting **both** the overridden field and the preserved one; and an Imagen guard that passes native fields through a cast and asserts they never reach `generateImages`. That guard fails if the `buildImagenConfig` hardening is reverted.
- New `tests/image-per-model-type-safety.test.ts` follows the existing `chat-per-model-type-safety.test.ts` pattern. Because `tsc` errors on an unused `@ts-expect-error`, a green typecheck is positive proof each negative case really is rejected.
- Docs snippets are checked by kiira, so the new native `modelOptions` example is compiled call-site proof of the fix.

**E2E coverage** — `testing/e2e/tests/gemini-native-image-wire.spec.ts` + `src/routes/api.gemini-native-image-wire.ts`, backed by a new `geminiNativeImageMount()` in `global-setup.ts`. The spec drives a real `generateImage()` call with `modelOptions: { safetySettings, thinkingConfig }` and the mount **rejects the request with a 400 if either field is missing from the raw body**, so a dropped option fails the spec instead of passing silently. Verified revert-proof: with `packages/ai-gemini/src` reverted to this branch's parent, the spec fails with `Missing top-level safetySettings (modelOptions.safetySettings did not reach the wire)`.

An earlier revision of this description claimed E2E was not achievable here, citing the Gemini exclusions in `testing/e2e/src/lib/feature-support.ts`. That was wrong, and worth correcting for anyone who reads those comments the same way: the `image-gen` exclusion is about **Imagen's `:predict`** (which aimock does in fact mock now, via `handleImages`), not the native path — native image models always go through `generateContent`. What genuinely does not work is aimock's *fixture* machinery: `handleGemini` never imports `isImageResponse`, so an image-shaped fixture falls through every branch and 500s; and its `/journal` stores a lossy OpenAI-shaped translation that strips `safetySettings` and `generationConfig` before recording. Both are sidestepped by `LLMock.mount()`, the repo's own documented escape hatch — already used 10+ times in `global-setup.ts`, including `geminiTTSMount()` for the identical `{model}:generateContent` → `inlineData` shape. No API key is needed and no fixture is recorded.

## ✅ Checklist

## 🚀 Release Impact

The changeset is `minor` with an explicit `**BREAKING (types only):**` paragraph. `packages/ai-gemini/CHANGELOG.md` 0.15.0 shipped a `**BREAKING:**` under `### Minor Changes`, so this follows precedent — but CONTRIBUTING says major needs maintainer coordination, so flag it if you'd rather cut this as major.

## Summary by CodeRabbit

* **New Features**
  * Added model-specific configuration options for Gemini native image models and Imagen models.
  * Added support for thinking, safety, system instructions, image settings, and seed configuration.
  * Added Gemini safety enums and image configuration types to public SDK exports.
  * Improved image-model routing and configuration handling.

* **Documentation**
  * Updated Gemini image-generation guidance with model-specific examples and configuration details.
  * Clarified native model settings, image configuration merging, and response modality behavior.


docs/adapters/gemini.md
```diff
@@ -414,7 +414,7 @@ The Gemini adapter supports two types of image generation:
 - **Gemini native image models** (NanoBanana) — Use the `generateContent` API with models like `gemini-3.1-flash-image-preview`. These support extended resolution tiers (1K, 2K, 4K) and aspect ratio control.
 - **Imagen models** — Use the `generateImages` API with models like `imagen-4.0-generate-001`. These are dedicated image generation models with WIDTHxHEIGHT sizing.
 
-The adapter automatically routes to the correct API based on the model name — models starting with `gemini-` use `generateContent`, while `imagen-` models use `generateImages`.
+The adapter automatically routes to the correct API based on membership in the known list of Gemini native models (`GEMINI_NATIVE_IMAGE_MODELS`) — the Gemini native models listed above use `generateContent`, while Imagen models, and any model id this package doesn't know about, use `generateImages`.
```
AlemTuzlak: **[suggestion]** The docs name `GEMINI_NATIVE_IMAGE_MODELS` as the routing list, but that constant (and `isGeminiNativeImageModel`) is not exported from `@tanstack/ai-gemini`. A reader who tries to import the name the docs use will get a compile error. The video adapter already exports the same kind of runtime helper (`GEMINI_VIDEO_DURATIONS`, `isInteractionsVideoModel`).

**Suggestion:** Either export `GEMINI_NATIVE_IMAGE_MODELS` and `isGeminiNativeImageModel` from `packages/ai-gemini/src/index.ts`, or drop the private name from the docs and say "the known Gemini native models listed above."

packages/ai-gemini/src/image/image-provider-options.ts
```diff
@@ -121,11 +133,77 @@ export interface GeminiImageProviderOptions {
...
+   * Use to raise or disable the thinking budget on models that support it
+   */
+  thinkingConfig?: ThinkingConfig
+
+  /**
+   * Native image output controls (aspect ratio, resolution tier, …)
+   * Merged over the values derived from the portable `size` option, so
+   * fields set here win per field while the rest of `size` is preserved
+   */
+  imageConfig?: ImageConfig
```
AlemTuzlak: **[suggestion]** `imageConfig` is the full SDK `ImageConfig`. On the Gemini Developer API (this adapter's default: `apiKey`, no `vertexai: true`), `@google/genai` throws if `personGeneration`, `prominentPeople`, `outputMimeType`, `outputCompressionQuality`, or `imageOutputOptions` is set. That is the same class of throw they avoided by omitting `labels` from the native type. After this PR, autocomplete on `imageConfig` offers those fields, so a caller moving off the old top-level Imagen `personGeneration` can type-check and then crash.

**Suggestion:** Narrow the public native `imageConfig` to `{ aspectRatio?: string; imageSize?: string }`, or keep the SDK type and say in the JSDoc / image-generation docs that only `aspectRatio` and `imageSize` work on the Gemini Developer API.

PR #1104, opened 2026-08-14
fix(ai-gemini): add GA image model ids and type sizes per model

L4Ph:
## 🎯 Changes

Two related problems in how `packages/ai-gemini` types image models.

### 1. The GA model ids are missing; the ids on offer are shut down

`GEMINI_IMAGE_MODELS` carries `gemini-3.1-flash-image-preview` and `gemini-3-pro-image-preview`. Both were **shut down on 2026-06-25**, and their GA replacements `gemini-3.1-flash-image` / `gemini-3-pro-image` appear nowhere in the package — so for two of the four Nano Banana models, the only ids this adapter offers are 404s. (`https://ai.google.dev/gemini-api/docs/models/gemini-3-pro-image-preview` returns HTTP 404.)

> `gemini-3.1-flash-image-preview` | February 26, 2026 | **June 25, 2026** | `gemini-3.1-flash-image`
> `gemini-3-pro-image-preview` | November 20, 2025 | **June 25, 2026** | `gemini-3-pro-image`
> — [deprecations](https://ai.google.dev/gemini-api/docs/deprecations)

This PR adds the GA ids and keeps the preview ids as `@deprecated` aliases carrying their shutdown dates, so **nothing is removed from the union** and existing code keeps compiling. `gemini-2.5-flash-image` is tagged too (shutdown 2026-10-02) but left fully functional. Docs, agent skills, examples and the stream-processor panel are moved onto the GA ids — several of them were recommending dead ids to anyone copying them.

Whether to actually *remove* the dead ids is a maintainer call; happy to follow up with that as a major if you want it.

### 2. One flat size union for four models with different capabilities

```ts
export type GeminiImageModelSizeByName = {
  [K in GeminiNativeImageModels]: GeminiNativeImageSize   // `${8 ratios}_${'1K'|'2K'|'4K'}`
} & { … }
```

That single union is simultaneously **too narrow** and **too wide**. Every Google source lists `4:5` and `5:4` for all four models, and neither is in it; meanwhile it offers `2K`/`4K` on a model that only does `1K`, and `21:9_4K` on a model that documents no resolutions at all.

Typed per model instead, from the documented matrix:

| model | aspect ratios | image sizes |
|---|---|---|
| `gemini-3.1-flash-image` | 14 | `512` `1K` `2K` `4K` |
| `gemini-3.1-flash-lite-image` | 14 (see note) | `1K` only |
| `gemini-3-pro-image` | 10 | `1K` `2K` `4K` |
| `gemini-2.5-flash-image` | 10 | none — bare aspect ratio |

- 14-ratio set: `1:1 2:3 3:2 3:4 4:3 4:5 5:4 9:16 16:9 21:9 1:4 4:1 1:8 8:1`. 10-ratio set: the first ten.
- **`9:21` is deliberately absent.** Cloud lists it for the Gemini 3 image models; it appears on no Gemini API surface — not the guide tables, not either `ImageConfig` in the REST reference, not the `@google/genai` typings. Including it would accept a call the Gemini API rejects.
- Google documents **no `image_size`** for `gemini-2.5-flash-image` — the `512/1K/2K/4K` ladder is scoped to "Gemini 3 image models", the guide says it "generates images at 1024px resolution", and the per-model table has a single unlabeled resolution column. So its sizes are bare aspect ratios (`'16:9'`) and the adapter sends no `imageSize` for it. `parseNativeImageSize` now accepts a ratio with no `_resolution` suffix.
- Resolution tokens are case-sensitive literals and the smallest is the string `'512'` — not `'512px'`, not `'0.5K'`. Google: *"You must use an uppercase 'K' … Lowercase parameters (e.g., 1k) will be rejected."*

### Sourcing

Every value was taken from `ai.google.dev` (the guide's per-model tables, the `GenerationConfig.imageConfig` field reference, the deprecations page) and cross-checked against the Cloud model pages, with divergences resolved in favour of the Gemini API — that is the surface `@google/genai`'s `config.imageConfig` targets.

Worth flagging: **the `@google/genai` 2.10.0 public `ImageConfig.aspectRatio` doc comment is stale** — it lists 8 values, omitting `4:5` and `5:4`, and appears to be a copy of the deprecated Interactions-API `ImageConfig`. Its internal `ImageConfigAspectRatio` union has the correct 14. I did not use the doc comment as a source.

**One soft cell, disclosed in the JSDoc, the docs tables and the changeset:** for `gemini-3.1-flash-lite-image`, the four extreme ratios `1:4`/`4:1`/`1:8`/`8:1` are enumerated only on the Cloud model page. `ai.google.dev` asserts the *count* ("a discrete set of 14 aspect ratios") but its only enumeration for that model is a 10-item bullet prefixed "New aspect ratios" — read here as a what's-new list, since reading it as exhaustive makes the page contradict its own count. The guide has per-model ratio tables for the other three models and none for Flash Lite. If you'd rather take zero over-acceptance risk, the conservative fallback is the 10-ratio set; say so and I'll narrow it.

## Test plan

- `packages/ai-gemini`: `pnpm test:lib` → 272 tests pass; `pnpm test:types` → clean.
- Root `pnpm run test:pr` → green (77 projects; kiira 1010 snippets; `scan-dangling-dts` clean).
- New runtime tests: `parseNativeImageSize` on a bare ratio and on `'4:5_2K'` / `'1:8_512'`; a `gemini-2.5-flash-image` call asserting `'imageSize' in config.imageConfig === false`; GA ids routing through `generateContent`.
- New `tests/image-per-model-type-safety.test.ts` proves the per-model gating with `@ts-expect-error` — `'1:8_1K'` rejected on pro, `'16:9_4K'` rejected on flash-lite, `'16:9_2K'` rejected on 2.5-flash, `'4:5_2K'` accepted on pro. `tsc` errors on an unused `@ts-expect-error`, so a green typecheck proves each negative really is rejected.
- **Revert-proofing:** reverting `packages/ai-gemini/src` to HEAD and re-running both gates produces 2 `test:lib` failures and 35 `test:types` errors, including `TS2578: Unused '@ts-expect-error'` on exactly the four negatives the new types introduce.

**E2E coverage** — `testing/e2e/tests/gemini-image-ga-models.spec.ts` + `src/routes/api.gemini-image-ga-models.ts`, backed by a new `geminiNativeImageMount()` in `global-setup.ts`. Two tests: a GA id (`gemini-3.1-flash-image`) is present in the exported model array and actually routes a `generateImage()` call through to an image; and `gemini-2.5-flash-image` sends `imageConfig.aspectRatio` with **no** `imageSize`. The mount validates the raw request body and pins the exact values the route sends (`aspectRatio: '16:9'`, and `imageSize: '2K'` only for the flash-image call), so a regression in how `parseNativeImageSize` splits `'16:9_2K'` fails the spec rather than slipping through a presence-only check. Verified revert-proof against the branch's parent: the GA id drops out of the exported array, and the bare ratio `'16:9'` stops parsing so no `imageConfig` reaches the wire.

An earlier revision of this description claimed E2E was not achievable here, citing the Gemini exclusions in `testing/e2e/src/lib/feature-support.ts`. That was wrong, and worth correcting for anyone who reads those comments the same way: the `image-gen` exclusion is about **Imagen's `:predict`** (which aimock does in fact mock now, via `handleImages`), not the native path — native image models always go through `generateContent`. What genuinely does not work is aimock's *fixture* machinery: `handleGemini` never imports `isImageResponse`, so an image-shaped fixture falls through every branch and 500s; and its `/journal` stores a lossy OpenAI-shaped translation that strips `generationConfig` before recording. Both are sidestepped by `LLMock.mount()`, the repo's own documented escape hatch — already used 10+ times in `global-setup.ts`, including `geminiTTSMount()` for the identical `{model}:generateContent` → `inlineData` shape. No API key is needed and no fixture is recorded.

## ✅ Checklist

## 🚀 Release Impact

`minor`, with an explicit `**BREAKING (types only):**` paragraph — the narrowing rejects ratio/resolution combinations the models never supported. Following the 0.15.0 precedent of shipping `**BREAKING:**` under `### Minor Changes`; flag it if you'd rather cut this as major.

---

Independent of #1103 (native `modelOptions`), but both touch `GeminiNativeImageModels` in `image-provider-options.ts`, so whichever lands second will need a trivial rebase.

## Summary by CodeRabbit

- **New Features**
  - Added support for current Gemini image-generation models, including Gemini 3.1 Flash, Gemini 3 Pro, and Flash Lite.
  - Added model-specific image sizes, aspect ratios, and resolution options, including bare ratios where supported.
  - Exposed image configuration types for improved validation and editor support.

- **Documentation**
  - Updated guides, examples, and model listings with current identifiers, capabilities, availability, and limits.
  - Documented deprecated preview identifiers, shutdown dates, and potential runtime errors.


packages/ai-gemini/src/model-meta.ts
```diff
@@ -146,7 +185,47 @@ const GEMINI_3_PRO_IMAGE = {
...
 
+/**
+ * Gemini 3.1 Flash Image ("Nano Banana 2") — GA. The only native image model
+ * that accepts the four extreme banner ratios (1:4, 4:1, 1:8, 8:1) and the
+ * 512 (0.5K) resolution tier.
+ * @see https://ai.google.dev/gemini-api/docs/models/gemini-3.1-flash-image
+ */
 const GEMINI_3_1_FLASH_IMAGE = {
+  name: 'gemini-3.1-flash-image',
+  max_input_tokens: 65_536,
```
AlemTuzlak: **[suggestion]** The new GA `gemini-3.1-flash-image` const copies the old preview numbers. Google's current model page lists input 131,072 and output 32,768. Pricing is $0.50 input / $3 text (and $60 image) per 1M tokens, not 0.25 / 1.5. The page also says structured output is not supported, but `capabilities` still includes `structured_output`. These fields are not exported and do not change `generateImage()` types or the request, so this is not a runtime break. It is still wrong data on a new public model id.

**Suggestion:** Set `max_input_tokens` to 131_072 and `max_output_tokens` to 32_768. Set pricing to the current paid rates. Drop `structured_output` (and `GeminiStructuredOutputOptions`) from this const. Leave the dead `-preview` const as-is if you only want to fix the live id.

packages/ai-gemini/src/model-meta.ts
```diff
@@ -146,7 +185,47 @@ const GEMINI_3_PRO_IMAGE = {
...
+  GeminiToolConfigOptions &
+    GeminiSafetyOptions &
+    GeminiCommonConfigOptions &
+    GeminiCachedContentOptions &
+    GeminiStructuredOutputOptions &
+    GeminiThinkingOptions
+>
+
+/**
+ * @deprecated `gemini-3.1-flash-image-preview` was shut down on 2026-06-25.
```
AlemTuzlak: **[suggestion]** `@deprecated` sits on the private preview consts. `GeminiImageModels` is `(typeof GEMINI_IMAGE_MODELS)[number]`, so JSDoc is dropped. `geminiImage('gemini-3.1-flash-image-preview')` still has no strikethrough. The changeset already says this. The tags therefore do not do the job the PR says they do.

**Suggestion:** Write the preview members as an explicit union with `@deprecated` on each alias (or export deprecated string consts and use those in the array). Then a call to a dead id shows a warning in the editor.

docs/advanced/runtime-adapter-switching.md
```diff
@@ -125,25 +125,28 @@ import { geminiImage } from '@tanstack/ai-gemini'
...
     prompt: 'A beautiful sunset over mountains',
-    size: '1024x1024',
+    size,
   })
 
   return Response.json(result)
 }
 ```
 
+`size` is provider-specific, which is why it can't be a single literal shared across branches: Gemini native image models take a `'<aspectRatio>_<tier>'` string (e.g. `'16:9_4K'`), while OpenAI and Imagen models take pixel dimensions (e.g. `'1024x1024'`).
```
AlemTuzlak: **[suggestion]** The new sentence says every Gemini native image model takes `'<aspectRatio>_<tier>'`. That is true for `gemini-3.1-flash-image` in the sample. It is false for `gemini-2.5-flash-image`, which now needs a bare ratio such as `'16:9'`.

**Suggestion:** Say the 3.x native models use `'<aspectRatio>_<tier>'`, and that `gemini-2.5-flash-image` uses a bare ratio with no suffix.

testing/e2e/global-setup.ts
```diff
@@ -505,6 +519,153 @@ function geminiBatchEmbedMount(): Mountable {
...
+  res.end(
+    JSON.stringify({
+      error: { code: 400, message, status: 'INVALID_ARGUMENT' },
+    }),
+  )
+  return true
+}
+
+/**
+ * Mounts Gemini native image generation (`POST /v1beta/models/{model}:generateContent`)
```
AlemTuzlak: **[suggestion]** The new mount JSDoc (and the Flash Lite size JSDoc in `image-provider-options.ts`) retell the change, aimock internals, and PR #1104. Comments here should say why the mount exists, not the full research path.

**Suggestion:** Keep a short why: aimock has no `generateContent` image branch; this mount checks `generationConfig.imageConfig` for the two models the spec calls. Move the long history to the changeset.

PR #1119, opened 2026-08-16
fix: resolve local-process wait after child close

### LINKED ISSUE #1112: ai-sandbox-local-process: wait() hangs when the child closes before wait() is called
### Which project does this relate to?

AI

### Describe the bug

`@tanstack/ai-sandbox-local-process` (0.2.1): `spawnProcess`'s `wait()` subscribes to the child's `close` event lazily, inside the `wait()` call:

```js
wait: () => new Promise((resolve, reject) => {
  child.on("error", reject);
  child.on("close", (code) => resolve(code ?? 0));
}),
```

`close` fires exactly once. A caller that drains stdout before calling `wait()` — which is what `spawnNdjson` does — can reach `wait()` after the child has already fully closed. The stdout iterator buffers lines in JS, so with a slow per-chunk consumer (for example a delivery layer with `batch: 1` that awaits one HTTP POST per chunk) the child exits and emits `close` while the consumer is still working through buffered lines. `wait()` then subscribes after the event and hangs forever.

Observed in production as a frozen run: no terminal chunk, the run row stuck `running`, a leaked run slot, heartbeats forever.

The docker provider already guards this case (its `wait()` comment says "…or this hangs after kill()"). Aligning local-process with it fixes the hang: create the `close` promise at spawn time and have `wait()` return it:

```js
const closed = new Promise((resolve, reject) => {
  child.on("error", reject);
  child.on("close", (code) => resolve(code ?? 0));
});
closed.catch(() => {});
// ...
wait: () => closed,
```

We currently carry exactly this as a pnpm patch and can confirm it resolves the hang. Happy to open a PR if useful.

### Your Example Website or App

n/a (server-side reproduction described above)

### Steps to Reproduce the Bug or Issue

1. `spawnProcess` a short-lived child that prints many lines and exits.
2. Consume `handle.stdout` with an artificially slow consumer (await ~10ms per line) so the child closes while lines are still buffered.
3. Call `await handle.wait()` after the stdout iterator is done.
4. `wait()` never resolves.

### Expected behavior

`wait()` resolves with the exit code even when the child closed before `wait()` was called.

### Platform

- OS: Linux (Debian, node:24-slim container) and macOS
- @tanstack/ai-sandbox-local-process: 0.2.1

### Additional context

The stdout drain ordering comes from `spawnNdjson`, which reads stdout to completion before calling `wait()`.

mikemikimike:
## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed an issue where local processes could remain stuck when a child process exited before `wait()` was called.
  - Improved process completion handling when output is still being consumed.
  - Ensured all emitted output is received in the correct order during process shutdown.

- **Tests**
  - Added regression coverage for processes that exit before output consumption finishes.


AlemTuzlak (approved):
This fixes #1112.

`wait()` used to subscribe to `close` only when you called it. `spawnNdjson` drains stdout first, so the child can emit `close` before that subscription. Then `wait()` hangs forever.

The fix is correct: create the close/error promise at spawn time and return that same promise from `wait()`. `closed.catch(() => {})` only stops an unused rejection from becoming an unhandledRejection. A later `await wait()` still sees the same result.

I merged latest `main` and resolved the test conflict (kept both the new wait test and the killableProcesses tests). `handle.ts` auto-merged and still has `track()`, `detached`, and the logger on `killTree`.

Local check: `pnpm --filter @tanstack/ai-sandbox... build` then `vitest tests/local-process.test.ts --run` — 15 passed, 1 skipped (POSIX killTree on Windows).

Pre-existing, not part of this PR: a signal kill still maps `code ?? 0`, so a killed child can look like success. Do not expand this bugfix to change that mapping.

PR #1127, opened 2026-08-17
fix: preserve fallback structured-output timestamp ordering

### LINKED ISSUE #1125: Fallback structured-output completion events can have timestamps earlier than `structured-output.start`
### TanStack AI version

0.44.1

### Framework/Library version

@tanstack/ai-groq: 0.6.0 | @tanstack/openai-base: 0.9.12

When an adapter implements `structuredOutput()` but not `structuredOutputStream()`, TanStack wraps the non-streaming provider call with `fallbackStructuredOutputStream()`. 

The fallback records one timestamp before awaiting the provider and reuses it for every event emitted after the provider returns.

For a streaming consumer, `TextEngine.runStructuredFinalization()` synthesizes `structured-output.start` when it receives the fallback's first text event. 

That synthesized start uses `Date.now()` after the provider request has completed. The resulting stream is ordered correctly, but its timestamps are reversed:

```text
RUN_STARTED                   timestamp=t0
await adapter.structuredOutput()
structured-output.start      timestamp=t1
TEXT_MESSAGE_START            timestamp=t0
TEXT_MESSAGE_CONTENT          timestamp=t0
TEXT_MESSAGE_END              timestamp=t0
structured-output.complete   timestamp=t0
RUN_FINISHED                  timestamp=t0
```

This produces negative or misleading durations for consumers that derive timing from event timestamps. It can also make an otherwise valid persistence write fail when a consumer records `structured-output.start` as the start boundary and `structured-output.complete` as the end boundary. 

The same problem exists on the error path: a `RUN_ERROR` emitted after the awaited provider rejection retains the pre-request timestamp and can appear earlier than the synthesized structured-output start.

The issue is limited to the fallback wrapper. An adapter that emits its own native `structuredOutputStream()` controls its event timestamps directly.

### Suspected cause

`fallbackStructuredOutputStream()` captures `const timestamp = Date.now()` before calling `adapter.structuredOutput()` and then reuses that value for `RUN_ERROR`, every text event, `structured-output.complete`, and `RUN_FINISHED`: [`index.ts#L3996-L4083`](https://github.com/TanStack/ai/blob/8bde89a2bc83df140bb5dcdc0c8d73521fee6604/packages/ai/src/activities/chat/index.ts#L3996-L4083).

The surrounding structured-finalization loop creates the missing `structured-output.start` with a fresh `Date.now()` after it receives the first fallback text event: [`index.ts#L2858-L2869`](https://github.com/TanStack/ai/blob/8bde89a2bc83df140bb5dcdc0c8d73521fee6604/packages/ai/src/activities/chat/index.ts#L2858-L2869) and [`index.ts#L2904-L2922`](https://github.com/TanStack/ai/blob/8bde89a2bc83df140bb5dcdc0c8d73521fee6604/packages/ai/src/activities/chat/index.ts#L2904-L2922).

### Expected behavior

Fallback events emitted after the provider call should use timestamps captured after that call settles. `RUN_STARTED` should retain its pre-request timestamp, while successful text, completion, and `RUN_FINISHED` events should use current, nondecreasing emission timestamps. 

A `RUN_ERROR` should likewise use the time at which the provider failure is emitted.

This does not require a new event or API. It only requires the existing `timestamp` fields to reflect their emission boundaries. The following invariants should hold:

```text
RUN_STARTED.timestamp <= structured-output.start.timestamp
structured-output.start.timestamp <= structured-output.complete.timestamp
structured-output.complete.timestamp <= RUN_FINISHED.timestamp
```

The error path should preserve the equivalent ordering through `RUN_ERROR`.

https://stackblitz.com/edit/vitejs-vite-ue2jettl?file=package.json,index.html,src%2Fmain.ts

None

### Terms & Code of Conduct

mikemikimike:
Fixes #1125.

## Problem

Fallback structured-output events reused a timestamp captured before awaiting the provider. When `structured-output.start` was synthesized after the provider returned, later events could have earlier timestamps.

## Changes

- Keep `RUN_STARTED` at the request start boundary.
- Capture success and error timestamps after the provider settles.
- Add a regression test covering a delayed provider result.
- Add a patch changeset for `@tanstack/ai`.

## Compatibility

No API or wire-shape changes. Fallback lifecycle timestamps now reflect their emission boundaries and remain nondecreasing.

## Test plan

- `git diff --check` — passed.
- `pnpm install --frozen-lockfile --ignore-scripts` — not completed because npm registry requests repeatedly reset in the environment.
- Full test suite — not run because dependencies could not be installed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected lifecycle timestamps for fallback structured-output streams so events remain in chronological order when provider responses are delayed or fail.
* **Tests**
  * Added coverage for successful delayed responses and delayed provider errors to verify timestamp ordering through completion and failure.

### DIFF
--- a/.changeset/gentle-dots-fallback-timestamps.md
+++ b/.changeset/gentle-dots-fallback-timestamps.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/ai': patch
+---
+
+Fix fallback structured-output lifecycle timestamps being emitted before the provider request settles.
--- a/packages/ai/src/activities/chat/index.ts
+++ b/packages/ai/src/activities/chat/index.ts
@@ -5114,8 +5114,6 @@ async function* fallbackStructuredOutputStream(
     return
   }
 
-  const completedAt = Date.now()
-
   yield {
     type: EventType.TEXT_MESSAGE_START,
     messageId,
--- a/packages/ai/src/activities/chat/index.ts
+++ b/packages/ai/src/activities/chat/index.ts
@@ -5114,6 +5114,8 @@ async function* fallbackStructuredOutputStream(
     return
   }
 
+  const completedAt = Date.now()
+
   yield {
     type: EventType.TEXT_MESSAGE_START,
     messageId,
--- a/packages/ai/tests/chat-structured-output-stream.test.ts
+++ b/packages/ai/tests/chat-structured-output-stream.test.ts
@@ -480,6 +480,44 @@ describe('chat({ outputSchema, stream: true })', () => {
       expect(finished).toBeDefined()
       expect('usage' in finished!).toBe(false)
     })
+
+    it('keeps synthesized lifecycle timestamps ordered after a delayed provider result', async () => {
+      const adapter = makeAdapter({
+        structuredOutput: async () => {
+          await new Promise((resolve) => setTimeout(resolve, 5))
+          return { data: validPerson, rawText: JSON.stringify(validPerson) }
+        },
+      })
+
+      const stream = chat({
+        adapter,
+        messages: [{ role: 'user', content: 'extract' }],
+        outputSchema: PersonSchema,
+        stream: true,
+      })
+
+      const chunks = await collectChunks(stream)
+      const start = chunks.find(
+        (c) =>
+          c.type === EventType.CUSTOM &&
+          (c as { name?: string }).name === 'structured-output.start',
+      )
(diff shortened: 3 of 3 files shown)


tombeckenham:

tombeckenham:

tombeckenham:

PR #1148, opened 2026-08-19
fix(ai-react): avoid random IDs during SSR render

### LINKED ISSUE #1089: bug(ai-react): useChat calls crypto.randomUUID during Next.js 16.3 App Router SSR
## TanStack AI version

`@tanstack/ai-react@0.19.2` from the current `main` source tree.

## Framework/Library version

- Next.js `16.3` with the **App Router**
- React `19.2.3` in the repository test setup; the issue applies to React SSR generally
- React Server Components / Client Components rendered through the App Router

This report is specifically about **Next.js 16.3 App Router SSR**. A component using `useChat` can be a Client Component, but the App Router still renders it on the server for the initial response, so the hook's render-time work executes in the SSR process.

## Describe the bug and the steps to reproduce it

`useChat` constructs its internal `ChatClient` during render via `useMemo`. The React hook passes `createChatDevtoolsBridge` into the client, and the devtools bridge constructor generates a bridge ID. When `globalThis.crypto.randomUUID` exists, `createBridgeId()` calls it during the render phase.

The current call path is:

```text
Next.js 16.3 App Router server render
  -> useChat render
  -> new ChatClient()
  -> createChatDevtoolsBridge()
  -> new ChatDevtoolsBridge()
  -> createBridgeId()
  -> crypto.randomUUID()
```

Relevant source locations:

- [`packages/ai-react/src/use-chat.ts#L102-L127`](https://github.com/TanStack/ai/blob/main/packages/ai-react/src/use-chat.ts#L102-L127)
- [`packages/ai-client/src/chat-client.ts#L461-L464`](https://github.com/TanStack/ai/blob/main/packages/ai-client/src/chat-client.ts#L461-L464)
- [`packages/ai-client/src/devtools.ts#L459-L462`](https://github.com/TanStack/ai/blob/main/packages/ai-client/src/devtools.ts#L459-L462)
- [`packages/ai-client/src/devtools.ts#L747-L757`](https://github.com/TanStack/ai/blob/main/packages/ai-client/src/devtools.ts#L747-L757)

A minimal Next.js 16.3 App Router reproduction is a client component rendered by an App Router route:

```tsx
// app/chat.tsx
'use client'

import { fetchServerSentEvents, useChat } from '@tanstack/ai-react'

export function Chat() {
  useChat({
  })
  return null
}
```

```tsx
// app/page.tsx
import { Chat } from './chat'

export default function Page() {
  return <Chat />
}
```

When Next.js 16.3 renders `/` through the App Router, the `Chat` Client Component is also evaluated during the server render. The connection is not used by this reproduction. The `ChatClient` and devtools bridge are constructed before effects or commit.

In an SSR runtime that restricts random-value generation during server/global execution, the render can fail when `crypto.randomUUID()` is reached. Even when it does not throw, each server render, hydration render, or React-discarded render receives a different bridge ID.

There is a second nondeterministic path when `threadId` is omitted: `ChatClient` generates a thread ID with `Date.now()` and `Math.random()` in the same constructor.

## Expected behavior

Calling `useChat` from a Next.js 16.3 App Router Client Component during SSR should not perform nondeterministic random ID generation in the server render phase. SSR and hydration should be stable, and constructing a discarded render instance should not mint a new runtime identity as a side effect.

## Actual behavior

`useChat` reaches `crypto.randomUUID()` while constructing `ChatClient` during the Next.js 16.3 App Router server render. The current `useChat` tests cover StrictMode and client recreation, but do not cover Next.js App Router SSR rendering or hydration with restricted/nondeterministic randomness.

## Your Minimal, Reproducible Example

The Next.js 16.3 App Router reproduction above is repository-level and does not require an external sandbox.

## Related records

- [#667](https://github.com/TanStack/ai/issues/667) / [#683](https://github.com/TanStack/ai/pull/683) fixed module-scope random runtime-ID generation in `@tanstack/ai-event-client`; this issue is about render-time generation from `useChat` during Next.js App Router SSR.
- [#864](https://github.com/TanStack/ai/pull/864) proposes replacing the `useMemo` client initializer with a `useState` lazy initializer. It addresses duplicate `ChatClient` construction when React discards `useMemo`, but it still constructs the client during render and does not remove the `crypto.randomUUID()` call.

## PR

No PR attached.

mikemikimike:
Fixes #1089

Avoids render-time randomness in useChat by passing React's deterministic useId to ChatClient when threadId is omitted, and makes the devtools bridge identity derive solely from the stable hook ID. This keeps SSR render and hydration deterministic and avoids crypto.randomUUID during server render.

Tests: ai-client devtools suite (34 passed). ai-react use-chat suite could not run because the local workspace package link for @tanstack/ai-client is missing.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved chat thread consistency when no explicit thread ID is provided.
  - Empty thread ID values now receive a stable identifier automatically.
  - Ensured bridge identifiers remain stable across sessions and interactions.
  - Preserved explicitly configured thread IDs without changes.
  - Prevented unnecessary unique ID generation during chat initialization.
  - Improved reliability when initializing chats with omitted or empty thread settings.

### DIFF
--- a/packages/ai-client/src/devtools.ts
+++ b/packages/ai-client/src/devtools.ts
@@ -742,23 +742,11 @@ export class ClientDevtoolsBridge<TSnapshot extends object> {
   }
 }
 
-let bridgeIdSequence = 0
-
 function createBridgeId(hookId: string): string {
-  const cryptoLike = (
-    globalThis as {
-      crypto?: {
-        randomUUID?: () => string
-      }
-    }
-  ).crypto
-
-  if (cryptoLike?.randomUUID) {
-    return `bridge:${hookId}:${cryptoLike.randomUUID()}`
-  }
-
-  bridgeIdSequence += 1
-  return `bridge:${hookId}:${bridgeIdSequence}`
+  // hookId comes from React's useId and is stable across SSR and hydration.
+  // Do not generate randomness during render: crypto.randomUUID is unavailable
+  // or restricted in some server runtimes, and discarded renders must be pure.
+  return `bridge:${hookId}`
 }
 
 // Owns the chat-client devtools surface so the chat client itself stays a
--- a/packages/ai-react/src/use-chat.ts
+++ b/packages/ai-react/src/use-chat.ts
@@ -41,10 +41,10 @@ export function useChat<
 ): UseChatReturn<TTools, TSchema> {
   // The hook's identity is its `threadId` — also the persistence key, so a
   // reload with the same `threadId` restores the same conversation. `hookId` is
-  // only a stable fallback for React's client-recreation keying when no
-  // `threadId` is given (an ephemeral chat), never a persistence key.
+  // a stable fallback when an empty or missing `threadId` creates an ephemeral
+  // chat, never a persistence key.
   const hookId = useId()
-  const clientId = options.threadId ?? hookId
+  const clientId = options.threadId || hookId
 
   const [messages, setMessages] = useState<Array<UIMessage<TTools>>>(
     options.initialMessages || [],
@@ -128,10 +128,7 @@ export function useChat<
       ...transport,
       initialMessages: messagesToUse,
       ...(initialOptions.body !== undefined && { body: initialOptions.body }),
-      ...(initialOptions.threadId !== undefined && {
-        threadId: initialOptions.threadId,
-      }),
-      ...(initialOptions.threadId === undefined && { threadId: hookId }),
+      threadId: clientId,
       ...(initialOptions.forwardedProps !== undefined && {
         forwardedProps: initialOptions.forwardedProps,
       }),
--- a/packages/ai-react/src/use-chat.ts
+++ b/packages/ai-react/src/use-chat.ts
@@ -131,6 +131,7 @@ export function useChat<
       ...(initialOptions.threadId !== undefined && {
         threadId: initialOptions.threadId,
       }),
+      ...(initialOptions.threadId === undefined && { threadId: hookId }),
       ...(initialOptions.forwardedProps !== undefined && {
         forwardedProps: initialOptions.forwardedProps,
       }),
--- a/packages/ai-react/tests/use-chat.test.ts
+++ b/packages/ai-react/tests/use-chat.test.ts
@@ -290,6 +290,20 @@ describe('useChat', () => {
       expect(typeof messageId).toBe('string')
     })
 
+    it('uses the hook ID instead of generating a random ID for an empty threadId', () => {
+      const chatClientPrototype = ChatClient.prototype as unknown as {
+        generateUniqueId: () => string
+      }
+      const generateUniqueId = vi.spyOn(chatClientPrototype, 'generateUniqueId')
+
+      renderUseChat({
+        connection: createMockConnectionAdapter(),
+        threadId: '',
+      })
+
+      expect(generateUniqueId).not.toHaveBeenCalled()
+    })
+
     it('should generate id if not provided', async () => {
       const chunks = createTextChunks('Response')
       const adapter = createMockConnectionAdapter({ chunks })
--- a/packages/ai-react/tests/use-chat.test.ts
+++ b/packages/ai-react/tests/use-chat.test.ts
@@ -290,7 +290,9 @@ describe('useChat', () => {
       expect(typeof messageId).toBe('string')
     })
 
-    it('uses the hook ID instead of generating a random ID for an empty threadId', () => {
+    it.each(['', undefined])(
+      'uses the hook ID instead of generating a random ID for threadId=%s',
+      (threadId) => {
       const chatClientPrototype = ChatClient.prototype as unknown as {
         generateUniqueId: () => string
       }
@@ -298,10 +300,11 @@ describe('useChat', () => {
 
       renderUseChat({
         connection: createMockConnectionAdapter(),
-        threadId: '',
+        ...(threadId === undefined ? {} : { threadId }),
       })
 
-      expect(generateUniqueId).not.toHaveBeenCalled()
+        expect(generateUniqueId).not.toHaveBeenCalled()
+      },
     })
 
     it('should generate id if not provided', async () => {


packages/ai-react/src/use-chat.ts
```diff
@@ -131,6 +131,7 @@ export function useChat<
       ...(initialOptions.threadId !== undefined && {
         threadId: initialOptions.threadId,
       }),
+      ...(initialOptions.threadId === undefined && { threadId: hookId }),
```
AlemTuzlak: **[bug]** When `threadId` is omitted, this line sets `ChatClient.threadId` to React `useId()`. That is not a per-session id. `useId()` is unique per hook call on one page (two `useChat()` calls do not collide, Strict Mode remounts stay the same, SSR and hydration match). It is also the same string after a full reload if the tree is the same, and the same string for every user of that page. `ChatClient` then uses that value as the AG-UI wire `threadId` and as the persistence key (`persistenceKey = options.id ?? this.threadId`). The comment a few lines above this hook still says `hookId` is only a React recreation key and is never a persistence key. ChatClient docs still say an omitted `threadId` is unique per session. Concrete break: `useChat({ connection, persistence: adapter })` with no `threadId` used to write a new random key every load (reload starts empty). After this change it restores under `:R1:` / `_R_…`. `persistence: true` with no `threadId` is still type-legal and would hydrate one server thread shared by every visitor. Generation avoids this on purpose (`persistenceScope = options.threadId` only). Empty string / other falsy values are unchanged and still hit `options.threadId || generateUniqueId` inside ChatClient. Later `threadId` updates are fine: `clientId = options.threadId ?? hookId` recreates the client. `ai-preact` `use-chat` is a copy of this hook and was not updated.

**Suggestion:** Do not put `useId()` on the wire or in the persistence key. Keep `hookId` as the React recreation key / devtools `id` only. Stop `generateUniqueId('thread')` during render another way: build `ChatClient` after mount, or give ChatClient a generation-style split (`persistenceScope` is only an explicit `threadId`; the fallback id is wire-only and is created after commit). If you keep this line, update the hook comment and require `threadId` whenever `persistence` is set, the same way generation hooks already do.

packages/ai-client/src/devtools.ts
```diff
@@ -742,23 +742,11 @@ export class ClientDevtoolsBridge<TSnapshot extends object> {
...
-  if (cryptoLike?.randomUUID) {
-    return `bridge:${hookId}:${cryptoLike.randomUUID()}`
-  }
-
-  bridgeIdSequence += 1
-  return `bridge:${hookId}:${bridgeIdSequence}`
+  // hookId comes from React's useId and is stable across SSR and hydration.
+  // Do not generate randomness during render: crypto.randomUUID is unavailable
+  // or restricted in some server runtimes, and discarded renders must be pure.
+  return `bridge:${hookId}`
```
AlemTuzlak: **[bug]** `createBridgeId` is now `bridge:${hookId}` with no per-instance suffix. `bridgeId` is the envelope `correlationId`. The hook registry ignores a stale `hook:unregistered` only when `correlationId` (or `clientId`) differs from the live hook (`packages/ai-devtools/src/store/hook-registry.ts` and the test "ignores unregister events from a previous hook instance"). After this change, two bridges that share a `hookId` also share `correlationId`. That already happens for React generation hooks: they pass `id: hookId` (`useId()`), so `uniqueId` / bridge `hookId` is stable today. It will also happen for `useChat` after issue 1, because `uniqueId` becomes `useId()`. `useChat` defers `dispose()` by `setTimeout(..., 0)`. Unmount + remount in the same commit (swap two chat components at the same tree slot, Fast Refresh) registers the new client, then the old timeout unregisters with the same `correlationId` and the same `clientId`. `registeredAt > unregistered.timestamp` does not save you when unregister is later. The live hook disappears from the panel. Discarded `useMemo` clients that never mount are fine (they never emit unregister).

**Suggestion:** Keep render pure, but keep instance identity. Do not call `crypto.randomUUID` in the constructor. Assign the suffix in `mountDevtools` / `emitRegistered` (commit / effect), or use a mount-only counter. Constructor can use `bridge:${hookId}` as a placeholder and replace `correlationId` on first mount emit. Do not drop the uniqueness the registry test is built on.

packages/ai-client/src/devtools.ts
```diff
@@ -742,23 +742,11 @@ export class ClientDevtoolsBridge<TSnapshot extends object> {
...
-    }
-  ).crypto
-
-  if (cryptoLike?.randomUUID) {
-    return `bridge:${hookId}:${cryptoLike.randomUUID()}`
-  }
-
-  bridgeIdSequence += 1
-  return `bridge:${hookId}:${bridgeIdSequence}`
+  // hookId comes from React's useId and is stable across SSR and hydration.
```
AlemTuzlak: **[suggestion]** This comment is too long and it is wrong. `createBridgeId` lives in `ai-client`. Vue, Solid, Svelte, and a raw `new ChatClient()` all call it. `hookId` is `ChatClient.uniqueId` (explicit `threadId`, or `id`, or `generateUniqueId`). It is not "React's `useId`" except at some React hook call sites. The three lines also retell the design (SSR, discarded renders, `crypto.randomUUID`) instead of a short why.

**Suggestion:** Delete the comment, or keep one short line: "Instance suffix is assigned on mount so render stays pure." Do not claim every `hookId` comes from React `useId`.

packages/ai-react/src/use-chat.ts
AlemTuzlak: **[suggestion]** Issue #1089 already says the current `useChat` tests cover Strict Mode and client recreation, not Next.js App Router SSR / hydration, and not restricted `crypto.randomUUID`. This PR adds no test. Existing `use-chat` tests never read the omitted-`threadId` value sent on the wire (`should generate id if not provided` only checks message ids). `ai-client` devtools tests do not assert `createBridgeId` / `correlationId` shape. CI did run Test and E2E and they passed; the author's missing `@tanstack/ai-client` workspace link is a fork setup problem, not a repo problem. Green CI is not coverage of this bug.

**Suggestion:** Add a `useChat` test that omits `threadId`, captures `runContext.threadId` from the mock adapter, and asserts it is the hook `useId` (or whatever identity you settle on in issue 1) and is stable across rerender / Strict Mode. Add a registry or bridge test that a late unregister from a previous instance with the same `hookId` does not drop the new instance. A render-purity test that stubs `crypto.randomUUID` to throw during `useChat` render would lock the original #1089 failure.

AlemTuzlak: **[nit]** The new spread is the inverse of the block on lines 131-133. Both are mutually exclusive. `threadId` is always a string after this change (`hookId` from `useId()`), so the `exactOptionalPropertyTypes` omit-if-absent dance is not needed for this field.

**Suggestion:** Replace both spreads with `threadId: initialOptions.threadId ?? hookId` if you still pass `hookId` here after issue 1. If you stop passing `hookId` as `threadId`, delete line 134.

AlemTuzlak:
Fixed and addressed in another PR

PR #1169, opened 2026-08-20
fix(ai-client): drain client-tool results after joinRun replay

### LINKED ISSUE #1058: `joinRun` rejoin never drains post-stream actions, stalling a client-tool turn and blocking all later sends
### Package

`@tanstack/ai-client` 0.23.0 (via `@tanstack/ai-react` 0.19.0, `@tanstack/ai` 0.43.0)

### Summary

When a run is rejoined with `joinRun` (resumable stream replay after a reload / app relaunch) and that run ends on a **client tool** call, the tool executes but its result is never sent to the server. The turn stalls forever, and because the interrupt stays pending, every subsequent `sendMessage` is silently held too — the chat is permanently wedged until the persisted state is cleared.

Live runs are unaffected; only the rejoin path is broken.

### Reproduction

1. Configure `useChat` with a resumable connection (`fetchServerSentEvents` against a route using `toServerSentEventsResponse(stream, { durability: { adapter: memoryStream(request) } })`) and a `persistence` adapter.
2. Register a client tool (`toolDefinition(...).client(fn)`) and pass the bare definition to `chat({ tools })` on the server.
3. Send a message that makes the model call that client tool.
4. Kill the app / reload the page while the run is still streaming.
5. Reopen. The client rejoins via `GET ...?offset=-1&runId=...` and replays the run correctly: text and the tool call are repainted, and the client tool executes.

Expected: the tool result is posted back and the run continues to completion.
Actual: no follow-up request is made. The turn never finishes, and every later `sendMessage` produces no request at all.

### Cause

`resumeInFlightRun` sets `isLoading = true` for the duration of the replay:

```js
resumeInFlightRun(runId) {
  ...
  this.setIsLoading(true);
  this.setStatus("streaming");
  (async () => {
    try {
      for await (const chunk of joinRun(runId, controller.signal)) { ... }
    } catch (error) { ... }
    finally {
      ...
      if (this.abortController === controller) {
        this.abortController = null;
        this.setIsLoading(false);
        if (this.status === "streaming") this.setStatus("ready");
      }
    }
  })();
}
```

During the replay `onToolCall` fires, the client tool runs, and `addToolResultForClientTool` resolves the interrupt, which makes `InterruptManager` call `submit` → `resumeInterruptsUnsafe`. That method defers whenever a stream is in flight:

```js
resumeInterruptsUnsafe(resume, state) {
  const target = state ?? this.lastResume;
  if (!target) return Promise.resolve(false);
  if (this.isLoading) return new Promise((resolve, reject) => {
    this.queuePostStreamAction(async () => { ... });   // deferred
  });
  ...
}
```

So the submission is queued. But `drainPostStreamActions()` is only ever called from `streamResponse()`'s teardown — `resumeInFlightRun` has its own `finally` and never drains. The queued submission is therefore never executed:

- the continuation request is never sent, so the turn never completes;
- `activeInterruptSubmission` / the pending interrupt are never cleared, so later sends are blocked behind them.

The same would apply to any post-stream action queued during a rejoin, including the `checkForContinuation()` queued by `addToolResultForClientTool` when it takes the non-interrupt path.

### Suggested fix

Drain the queue in `resumeInFlightRun`'s `finally`, after `isLoading` is cleared so the deferred call takes the live path on re-entry:

```diff
                 if (this.abortController === controller) {
                     this.abortController = null;
                     this.setIsLoading(false);
                     if (this.status === "streaming") this.setStatus("ready");
+                    await this.drainPostStreamActions();
                 }
```

`drainPostStreamActions` already guards against re-entrancy with `this.draining`, so this is safe next to the existing call in `streamResponse`.

I'm running this as a `patch-package` patch and it resolves both symptoms.

AlemTuzlak:
After a page reload, a `joinRun` replay that ended on a client tool never sent the tool result back to the server. Later sends stayed blocked.

The live `streamResponse` path already drains that queue in `finally`. Rejoin used its own teardown and skipped the drain. This PR waits for pending client tools, then drains after `joinRun`, same as a live stream.

## 🎯 Changes

- `resumeInFlightRun` waits for `pendingToolExecutions`, then calls `drainPostStreamActions()`.
- Unit test holds `joinRun` open until the client tool finishes, then asserts the continuation POST.
- E2E reloads mid client-tool run, tails with `joinRun`, and asserts `JOIN_OK` (that text exists only on the continuation POST).
- Docs: one sentence on the resumable-streams advanced page.
- Patch changeset for `@tanstack/ai-client`.

## ✅ Checklist

## 🚀 Release Impact

## Testing

**Commands run**

1. `pnpm --filter @tanstack/ai-client test:lib -- tests/chat-client-join-run-client-tool.test.ts` — pass
2. `pnpm --filter @tanstack/ai-e2e test:e2e -- join-run-client-tool.spec.ts --retries=0` — fail without drain (`JOIN_OK` missing, tool ran), pass with drain (4.9s)
3. `pnpm --filter @tanstack/ai-e2e test:types` — pass after the `output` narrow
4. `pnpm exec oxfmt --check` on the touched files — pass
5. `pnpm test:pr` — ran. Failed on local dirt: knip on untracked `marketing/`, docs links under `docs/superpowers/worktrees`, and a missing local `vitest.mjs` in some parallel `test:lib` tasks. Those files are not in this PR.

**Manual test**

1. Start a chat with a client tool and `joinRun` (durable connection + `persistence: true`).
2. Reload while the run is still open and the client tool is pending.
3. Confirm the remount tails the run, the tool finishes, and the server gets the tool result (a follow-up assistant message).
4. Send another user message. Confirm it is not stuck on loading.

- Unit: `packages/ai-client/tests/chat-client-join-run-client-tool.test.ts`
- E2E: `pnpm --filter @tanstack/ai-e2e test:e2e -- join-run-client-tool.spec.ts`

## Linked issues

Fixes #1058

## Risk / rollback

Low. The drain matches the live stream path. Revert the PR to undo.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved reconnect and replay handling when a run pauses for a client-side tool.
  - Ensured tool results are sent after replay, allowing conversations to continue normally.
  - Preserved follow-up messaging after reconnecting to an interrupted run.

- **Documentation**
  - Updated resumable-stream guidance with client-tool replay behavior.
  - Added end-to-end testing documentation for reconnect and replay scenarios.

- **Tests**
  - Added coverage for reload, replay, client-tool completion, and subsequent messages.

### DIFF
--- a/.changeset/join-run-drain-client-tool.md
+++ b/.changeset/join-run-drain-client-tool.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/ai-client': patch
+---
+
+Drain queued post-stream actions after a `joinRun` replay so a client-tool result from a rejoin is sent back to the server.
--- a/docs/config.json
+++ b/docs/config.json
@@ -243,7 +243,7 @@
           "label": "Advanced",
           "to": "resumable-streams/advanced",
           "addedAt": "2026-08-04",
-          "updatedAt": "2026-08-19"
+          "updatedAt": "2026-08-20"
         },
         {
           "label": "WebSockets",
--- a/docs/resumable-streams/advanced.md
+++ b/docs/resumable-streams/advanced.md
@@ -101,7 +101,9 @@ A durable run's producer is decoupled from the delivery socket. When the client
 disconnects (a page reload, a dropped connection), the response is cancelled but
 the run keeps draining into the log to its own terminal, so a reconnect or a
 mount-time `joinRun` tails it to completion. This is what makes a run resumable
-across a full reload, not just an in-session reconnect.
+across a full reload, not just an in-session reconnect. If that replay ends on
+a client tool, the client sends the tool result after the replay finishes. That
+is the same path as a live stream.
 
 A run ends early only on a genuine cancel or a failure:
 
--- a/packages/ai-client/src/chat-client.ts
+++ b/packages/ai-client/src/chat-client.ts
@@ -1662,6 +1662,12 @@ export class ChatClient<
           }
           await this.processIncomingChunk(chunk, { defer: false })
         }
+        // Same contract as `streamResponse`: client tools may finish (and
+        // queue a resume) while `isLoading` is still true. Wait for them
+        // before teardown so `drainPostStreamActions` below sees the queue.
+        if (this.pendingToolExecutions.size > 0) {
+          await Promise.all(this.pendingToolExecutions.values())
+        }
       } catch (error) {
         // Pre-attach failures (unknown/evicted run, connect deadline abort)
         // stay soft: keep the restored transcript. Post-attach transport/parser
@@ -1703,6 +1709,7 @@ export class ChatClient<
           this.abortController = null
           this.setIsLoading(false)
           if (this.status === 'streaming') this.setStatus('ready')
+          await this.drainPostStreamActions()
         }
       }
     })()
--- a/packages/ai-client/tests/chat-client-join-run-client-tool.test.ts
+++ b/packages/ai-client/tests/chat-client-join-run-client-tool.test.ts
@@ -1,0 +1,211 @@
+import { describe, expect, it, vi } from 'vitest'
+import {
+  EventType,
+  canonicalInterruptJson,
+  convertSchemaToJsonSchema,
+  digestInterruptJson,
+  hashSchemaInput,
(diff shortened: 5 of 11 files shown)


jherr (approved):
LGTM

PR #1198, opened 2026-08-22
fix(ai-gemini): dedupe functionResponse parts by id, not name

### LINKED ISSUE #894: ai-gemini: parallel calls to the same tool drop a functionResponse part, causing Gemini 400 (function response/call part count mismatch)
### TanStack AI version

`@tanstack/ai-gemini` 0.19.0 (with `@tanstack/ai` 0.39.0)

### Framework/Library version

Node.js 24.15.0, server-side (no UI framework).

When Gemini returns two or more parallel function calls to the **same** tool in one turn, the adapter throws away all but one of the matching `functionResponse` parts while building the next request. Gemini requires exactly one response part per call part in a turn, so the follow-up request gets rejected with:

> 400 INVALID_ARGUMENT: Please ensure that the number of function response parts is equal to the number of function call parts of the function call turn.

I ran into this with a currency-lookup tool. A message like "50 USD and 30 EUR" makes the model fire `lookupCurrency` twice in the same turn, and the second call's response never makes it into the request, so the run dies on the next step.

The cause is in `mergeConsecutiveSameRoleMessages` (`packages/ai-gemini/src/adapters/text.ts`, ~L855). After merging the tool-result turns, it dedupes `functionResponse` parts by tool **name**:

```ts
const seenFunctionResponseNames = new Set<string>()
msg.parts = msg.parts.filter((part) => {
  if ('functionResponse' in part && part.functionResponse?.name) {
    if (seenFunctionResponseNames.has(part.functionResponse.name)) {
      return false
    }
    seenFunctionResponseNames.add(part.functionResponse.name)
  }
  return true
})
```

Two parallel calls to the same tool share a `name` but have different `id`s, so the second response gets filtered out. The comment above it calls the key "(tool call ID)", but it's actually keying on `name`. Every `functionResponse` is already built with a unique `id` (the `toolCallId`), so keying on `id` instead keeps both parts, and still collapses a genuine duplicate if that's what the filter was guarding against.

Steps:

1. Register any tool.
2. Send a message that makes the model call it twice in one turn (two currencies to look up, two lookups, etc.).
3. The two responses merge into one turn, the dedup drops one, and the next request 400s.

It's intermittent in practice because it only triggers when the model happens to pick same-tool parallel calls. A retry that doesn't repeat that choice appears to "fix" it, which hides the real cause.

### Your Minimal, Reproducible Example

Self-contained, no network or API key needed. It only drives the adapter's message formatting:

```js
// npm i @tanstack/ai-gemini@0.19.0
// node repro.mjs
import { GeminiTextAdapter } from '@tanstack/ai-gemini'

const adapter = new GeminiTextAdapter({ apiKey: 'not-used' }, 'gemini-2.5-flash-lite')

// One assistant turn with two parallel calls to the SAME tool (distinct ids),
// followed by the two matching tool results.
const messages = [
  { role: 'user', content: 'log 50 USD and 30 EUR' },
  { role: 'assistant', content: '', toolCalls: [
    { id: 'call_1', function: { name: 'lookupCurrency', arguments: '{"query":"USD"}' } },
    { id: 'call_2', function: { name: 'lookupCurrency', arguments: '{"query":"EUR"}' } },
  ] },
  { role: 'tool', toolCallId: 'call_1', content: 'USD -> US Dollar' },
  { role: 'tool', toolCallId: 'call_2', content: 'EUR -> Euro' },
]

const out = adapter.formatMessages(messages)
const calls = out.find((c) => c.role === 'model').parts.filter((p) => 'functionCall' in p)
const responses = out.flatMap((c) => c.parts).filter((p) => 'functionResponse' in p)

console.log('call parts:', calls.length, '| response parts:', responses.length)
// call parts: 2 | response parts: 1  -> counts don't match, Gemini 400s the next request
```

Output:

```
call parts: 2 | response parts: 1
```

### Suggested fix

Key the dedup on `part.functionResponse.id` instead of `.name` (or drop it, since the ids are already unique). Looks like a one-liner. Happy to open a PR if that helps.

Yes, I think I know the fix (dedup by `id` rather than `name`) and can put up a PR.

citizen204:
## Summary

When Gemini fires two or more parallel calls to the *same* tool in one turn, `mergeConsecutiveSameRoleMessages` deduped `functionResponse` parts by `name`, so the second (and any further) response to a repeated same-tool call was silently dropped. Gemini requires exactly one response part per call part in a turn, so the next request 400s:

> INVALID_ARGUMENT: Please ensure that the number of function response parts is equal to the number of function call parts of the function call turn.

Every `functionResponse` already carries a unique `id` (`msg.toolCallId`, set at both construction sites) — keying the dedup on `id` instead keeps both parallel responses while still collapsing a genuine duplicate tool result (same id sent twice).

Fixes #894

## Changes

- `packages/ai-gemini/src/adapters/text.ts`: dedupe `functionResponse` parts by `id` instead of `name` in `mergeConsecutiveSameRoleMessages`.
- `packages/ai-gemini/tests/gemini-adapter.test.ts`: regression test with two parallel calls to the same tool (distinct ids) — asserts both `functionResponse` parts survive.
- `.changeset/gemini-parallel-tool-dedup.md`: patch changeset.

Ran the full `ai-gemini` suite (`vitest run` in `packages/ai-gemini`, after building its `@tanstack/ai`, `@tanstack/ai-event-client`, `@tanstack/ai-utils` workspace deps) — 325 passed, 18 files.

Note: an earlier PR (#960) attempted this exact fix and was self-closed by its author for unrelated reasons ("keeping open PRs lean") after CodeRabbit review raised no substantive objections — the bug is still present on `main`, so re-submitting with the same diagnosis plus a parallel-call regression test.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed parallel tool calls using the same tool so all distinct responses are preserved.
  * Prevented Gemini request failures caused by incorrectly removing valid tool responses.
  * Genuine duplicate responses continue to be consolidated.


AlemTuzlak:

**Verdict:** ready
**Head SHA:** 7ec18ee24c60ddf6b5f73ccc7d4703fd5e20a99d
**Label:** `ai-ready`

**Findings**
- None

**Push**
Did not push.

**Security**
clean. Added label `secure`. Approved 5 waiting workflow runs.

AlemTuzlak (approved):

PR #1209, opened 2026-08-23
fix(openai-base): preserve tuple schema items

### LINKED ISSUE #1208: @tanstack/openai-base: coerceStrictSchema mangles tuple schemas (`items` as an array) into a numeric-keyed object, silently dropping element constraints
### TanStack AI version

0.45.1 (`@tanstack/openai-base@0.9.14`; unchanged on current main)

### Framework/Library version

Tanstack Start

**Package:** `@tanstack/openai-base`

## Summary

`coerceStrictSchema` in [`packages/openai-base/src/utils/schema-converter.ts`](https://github.com/TanStack/ai/blob/main/packages/openai-base/src/utils/schema-converter.ts) assumes an array schema's `items` is a single schema object. In draft-07 a positional tuple declares `items` as an *array* of per-position schemas (this is what e.g. `z.tuple(...)` produces via zod-to-json-schema). Both array branches pass that array straight into `coerceStrictSchema`, whose first line is `const result = { ...schema }`. Spreading an array produces a numeric-keyed object:

```json
{ "type": "array", "items": [ { "type": "number", "minimum": -180 }, { "type": "number", "minimum": -90 } ] }
```

becomes

```json
{ "type": "array", "items": { "0": { "type": "number", "minimum": -180 }, "1": { "type": "number", "minimum": -90 } } }
```

That `items` is not a valid schema, and every per-position constraint is silently dropped. The model receives a tool whose tuple parameter carries no element types or bounds at all.

Notably the strict-compatibility gate already understands the tuple form: `containsTypelessSchema` spreads `Array.isArray(schema.items) ? schema.items : [schema.items]`. So a tuple passes the gate as strict-compatible and is then mangled by the coercion. The two walks disagree about what `items` can be.

## How we hit it

A tool parameter was a bbox tuple with per-edge bounds (west, south, east, north). The model kept receiving the parameter with no element constraints and produced malformed arguments. We worked around it by remodeling the tuple as a named object `{ west, south, east, north }`, so we carry no patch for this.

## Suggested fix

Handle `Array.isArray(items)` in both array branches: coerce each position schema individually and keep the array shape (plus `additionalItems` when present). The 2020-12 equivalent `prefixItems` deserves the same treatment while you are in there; today it passes through untouched, which at least does not corrupt the schema, but its position schemas also never get the strict-mode normalization.

Happy to PR the fix with a regression test.

mikemikimike:
Closes #1208

## Summary
Fix tuple schemas where JSON Schema items is an array. The converter now preserves the positional array and recursively coerces each item schema. Draft 2020-12 prefixItems are handled similarly.

## Compatibility
This only changes strict-schema normalization for tuple and prefix-item arrays. Single-schema items behavior is unchanged.

## Verification
- Added a regression test covering positional tuple schemas.
- Full pnpm test:pr and E2E checks were not run because the repository could not be cloned in the environment; changes were applied through the GitHub Contents API.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved structured-output schema conversion for optional fields using enums, constants, and nullable variants.
  * Preserved positional tuple schemas and correctly handled nested objects and arrays.
  * Added stricter validation for schemas that cannot be safely converted.

### DIFF
--- a/.changeset/fix-tuple-schema-coercion.md
+++ b/.changeset/fix-tuple-schema-coercion.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/openai-base': patch
+---
+
+Preserve and recursively coerce positional tuple schemas in OpenAI structured output conversion.
--- a/packages/openai-base/src/utils/schema-converter.ts
+++ b/packages/openai-base/src/utils/schema-converter.ts
@@ -1,3 +1,5 @@
+import type { NullWideningMap } from '@tanstack/ai-utils'
+
 /**
  * String `format` values accepted by OpenAI's strict Structured Outputs subset.
  * Any other format (e.g. "uri", "uri-reference", "regex") causes the API to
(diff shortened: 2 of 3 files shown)


mikemikimike:
Closing this superseded PR because its fork baseline included unrelated upstream changes. A clean minimal PR will follow.

PR #1232, opened 2026-08-24
feat(ai-client): publish observable client snapshots

AlemTuzlak:
Client UI state is now a frozen snapshot. Vanilla clients and every framework hook observe that snapshot. React keeps the first paint on `initialMessages` so SSR hydration stays stable. Solid and Octane apply the snapshot during setup so a persisted transcript is on screen before first paint.

Examples and the e2e app now use Router 1.170 and Start 1.168. `router-core` no longer constructs Store 0.8 next to `@tanstack/ai-client` Store 0.11.

## 🎯 Changes

`@tanstack/ai-client` stores chat, generation, video, realtime, audio, and BYOK UI state in TanStack Store atoms. Clients publish frozen snapshots. Framework hooks keep the existing return shape and read `getSnapshot()`.

`ChatClientSnapshot.messages` and `queue` are `ReadonlyArray`. Nested message and queue entries are copies. In-place `messages.push` throws in every adapter. Use `setMessages`.

`ChatClient` adds `getSnapshot()` and `subscribeSnapshot()`. Other clients add `getSnapshot()` and `subscribe()` where the name does not collide. `AudioRecorder.subscribe` still passes the new state value.

Generation and video keep the local run id while a request is in flight. Published `result` is a shallow copy, not the live client object.

Docs: `docs/api/ai-client.md` and `docs/persistence/client-persistence.md`. Changeset: `.changeset/ai-client-store-snapshot.md`.

## ✅ Checklist

## 🚀 Release Impact

## Testing

**Commands run**

1. `pnpm test:pr` passed after rebase onto current `main` at HEAD `47ea59cda`.
2. I did not re-run Playwright after that rebase.
3. After the Router and Start bump, openai chat SSE, openai chat fetcher, and abort Playwright specs passed locally.

**Manual test**

1. Open the `ts-react-chat` example.
2. Send a chat message and wait for streaming to finish.
3. Reload the page and make sure that the persisted transcript returns without a hydration error.
4. Send another message and make sure that status, queue, and message updates continue.
5. Make sure that `messages.push(...)` throws and that `setMessages` still works.

- Snapshot unit coverage: `packages/ai-client/tests/client-snapshots.test.ts` and `snapshot-atom.test.ts`.
- Frozen `messages.push`: React and Preact `use-chat` tests.
- Generation run id during a chunk with no resume pointer: `generation-client.test.ts`.
- Solid and Octane first-paint persisted messages: `use-chat` tests.
- E2E coverage for chat, image-gen, video-gen, and persistence durability in `testing/e2e/tests/`.

## Risk / rollback

A missed client mutation can publish stale UI state. Incorrect hydration order can replace the server snapshot before React commits. Revert this PR to restore callback-based framework state copies.

React, Preact, Vue, Svelte, Angular, Solid, and Octane all hand out frozen snapshot arrays. In-place `messages.push` throws. Use `setMessages`.

Apps that still use Router 1.159 can hit a Store 0.8 / 0.11 Vite merge. Hydration then throws `actionsFactory is not a function` and the page stays blank. Upgrade Router and Start with this PR, or isolate `@tanstack/store` in Vite.

## Public API change

The additions are backward-compatible at the hook return shape. New snapshot types are exported from `@tanstack/ai-client`. Snapshot `messages` and `queue` are typed `ReadonlyArray`.

**Before**

```ts
const client = new ChatClient({
  onMessagesChange: renderMessages,
})
```

**After**

```ts
const client = new ChatClient({
})

const unsubscribe = client.subscribeSnapshot(() => {
  const snapshot = client.getSnapshot()
  renderMessages(snapshot.messages)
})
```

`GenerationClient`, `VideoGenerationClient`, `RealtimeClient`, and `ByokClient` expose `subscribe` / `getSnapshot`. `AudioRecorder.subscribe` still passes the new state value.


packages/ai-solid/src/use-chat.ts
```diff
@@ -159,53 +154,35 @@ export function useChat<
...
+    setStatus(next.status)
+    setIsSubscribed(next.isSubscribed)
+    setConnectionStatus(next.connectionStatus)
+    setSessionGenerating(next.sessionGenerating)
+    setQueue(next.queue)
+    setRunId(next.runId)
+    setInterruptState(next.interruptState)
+  }
+
+  createEffect(() => {
```
AlemTuzlak: **[bug]** Solid `useChat` now applies client state only inside `createEffect`. In Solid, `createEffect` runs after the first paint (and not during SSR). BASE applied `client().getMessages()` and resume state in setup, before the hook returned, so the first `messages()` / `runId()` / interrupt read already matched a sync store. HEAD still creates the `ChatClient` in setup (and the client still reads a sync store in its constructor), but the signals stay on `options.initialMessages` until the effect. That is a first-paint regression for CSR apps with `localStorage` / `sessionStorage`. It also disagrees with `docs/persistence/client-persistence.md`, which says only React delays persistence, and with Solid `useGeneration` / `useGenerateVideo`, which still call `applySnapshot()` during setup.

**Suggestion:** Apply the snapshot once during setup (same as BASE `setMessages(client().getMessages())` and the Solid generation hooks). Keep `createEffect` only to subscribe and to re-bind when `client()` changes.

packages/ai-react/src/use-chat.ts
```diff
@@ -555,7 +491,7 @@ export function useChat<
   // a stale assistant turn or a system prompt) we deliberately return null
   // rather than scanning historical assistants — otherwise a `final` from a
   // previous session would leak into the hook value on first render.
-  const renderedMessages = client.getMessages()
+  const renderedMessages = snapshot.messages
```
AlemTuzlak: **[suggestion]** `useChat` now returns `snapshot.messages`, `snapshot.queue`, and `snapshot.interruptState.*` from `useSyncExternalStore`. Those arrays/objects are frozen in `ChatClient` (`freezeSnapshotMessages` / `freezeSnapshotQueue`, and a shallow `Object.freeze` in `packages/ai-client/src/snapshot-atom.ts`). BASE returned `client.getMessages()`, which is the live processor array (`this.messages`), so `messages.push(...)` or editing a part did not throw. HEAD throws `TypeError` on in-place mutation. Field names and TypeScript types are still `Array<...>`. `setMessages` still works.

**Suggestion:** Keep freeze for `useSyncExternalStore` identity. Document `messages` / `queue` / interrupt lists as read-only, or type them as `ReadonlyArray`. Do not clone on every snapshot unless you cache the clone; a new clone each `getSnapshot()` call would break tearing checks.

packages/ai-preact/src/use-chat.ts
```diff
@@ -521,31 +391,31 @@ export function useChat<
     [client],
   )
 
-  const renderedMessages = client.getMessages()
+  const renderedMessages = snapshot.messages
```
AlemTuzlak: **[suggestion]** Same as React `useChat`: Preact now returns `snapshot.messages` / `snapshot.queue` / `snapshot.interruptState.*` from `useSyncExternalStore`. BASE returned `client.getMessages()` (live, unfrozen). In-place mutation of the returned arrays now throws. No persistence gate and no `getServerSnapshot` (same first-paint hydration as BASE `getMessages()`).

**Suggestion:** Same as React: document read-only snapshot fields, or type them as `ReadonlyArray`. Keep freeze; do not clone per read.

packages/ai-react/src/use-realtime-chat.ts
```diff
@@ -258,16 +235,16 @@ export function useRealtimeChat(
...
     connect,
     disconnect,
 
     // Conversation state
-    mode,
-    messages,
-    pendingUserTranscript,
-    pendingAssistantTranscript,
+    mode: snapshot.mode,
+    messages: snapshot.messages,
```
AlemTuzlak: **[suggestion]** `useRealtimeChat` now returns `snapshot.messages` (and the other snapshot fields) directly. `RealtimeClient` freezes the messages array and each message/part when it builds the snapshot (`packages/ai-client/src/realtime-client.ts`, `Object.freeze` on `messages`). BASE kept messages in `useState` and appended with `[...prev, message]`, so the array the app held was a normal mutable copy. `connect()` still clears transcript state on the client before `await`, so dropping the local `setMessages([])` in `connect` is not a timing break.

**Suggestion:** Document realtime `messages` as read-only, or copy into hook state if you want the old mutability. Returning the frozen snapshot is fine for `useSyncExternalStore`.

PR #1235, opened 2026-08-24
feat(ai-compaction): pluggable context-window compaction

jherr:
Add `@tanstack/ai-compaction` so a long chat can stay under the model context limit. The full transcript stays in `messages`. The model sees a smaller `providerMessages` view. Middleware can also call `ctx.emitCustomEvent` so CUSTOM events reach the client while a hook is still running.

This branch includes latest `main`. DevTools keeps both the Compaction tab from this PR and the Skills tab from `main`.

Pending review comments on this PR are addressed:

1. Trailing tool results stay paired with their assistant. The tail is never empty.
2. Compaction skips `init`. `summarize` / `onCompact` fire once per `beforeModel` call.
3. Folding needs a checkpoint. `summarizeOldest` has a built-in strategy key.
4. The default summary is an `assistant` note inside `<untrusted-conversation-summary>`.
5. The `ponytail:` comment prefix is gone.

## 🎯 Changes

- New package `@tanstack/ai-compaction`. `withCompaction({ maxTokens, strategy })` runs on `beforeModel` and `structuredOutput`, not `init`.
- Three strategies: `evictOldest` (default), `summarizeOldest`, `clearToolResults`. Combine them with `composeStrategies`.
- Compaction writes `providerMessages`. Persistence still saves the full `messages` list.
- `ctx.emitCustomEvent(name, value)` on `ChatMiddlewareContext`. The engine yields each CUSTOM chunk while the hook is still running.
- Compaction emits `compaction:started`, then `compaction:state`, then `compaction:ended`. DevTools has a Compaction tab.

## ✅ Checklist

## 🚀 Release Impact

## Testing

**Commands run.**

1. `packages/ai-compaction` `test:lib`: 22 passed
2. Earlier: `packages/ai-client` 727 passed, `packages/ai-devtools` 53 passed, sherif clean, oxlint 0 errors
3. GitHub PR Test on `0b37970` passed. This push is `719263cb6`.

**Manual test.**

1. Run `pnpm --dir examples/ts-react-chat exec vite dev --port 3011 --strictPort`.
2. Open `http://localhost:3011/compaction`.
3. Send a few long messages until compaction runs.
4. Reload. The chat still shows every message.
5. Send another message. It must not throw on `createdAt`.

**How this PR makes testing easy.**

- Trailing tool groups: `packages/ai-compaction/src/index.test.ts`
- Skip `init`, one summarize per `beforeModel`: same file
- Example: `/compaction` in `examples/ts-react-chat`

## Risk / rollback

`emitCustomEvent` is a new required field on `ChatMiddlewareContext`. Hand-built context stubs must add it. Nothing else runs unless you add `withCompaction`. To undo, revert this PR.

## Public API change

**Before**

```ts
const mw: ChatMiddleware = {
  async onConfig() {
    await slowWork()
  },
}
```

**After**

```ts
const mw: ChatMiddleware = {
  async onConfig(ctx) {
    ctx.emitCustomEvent('job:started', { step: 'prepare' })
    await slowWork()
    ctx.emitCustomEvent('job:ended', { step: 'prepare' })
  },
}
```

New package usage:

```ts
import { withCompaction } from '@tanstack/ai-compaction'
import { withPersistence } from '@tanstack/ai-persistence'

chat({
  adapter,
  messages,
  threadId,
  runId,
  middleware: [
    withPersistence(persistence),
    withCompaction({ maxTokens: 100_000 }),
  ],
})
```

## Summary by CodeRabbit

* **New Features**
  * Added configurable chat context compaction with eviction, summarization, tool-result clearing, and composable strategies.
  * Preserved complete transcripts while optimizing context sent to models.
  * Added checkpoint reuse through persistence metadata.
  * Added custom middleware events and compaction lifecycle notifications.
  * Added a Compaction view in AI DevTools with previews, statistics, and event history.
  * Added interactive compaction examples with model, token-limit, and strategy controls.

* **Documentation**
  * Expanded guidance for compaction, middleware events, persistence, and provider-specific message context.


AlemTuzlak (approved):

packages/ai-compaction/src/index.ts
```diff
@@ -0,0 +1,387 @@
...
+    const prev = messages[cut - 1]
+    if (!prev) break
+    const size = estimate(prev)
+    if (kept + size > keepRecentTokens) break
+    kept += size
+    cut--
+  }
+  // Always keep at least the last message.
+  if (cut >= messages.length) cut = messages.length - 1
+  while (cut < messages.length && messages[cut]?.role === 'tool') cut++
```
AlemTuzlak: **[bug]** `splitAtRecent` first forces a keep of the last message, then walks `cut` forward while that message is `role === 'tool'`. If the transcript ends in tool results, `cut` becomes `messages.length`. `evictOldest` / `summarizeOldest` then return only the marker/summary (`messages.slice(cut)` is empty). That is the normal state after a tool phase, right before the next model call. Large tool output is also the case compaction is meant to handle. The unit test at `packages/ai-compaction/src/index.test.ts:216` always puts a user message after the tool result, so this path is not covered. Docs at `docs/advanced/compaction.md:192` say the request stays valid because tool calls stay paired. The code avoids orphans by deleting the last turn instead.

**Suggestion:** After the tool skip, if `cut >= messages.length`, walk backward to the assistant message that owns those tool results and keep that whole group. Never return a cut that drops every real message. Add a test whose last message is a tool result (and one whose last messages are several parallel tool results).

packages/ai-compaction/src/index.ts
```diff
@@ -0,0 +1,387 @@
...
+    options.strategyKey ??
+    (options.estimateTokens ? undefined : strategyKeys.get(strategy))
+  const checkpointStrategyKey = strategyKey
+    ? `${strategyKey}:maxTokens=${options.maxTokens}`
+    : undefined
+
+  return {
+    name: 'compaction',
+    optionalRequires: [MetadataCapability],
+    async onConfig(ctx, config) {
```
AlemTuzlak: **[bug]** `onConfig` runs on every phase, including `init`. The engine then builds the next config with `providerMessages: this.messages` (`packages/ai/src/activities/chat/index.ts:3955`), so the `init` provider rewrite is thrown away. The first `beforeModel` run starts from the full canonical list again. With no checkpoint (no metadata store, or `summarizeOldest` with no `strategyKey`), the strategy runs twice before the first adapter call. `onCompact` fires twice. `summarizeOldest` makes two summarize calls. With `[withCompaction, withPersistence]`, `init` can also compact the client list before persistence hydrates the thread.

**Suggestion:** Return at once when `ctx.phase === 'init'`. Compaction only needs `beforeModel` and `structuredOutput`. Add a test that `onCompact` / `summarize` fire once per model call with no metadata store.

docs/advanced/compaction.md
```diff
@@ -0,0 +1,220 @@
...
+| `summarizeOldest` | `summarize` (**required**), `keepRecentTokens`, `summaryRole` |
+| `clearToolResults` | `keepRecentToolResults` (default `3`), `stub` |
+
+The token count is a rough `characters / 4` estimate. It is good enough to trigger on, not exact. Pass `estimateTokens` for provider-accurate counts.
+
+## What it keeps safe
+
+- **The system prompt is never dropped.** `chat()` keeps it separate from `messages`, so compaction only touches the conversation.
+- **Tool calls stay paired with their results.** The built-in strategies never leave an orphaned tool result, so the request stays valid.
+- **It runs before every model call.** Compaction is incremental: as the chat keeps growing it compacts again, and a later `summarizeOldest` pass folds an earlier summary into the new one.
```
AlemTuzlak: **[suggestion]** This line says a later `summarizeOldest` pass folds an earlier summary into the new one. That is true only when a checkpoint is reused, so the next pass sees the previous summary plus new messages. `summarizeOldest` does not call `identifyStrategy` (`packages/ai-compaction/src/index.ts:209`), so it has no automatic key. Each `onConfig` without a key runs on the full canonical transcript and summarizes the original head again. The later persistence section states the key rule, but "What it keeps safe" states folding as a general guarantee.

**Suggestion:** Move the fold claim next to checkpoints. Say folding needs a metadata store and an explicit `strategyKey`. Give `summarizeOldest` a built-in key only for the non-callback options, or fail fast when persistence is on and no key is set.

packages/ai-compaction/src/index.ts
```diff
@@ -0,0 +1,387 @@
...
+}): CompactionStrategy {
+  return async (messages, ctx) => {
+    const keep = options.keepRecentTokens ?? Math.floor(ctx.maxTokens / 2)
+    const cut = splitAtRecent(messages, ctx.estimate, keep)
+    if (cut <= 0) return null
+    const summary = await options.summarize(messages.slice(0, cut))
+    return [
+      {
+        role: options.summaryRole ?? 'user',
+        content: `Summary of earlier conversation:\n${summary}`,
```
AlemTuzlak: **[suggestion]** The summary is injected as a `user` message: `Summary of earlier conversation:\n${summary}`. Dropped turns can contain prompt-injection text. That text goes to the summarizer, then comes back as user content on later calls. `summaryRole` can change the role, but the default is `user`, and the prefix does not mark the text as untrusted.

**Suggestion:** Default to a clearly delimited assistant note, or put a fixed instruction in `systemPrompts` and keep the summary body separate. Treat summarizer output as untrusted in the default prompt.

packages/ai-compaction/src/index.ts
```diff
@@ -0,0 +1,387 @@
...
+    /** Tokens of recent messages to keep verbatim. Default `floor(maxTokens/2)`. */
+    keepRecentTokens?: number
+    /** Build the marker that replaces the dropped head. */
+    marker?: (droppedCount: number) => string
+  } = {},
+): CompactionStrategy {
+  const strategy: CompactionStrategy = (messages, ctx) => {
+    const keep = options.keepRecentTokens ?? Math.floor(ctx.maxTokens / 2)
+    const cut = splitAtRecent(messages, ctx.estimate, keep)
+    // ponytail: can't shrink past the recent window; raise keepRecentTokens or
```
AlemTuzlak: **[nit]** The comment starts with `ponytail:` and restates the `cut <= 0` branch. That is a process note, not a reason the code exists.

**Suggestion:** Delete the `ponytail:` prefix. Keep one short line only if you still want the keep-window hint.

PR #1236, opened 2026-08-24
feat(ai-skills): add @tanstack/ai-skills — portable Agent Skills middleware

jherr:
`withSkills(...)` now gives any tool-calling model a skill catalog and a `load_skill` tool. The model loads a `SKILL.md` only when it needs it. No hosted sandbox.

TanStack AI DevTools now has a Skills tab. It shows the catalog from a `skills:state` stream chunk, and it marks a skill loaded when the model calls `load_skill`. Click the Skills `useChat` row and the detail pane opens. Before this PR that click did nothing, because the hook registered with an empty id. JSON cards around the tree grow with the object and cap at a max height. A small payload such as `{ "name": "pirate-speak" }` no longer sits in a tall empty box.

This also fixes review bugs: `skillDirectory` can load on Windows and by frontmatter name, `filter()` now hides `load`/`readResource`, filesystem resources return utf8 for `references/`, and catalog overflow throws `SkillLimitError`.

## Changes

- New package `@tanstack/ai-skills` with `withSkills`, `inlineSkill`, `skillDirectory` (`/node`), `staticSkills`, and combinators.
- `createResourceTool` exposes bundled files as `read_skill_resource`.
- Portable and hosted (native) skills refuse to combine in one call.
- Core exports `SkillLimitError`. Native factories throw it for hosted caps. Portable catalog overflow throws it too.
- DevTools Skills tab: catalog + loaded badges. Hook identity stays live after mount, so a `useChat` with no `threadId` is selectable. JSON cards size to content (max-height only).

## Checklist

## Release Impact

## Testing

**Commands run.** Package tests passed:

1. `pnpm --filter @tanstack/ai-skills test:lib` (61 tests)
2. `pnpm --filter @tanstack/ai-skills test:types`
3. `pnpm --filter @tanstack/ai-client test:lib -- tests/devtools.test.ts` (38 tests)
4. `pnpm --filter @tanstack/ai-devtools-core test:lib` (50 tests)

I did not run `pnpm test:pr`.

**Manual test.**

1. Start `testing/panel` (`pnpm dev` on a free port, for example 3020) and open `/skills`.
2. Open TanStack DevTools and click the Skills `useChat` row. The right pane must show the hook header and tabs.
3. Send a pirate-speak message. Wait for `load_skill`.
4. Open `{ "name": "pirate-speak" }`. The JSON card must hug that object. It must not leave a tall empty area under it.
5. Open the Skills tab. The catalog lists pirate-speak, haiku, emoji-storyteller, and the loaded skill has a badge.

**How this PR makes testing easy.**

- Unit tests in `packages/ai-skills/tests/` (parse, walk, combinators, `withSkills`, `skillDirectory`, load-skill).
- DevTools: `packages/ai-client/tests/devtools.test.ts` (`skills:state` CUSTOM chunk, plus a `useChat` with no `threadId` must register a generated `hookId`) and `packages/ai-devtools/tests/skills-registry.test.ts`.
- E2E: `testing/e2e/tests/portable-skills-wire.spec.ts`.
- Example: `/skills` in `testing/panel`.

## Risk / rollback

Low risk. The package is opt-in. Nothing loads skills unless you add `withSkills(...)`. The DevTools tab is empty when no `skills:state` chunk arrives. To roll back, revert the PR.

## Public API change

New export `withSkills` from `@tanstack/ai-skills`, plus `skillDirectory` from `@tanstack/ai-skills/node`. Caller usage:

**Before**

```ts
chat({ adapter, model, messages })
```

**After**

```ts
import { withSkills } from '@tanstack/ai-skills'
import { skillDirectory } from '@tanstack/ai-skills/node'

const skills = skillDirectory('./skills')

chat({
  adapter,
  model,
  messages,
  middleware: [withSkills(skills)],
})
```

Core `@tanstack/ai` also adds `SkillLimitError`. Portable catalog overflow throws it with `path: 'portable'`. DevTools users need no new calls: the Skills tab appears when `withSkills` is on the server.

## Summary by CodeRabbit

* **New Features**
  * Added portable Agent Skills support through `SKILL.md` files, with on-demand skill loading during chat.
  * Added skill sources for inline definitions, directories, build-time bundles, and custom stores.
  * Added resource access for bundled skill files.
  * Added a Skills view in DevTools showing available and loaded skills.
  * Added a Skills demo with chat, catalog inspection, and sample skills.
* **Bug Fixes**
  * Improved validation and structured errors for unsupported skill limits and invalid skill references.
* **Documentation**
  * Added guides covering portable skills, skill sources, custom adapters, and provider differences.


packages/ai-skills/src/node/index.ts
```diff
@@ -0,0 +1,216 @@
...
+    const map = new Map<string, string>()
+    for (const r of roots) {
+      const dirs = await walkSkillDirs(nodeLister, r, { maxDepth })
+      for (const d of dirs) map.set(d.name, d.dir)
+    }
+    return map
+  }
+
+  const dirOf = async (name: string): Promise<string> => {
+    const dir = (await scan()).get(name)
```
AlemTuzlak: **[bug]** `dirOf` / `load()` look up `scan()` keys, and those keys are `walkSkillDirs` folder names (`d.name`). `list()` returns `parseSkill(...).metadata.name` (frontmatter, with `dirName` from `node:path.basename`). Those strings are not the same when (a) the frontmatter `name` does not match the folder (allowed in `{ strict: false }`, which the panel uses), or (b) `d.name` is a Windows path. Then `load_skill` is advertised with name `alpha` and `load('alpha')` throws `no skill named "alpha"`.

**Suggestion:** Key the map by the parsed skill name, not the folder name. Build it in `list()`/`scan()` from `parseSkill` metadata. Add a test where folder `foo` has `name: bar` under `{ strict: false }`, and assert `load('bar')` works and `load('foo')` throws.

packages/ai-skills/src/walk.ts
```diff
@@ -0,0 +1,83 @@
...
+}
+
+export type ListDir = (dir: string) => Promise<Array<WalkEntry>>
+
+export const SKILL_FILE = 'SKILL.md'
+export const MAX_SKILL_WALK_DEPTH = 6
+const SKIP_DIR_NAMES = new Set(['.git', 'node_modules'])
+
+function basenameOf(path: string): string {
+  const segments = path.split('/').filter((segment) => segment !== '')
```
AlemTuzlak: **[bug]** `basenameOf` splits only on `/`. `skillDirectory`'s `nodeLister` uses `path.join`, so on Windows `entry.path` is `C:\proj\skills\alpha`. `basenameOf` then returns the whole path. `scan()` stores that as the map key. `list()` still reports `alpha` via `path.basename`. Result: `skillDirectory` cannot `load()` on Windows.

**Suggestion:** Split on both `/` and `\\`, or posix-normalize (`path.replace(/\\/g, '/')`) before taking the last segment. Add a unit test with Windows-style paths in the in-memory lister.

packages/ai-skills/src/node/index.ts
```diff
@@ -0,0 +1,216 @@
...
+      const files: Array<string> = []
+      for (const sub of RESOURCE_DIRS) {
+        files.push(...(await collectFiles(join(dir, sub), dir)))
+      }
+      return files
+    },
+    readResource: async (name, path) => {
+      assertSafeResourcePath(path)
+      const dir = await dirOf(name)
+      return readFile(join(dir, path))
```
AlemTuzlak: **[bug]** `readFile(join(dir, path))` has no encoding, so Node returns a `Buffer`. `createResourceTool` only treats `typeof value === 'string'` as utf8; a `Buffer` is a `Uint8Array`, so every filesystem resource (including `references/note.md`) is returned as `encoding: 'base64'`. `inlineSkill` / `staticSkills` return strings, so the same markdown is utf8 from those sources and base64 from `skillDirectory`.

**Suggestion:** For text resources, `readFile(..., 'utf8')`. For binary `assets/`, keep bytes. Test the tool output for a `.md` resource from `skillDirectory`, not only `source.readResource`.

packages/ai-skills/src/node/index.ts
```diff
@@ -0,0 +1,216 @@
...
+    load: async (name) => readFile(join(await dirOf(name), 'SKILL.md'), 'utf8'),
+    listResources: async (name) => {
+      const dir = await dirOf(name)
+      const files: Array<string> = []
+      for (const sub of RESOURCE_DIRS) {
+        files.push(...(await collectFiles(join(dir, sub), dir)))
+      }
+      return files
+    },
+    readResource: async (name, path) => {
```
AlemTuzlak: **[bug]** `assertSafeResourcePath` only rejects absolute paths, `..`, and `~`. `readResource` / `readScript` then `readFile(join(dir, path))` with no check that the file is under `references/` or `assets/`, no check against `listResources()`, and no `realpath` prefix check. The tool copy says resources come from `references/` or `assets/`; the code allows any relative file under the skill folder (`SKILL.md`, `scripts/`, `.env`).

**Suggestion:** After join, `realpath` both the skill dir and the target, and require the resolved path stays under the skill root. Also require the relative path to start with `references/` / `assets/` (and `scripts/` for `readScript`).

packages/ai-skills/src/combinators.ts
```diff
@@ -0,0 +1,179 @@
...
+/** Hide skills the predicate rejects. Filtered skills never reach the catalog. */
+export function filter(
+  source: SkillSource,
+  predicate: FilterPredicate,
+  ctx?: FilterContext,
+): SkillSource {
+  return {
+    ...source,
+    revision: forwardRevision(source),
+    list: async () => (await source.list()).filter((s) => predicate(s, ctx)),
```
AlemTuzlak: **[bug]** `filter()` only wraps `list()`. `load`, `listResources`, and `readResource` still come from `...source`. `createResourceTool` takes `skill: z.string()` (not the catalog enum `load_skill` uses) and does not check activation. A hidden skill is gone from the catalog, but `read_skill_resource({ skill: 'secret', path: 'references/keys.md' })` still works if the model guesses the name.

**Suggestion:** Wrap `load` / `listResources` / `readResource` / `listScripts` / `readScript` so a filtered-out name throws. Constrain `createResourceTool`'s `skill` to listed names, or reject names that are not in `source.list()`.

packages/ai-skills/src/combinators.ts
```diff
@@ -0,0 +1,179 @@
...
+      if (!listPromise || !fresh()) {
+        listAt = now()
+        listPromise = source.list().catch((err) => {
+          listPromise = undefined // don't cache failures
+          throw err
+        })
+      }
+      return listPromise
+    },
+    load: (name) => {
```
AlemTuzlak: **[bug]** `cache({ refreshInterval })` expires `list()` and never expires `load()`. After the interval, `list()` can show new metadata while `load()` still returns the first memoized body for that name. `readResource` is not cached at all, so the three tiers disagree.

**Suggestion:** Clear `loads` when `list()` refreshes. Test: mutate the underlying source after `refreshInterval`, then `load()` must return the new body.

packages/ai-skills/src/middleware.ts
```diff
@@ -0,0 +1,217 @@
...
+
+      // Catalog token cap (spec §4.2).
+      const limit = options.maxCatalogTokens ?? 4000
+      let catalog = (options.renderCatalog ?? renderCatalog)(skills, family)
+      if (estimateTokens(catalog) > limit) {
+        if (options.onLimitExceeded && options.onLimitExceeded !== 'error') {
+          skills = options.onLimitExceeded(skills, limit)
+          catalog = (options.renderCatalog ?? renderCatalog)(skills, family)
+        } else {
+          throw new Error(
```
AlemTuzlak: **[bug]** Catalog overflow throws a plain `Error`, not `SkillLimitError`. `SkillLimitErrorInit.path` includes `'portable'` and `@tanstack/ai-skills` re-exports the class, but the portable path never throws it. After `onLimitExceeded` reduces the list, the new catalog is not checked again, so a no-op reducer still sends an over-limit prompt.

**Suggestion:** Throw `SkillLimitError` with `path: 'portable'`. Re-run the cap after the reducer; if it still overflows, throw. Add a unit test for both the throw and the reducer re-check.

packages/ai-skills/src/catalog.ts
```diff
@@ -0,0 +1,39 @@
...
+import type { ModelFamily, SkillMetadata } from './types'
+
+/** Sort skills into a stable, cache-friendly order. */
+export function sortSkills(skills: Array<SkillMetadata>): Array<SkillMetadata> {
+  return [...skills].sort((a, b) =>
+    a.name < b.name ? -1 : a.name > b.name ? 1 : 0,
+  )
+}
+
+function escapeXml(s: string): string {
```
AlemTuzlak: **[suggestion]** `escapeXml` does not escape `"`. Names go into `name="${...}"`. `inlineSkill` does not validate `name` against `[a-z0-9-]`. A name like `foo"><skill name="evil` breaks the Anthropic catalog XML.

**Suggestion:** Escape `"` / `'` in attribute context (`&quot;`). Validate `inlineSkill` names the same way `parseSkill` does.

packages/ai-skills/src/middleware.ts
```diff
@@ -0,0 +1,217 @@
...
+    throw new Error(
+      '`instructionTemplate` must contain a `{skills}` placeholder',
+    )
+  }
+
+  return defineChatMiddleware({
+    name: 'skills',
+    provides: [SkillsCapability],
+
+    async setup(ctx) {
```
AlemTuzlak: **[suggestion]** There is no unit test file for `withSkills` or `createResourceTool`. Untested: catalog cap, `requireApproval`, `catalogPlacement: 'tool-description'`, empty source, idempotent second `onConfig`, native/portable refusal.

**Suggestion:** Add a middleware test that drives `setup` + `onConfig` with a fake ctx. Cover coexistence, empty catalog, overflow, and resource-tool detection. Add `createResourceTool` tests for utf8 vs base64, path rejection, and unknown skill names.

packages/ai-skills/src/types.ts
```diff
@@ -0,0 +1,82 @@
...
+  executable: false
+  reason?: 'no-runtime'
+}
+
+/** A source of skills. Bytes only — no filesystem assumption. */
+export interface SkillSource {
+  /** Stable identity for the current content. Used as a catalog cache key. */
+  revision?: () => Promise<string>
+
+  /** Tier 1. Called once per request; core memoizes on `revision()`. */
```
AlemTuzlak: **[suggestion]** `SkillSource.revision` is documented as used as a catalog cache key, and `list` is documented as memoized on `revision()`. `withSkills` never reads `revision()`. `cache()` also ignores it (time-based only). Docs tell authors to implement `revision()` so `withSkills` can cache the catalog. That contract is not implemented.

**Suggestion:** Either call `revision()` in `setup` / `cache()` and document the real policy, or stop claiming memoization on `revision()` in types and docs.

packages/ai-skills/src/testing/index.ts
```diff
@@ -0,0 +1,99 @@
...
+      if (!listResources || !readResource) return
+      const resources = await listResources('alpha')
+      expect(resources).toContain('references/note.md')
+      const value = await readResource('alpha', 'references/note.md')
+      // trimEnd: a file-backed source keeps the fixture's trailing newline (formatters add one); the payload is what matters.
+      expect(dec(value).trimEnd()).toBe('hello')
+      // Path traversal must be rejected — by the source or the shared guard.
+      await expect(
+        (async () => {
+          assertSafeResourcePath('../../etc/passwd')
```
AlemTuzlak: **[suggestion]** The path-traversal case calls `assertSafeResourcePath('../../etc/passwd')` before `readResource`. A source that does not reject traversal still passes, because the shared guard throws first. The writing-adapters doc says the suite proves the source rejects that path. It does not.

**Suggestion:** Call `readResource` first (and assert it rejects). Keep a separate unit test for `assertSafeResourcePath`.

packages/ai-skills/src/tools/load-skill.ts
```diff
@@ -0,0 +1,96 @@
...
+
+export interface LoadSkillDeps {
+  source: SkillSource
+  skills: Array<SkillMetadata>
+  /** per-conversation activation set (dedupe). */
+  activated: Set<string>
+  requireApproval?: boolean
+}
+
+export function createLoadSkillTool(deps: LoadSkillDeps): Tool {
```
AlemTuzlak: **[suggestion]** Loaded skills never show up in TanStack AI DevTools. The panel demo records activations with a one-off `onBeforeToolCall` observer and `/api/skills-inspect`. There is no skills panel, and `SkillsCapability.activated` is not surfaced.

**Suggestion:** Inject a `skills:state` CUSTOM chunk (same pattern as `memory:state`) with the catalog, re-emit it on the client event bus, and add a Skills tab in DevTools that shows available vs loaded skills.

packages/ai-skills/src/node/index.ts
```diff
@@ -0,0 +1,216 @@
...
+  options: { dir?: string; virtualId?: string; maxDepth?: number } = {},
+): SkillsCatalogPlugin {
+  const virtualId = options.virtualId ?? 'virtual:tanstack-skills'
+  const resolved = `\0${virtualId}`
+  const dir = options.dir ?? 'skills'
+  return {
+    name: 'tanstack-skills-catalog',
+    resolveId: (id) => (id === virtualId ? resolved : undefined),
+    load: async (id) => {
+      if (id !== resolved) return undefined
```
AlemTuzlak: **[suggestion]** `skillsCatalogPlugin` `load` is an arrow function, so it cannot use Vite's `this.addWatchFile`. Changing a `SKILL.md` during `vite dev` will not rebuild `virtual:tanstack-skills`.

**Suggestion:** Use a real Vite plugin method (`async load(id)`) and `this.addWatchFile` for every discovered `SKILL.md` (and resource file).

packages/ai-skills/src/node/index.ts
```diff
@@ -0,0 +1,216 @@
...
+): SkillSource {
+  const roots = Array.isArray(root) ? root : [root]
+  const { maxDepth, strict = true } = options
+
+  /** Fresh scan of every root → name → skill directory. */
+  const scan = async (): Promise<Map<string, string>> => {
+    const map = new Map<string, string>()
+    for (const r of roots) {
+      const dirs = await walkSkillDirs(nodeLister, r, { maxDepth })
+      for (const d of dirs) map.set(d.name, d.dir)
```
AlemTuzlak: **[nit]** `skillDirectory(['org', 'tenant'])` last-write-wins (`map.set` overwrites). `withSkills([org, tenant])` is `dedupe(aggregate(...))`, first-wins. Same intent, opposite collision rule.

**Suggestion:** Pick one rule and document it. Prefer first-wins to match `withSkills([a, b])`.

AlemTuzlak (approved):

PR #1246, opened 2026-08-26
fix(ai): keep first text delta after a tool call in the same turn

PatrM:
## 🎯 Changes

When an assistant turn opens with a tool call and then speaks, the first `TEXT_MESSAGE_CONTENT` delta of the post-tool text was dropped from the assembled message. `"Hello from the model."` rendered as `"from the model."`.

### Root cause

`TOOL_CALL_START` auto-creates the assistant message and sets `pendingManualMessageId`. The following `TEXT_MESSAGE_START` then takes the pending-message path (Case 1), which does not run the segment reset the existing-message path (Case 2) does, so `hasToolCallsSinceTextStart` stays set. The content handler's lazy reset was additionally gated on `previousSegment.length > 0`. With an empty prior segment that guard was false, so the stale flag survived to the second delta, whose reset wiped the already-accumulated first delta.

`getState().content` was correct throughout — only the assembled `UIMessage` TextPart lost the word. This breaks the `docs/chat-architecture.md` contract that every `TEXT_MESSAGE_CONTENT` delta appends.

### Fix

Run the segment reset on the first post-tool delta regardless of whether the prior segment is empty; gate only the flush of a non-empty prior segment on `previousSegment.length > 0`.

- Move the `length > 0` guard off the segment-reset condition onto the prior-segment flush.
- Add a unit test for the tool-then-text sequence in `stream-processor.test.ts`.
- Add a `text-tool-text` e2e case for a turn that opens with a tool call.

## ✅ Checklist

- [x] I have followed the steps in the Contributing guide.
- [x] I have tested code changes locally with `pnpm run test:pr`.
- [x] Docs: not user-facing — restores behavior already documented in `docs/chat-architecture.md`.
- [x] Changeset: added (`patch`, `@tanstack/ai`).

## 🚀 Release Impact

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where the first word of a response could be lost after a tool call.
  * Improved preservation of UI resources, structured output, tool results, metadata, and errors during streamed responses.
  * Improved handling of text transitions so repeated or empty segments are not emitted unnecessarily.

* **Tests**
  * Added coverage for tool-first conversations, interleaved streams, and UI-resource persistence.
  * Added end-to-end validation of text immediately following a tool call.

### DIFF
--- a/.changeset/fix-post-tool-first-text-delta.md
+++ b/.changeset/fix-post-tool-first-text-delta.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/ai': patch
+---
+
+Fix StreamProcessor dropping the first text delta of a post-tool segment. When an assistant turn opened with a tool call and then spoke, the message auto-created by `TOOL_CALL_START` left `hasToolCallsSinceTextStart` set through `TEXT_MESSAGE_START`, so the segment reset fired one delta late and folded away the first `TEXT_MESSAGE_CONTENT` chunk (e.g. losing the first word of the reply). The segment reset now runs on the first post-tool delta regardless of whether the prior segment was empty, and only the flush of a non-empty prior segment stays gated.
--- a/packages/ai/src/activities/chat/stream/processor.ts
+++ b/packages/ai/src/activities/chat/stream/processor.ts
@@ -1307,12 +1307,14 @@ export class StreamProcessor {
     // Detect if this is a NEW text segment (after tool calls) vs continuation
     const isNewSegment =
       state.hasToolCallsSinceTextStart &&
-      previousSegment.length > 0 &&
       this.isNewTextSegment(chunk, previousSegment)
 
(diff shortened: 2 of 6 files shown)


PatrM:
should coderabbits finding be handled here? seems like its outside the scope? Just let me know if I should consider it anyways 👍🏻

AlemTuzlak:
Thanks for this PR, and for tracing the dropped first text delta after a tool-first turn.

This is the same bug as https://github.com/TanStack/ai/pull/1248 (issue https://github.com/TanStack/ai/issues/1247). That PR is already merged. It resets the stale segment state on `TEXT_MESSAGE_START` Case 1, which is the path that was missing the reset.

I am closing this one so we do not land two patches for the same cause. Your test and write-up still helped confirm the bug.

Thank you for the work.

PR #1266, opened 2026-08-29
fix(openai-base): replay reasoning items on Responses tool loops

### LINKED ISSUE #1212: OpenAI Responses: tool loops intermittently 400 with "function_call ... provided without its required 'reasoning' item"
### TanStack AI version

@tanstack/ai@0.48.0, @tanstack/ai-openai@0.21.1

### Framework/Library version

node@24.19.0, zod@4.4.3 — server-side only, no UI framework involved

On the Responses API, a tool loop intermittently dies on the second turn with:

```
400 Item 'fc_04c5d379bb13126a006a8b1700e3ac87d19dae6257035321cf' of type 'function_call'
was provided without its required 'reasoning' item:
'rs_0bf74942a9c5d320006a8b0d57d26487d1981665bb013fd0f3'.
```

When a reasoning model emits a `reasoning` item alongside a `function_call`, the Responses API requires that reasoning item to be sent back with the `function_call` on the following turn. `OpenAITextAdapter` re-sends the `function_call` but not its paired `reasoning` item, so the request is rejected and the run ends with `RUN_ERROR` after the tools have already executed.

**It is intermittent, which is what makes it easy to miss** — it fails only on runs where the model actually produced a reasoning item. On `gpt-5.6-luna` with default effort I measured **4 failures in 6 consecutive runs** of the script below, unchanged between runs. My first attempt at a minimal reproduction passed twice and I nearly concluded the trigger was something else.

Not specific to parallel tool calls, large tool outputs, system prompts, or the number of registered tools — I bisected all four away. One tool and a one-line result reproduces it.

**Workaround:** `modelOptions: { reasoning: { effort: 'none' } }`. That suppresses the reasoning item entirely, so there is nothing to round-trip and the loop runs. Chat Completions is not an alternative — it rejects function tools alongside reasoning outright, with `"Function tools with reasoning_effort are not supported ... set reasoning_effort to 'none'"`.

This looks like the OpenAI counterpart of #340 ("Anthropic tool loops: thinking blocks missing on turn 2+"), which was fixed for Anthropic.

Run this a few times — expect roughly two failures in three. `npm i @tanstack/ai @tanstack/ai-openai zod`, set `OPENAI_API_KEY`, `npx tsx repro.mts`:

```ts
import { chat, maxIterations, toolDefinition } from '@tanstack/ai';
import { OpenAITextAdapter } from '@tanstack/ai-openai';
import { z } from 'zod';

const search = toolDefinition({
  name: 'search',
  description: 'Search passages of the archive.',
  inputSchema: z.object({ query: z.string() }),
  outputSchema: z.object({ passages: z.array(z.object({ id: z.string(), text: z.string() })) }),
}).server(({ query }) => ({ passages: [{ id: 'p1', text: `${query}: lorem ipsum dolor sit amet` }] }));

const stream = chat({
  adapter: new OpenAITextAdapter({ apiKey: process.env.OPENAI_API_KEY! }, 'gpt-5.6-luna'),
  messages: [{ role: 'user', content: 'Search the archive for anything about being tired, then summarise in one sentence.' }],
  tools: [search],
  agentLoopStrategy: maxIterations(6),
});

  if (chunk.type === 'RUN_ERROR') console.log('RUN_ERROR:', chunk.message);
  if (chunk.type === 'TEXT_MESSAGE_CONTENT') process.stdout.write(chunk.delta);
}
```

Expected: the loop completes and the model answers using the tool result.
Actual: about two runs in three end in `RUN_ERROR` with the 400 above, after the tool has run.

No

harshlocham:
On reasoning models, a Responses tool loop can fail with HTTP 400 after the tool runs. The second turn sends the `function_call` without its paired `reasoning` item (`rs_…`). This PR stores that item id on `thinking[].signature` and sends the reasoning item before the function call.

Fixes #1212

## 🎯 Changes

- Capture the Responses `reasoning` item id from `output_item.added`, `output_item.done`, and `response.completed`.
- Put that id on `STEP_FINISHED.signature` so `chat()` keeps it on `ModelMessage.thinking`.
- Replay `{ type: 'reasoning', id }` before `function_call` in `convertMessagesToInput`.
- Add a `chat()` unit test that drives two turns and asserts the second request includes `rs_required_1` before `fc_item_rs`.
- Docs stay the same. The public call site does not change.
- Patch changeset for `@tanstack/openai-base`.

## ✅ Checklist

## 🚀 Release Impact

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed multi-step tool interactions by preserving reasoning item IDs across turns.
  * Prevented errors when function calls require their associated reasoning item.
  * Ensured reasoning information is replayed in the correct order before related function calls.
* **Tests**
  * Added coverage verifying reasoning items and function calls are correctly replayed in subsequent requests.
* **Release**
  * Prepared a patch release for the OpenAI base integration.

### DIFF
--- a/.changeset/smart-crabs-fry.md
+++ b/.changeset/smart-crabs-fry.md
@@ -1,5 +1,5 @@
 ---
-"@tanstack/openai-base": patch
+'@tanstack/openai-base': patch
 ---
 
 Replay Responses reasoning item ids on the next tool-loop turn so function_call items keep their required reasoning partner.
--- a/.changeset/smart-crabs-fry.md
+++ b/.changeset/smart-crabs-fry.md
@@ -1,0 +1,5 @@
+---
+"@tanstack/openai-base": patch
+---
+
+Replay Responses reasoning item ids on the next tool-loop turn so function_call items keep their required reasoning partner.
--- a/packages/openai-base/src/adapters/responses-text.ts
+++ b/packages/openai-base/src/adapters/responses-text.ts
@@ -837,6 +837,14 @@ export abstract class OpenAIBaseResponsesTextAdapter<
     let hasEmittedTextMessageStart = false
     let reasoningMessageId: string | undefined
     let hasClosedReasoning = false
+    // Responses pairs a function_call with the preceding reasoning item.
+    // The next turn must send that item back or the API 400s (#1212).
+    let reasoningItemId: string | undefined
+    const captureReasoningItemId = (item: { type?: string; id?: string }) => {
+      if (item.type === 'reasoning' && item.id) {
+        reasoningItemId = item.id
+      }
+    }
     // Track whether we've emitted a terminal RUN_FINISHED so the
     // end-of-stream fallback below knows to synthesise one when the upstream
     // cuts off without a response.completed event.
@@ -899,12 +907,14 @@ export abstract class OpenAIBaseResponsesTextAdapter<
           model: currentModel,
           timestamp,
           content: accumulatedReasoning,
+          ...(reasoningItemId && { signature: reasoningItemId }),
         }
       }
       reasoningMessageId = undefined
       stepId = null
       hasClosedReasoning = false
       accumulatedReasoning = ''
+      reasoningItemId = undefined
     }
 
     const emitReasoningDelta = function* (
@@ -997,6 +1007,7 @@ export abstract class OpenAIBaseResponsesTextAdapter<
           hasEmittedTextMessageStart = false
           reasoningMessageId = undefined
           hasClosedReasoning = false
+          reasoningItemId = undefined
           stepId = null
           accumulatedContent = ''
           accumulatedReasoning = ''
@@ -1224,6 +1235,7 @@ export abstract class OpenAIBaseResponsesTextAdapter<
         // handle output_item.added to capture function call metadata (name)
         if (chunk.type === 'response.output_item.added') {
           const item = chunk.item
+          captureReasoningItemId(item)
           if (item.type === 'function_call' && item.id) {
             // Track the item as soon as we see it so subsequent arg deltas
             // aren't logged as orphans, but only emit TOOL_CALL_START when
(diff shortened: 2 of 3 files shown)


AlemTuzlak:

**Verdict:** reject
**Head SHA:** cc0fa5bff5bfa5c7158b961b74d2d2568a284c0a
**Label:** `ai-rejected`

**Findings**
- [bug] packages/openai-base/src/adapters/responses-text.ts:886 The claimed #1212 400 happens whenever a Responses `reasoning` item (`rs_…`) exists next to a stored `function_call` id. This PR only puts that id on STEP_FINISHED.signature after `openReasoning()`, which runs only on reasoning text/summary deltas. `closeReasoning()` no-ops when `reasoningMessageId` is unset, and `chat()` also drops a signature unless `currentThinkingContent` is non-empty. Default reasoning-model streams often emit `output_item` type `reasoning` with an empty summary and no `reasoning_text.delta`. An agent-written two-turn repro failed on clean main for both streamed-text and empty-item cases; on this PR the streamed-text case passed and the empty-item case still omitted `rs_probe_empty` from the second `input`. That empty-item path is the reported bug, so the root cause is not fixed.
- [suggestion] packages/openai-base/src/adapters/responses-text.ts:845 `captureReasoningItemId` keeps a single `reasoningItemId` and overwrites it. A response with more than one reasoning item collapses to the last id, so earlier required `rs_…` partners can still be missing on the next turn.
- [nit] packages/openai-base/tests/responses-text.test.ts:2582 The new test injects `response.reasoning_text.delta`, so it only locks the path that already works and never fails when the reasoning item has no text.

**Push**
Did not push.

tombeckenham:
Superseded by #1290.

This PR only round-trips the reasoning item id. Default gpt-5.x tool loops still 400 when the reasoning item has no streamed text, and they also need `encrypted_content`. #1290 covers those cases (id + encrypted blob, empty thinking, gated default `include`).

PR #1270, opened 2026-08-30
fix(ai-bedrock): use catalog Mantle path for Gemma 4

### LINKED ISSUE #925: @tanstack/ai-bedrock: `bedrock-mantle` base URL is hardcoded to `/v1`, but the path is per-model (`/openai/v1`, `/anthropic/v1/messages`, `/v1`) — Gemma/Claude 401

harshlocham:
On Bedrock Mantle, Gemma 4 is served at `/openai/v1`. The adapter always used `/v1`, so the request failed with a misleading 401. This PR reads `mantlePath` from the Bedrock compatibility config. `google.gemma-4-*` uses `/openai/v1`. Other models use `/v1`. An explicit `baseURL` still wins. Gemma 4 is chat and responses only, so the factory no longer treats every catalog id as a Converse model.

Fixes #925

## 🎯 Changes

- The compatibility config supplies `mantlePath`. Chat and responses adapters pass the model id into `withBedrockDefaults`.
- Gemma 4 ids are in the catalog (`google.gemma-4-31b`, `google.gemma-4-26b-a4b`, `google.gemma-4-e2b`).
- `bedrockText('google.gemma-4-31b', { api: 'chat', endpoint: 'mantle' })` constructs. Default Converse throws for chat-only models.

Claude on Mantle (`/anthropic/v1/messages`) is a different API. This PR does not add it.

## ✅ Checklist

## 🚀 Release Impact

## Root cause

**Issue.** Mantle Gemma 4 callers hit `/v1` and get 401 `access_denied`. `bedrockText('google.gemma-4-31b', { api: 'chat' })` also failed the chat-model check because Gemma 4 was not in the catalog.

**Cause.** `buildBaseURL` used a fixed Mantle path and never saw the model id. The catalog had no Gemma 4 row.

**Fix.** Compatibility rules supply `mantlePath`. Adapters pass the model id. The factory accepts chat-only ids when `api` is `chat` or `responses`.

## Possible alternatives

- **Prefix check in `client.ts`.** `startsWith('google.gemma-4-')` is a second model table. The catalog is the table.
- **Always `/openai/v1` on Mantle.** Gemma 3 and DeepSeek use `/v1` on the model cards.
- **Caller `baseURL` only.** That is the workaround. It does not fix the default.

## Testing

**Commands run.** `pnpm --filter @tanstack/ai-bedrock test:lib` (111 passed). `pnpm test:pr` was not run.

**Gate 1 (this session).** Agent-written `tsx` import of `withBedrockDefaults({ apiKey: 'k', region: 'eu-central-1', endpoint: 'mantle' }, undefined, 'google.gemma-4-31b')`.

1. On pinned `main` (`b76786252`): `baseURL= https://bedrock-mantle.eu-central-1.api.aws/v1` then `FAIL expected .../openai/v1` (exit 1).
2. On this PR (`b735c382e`): `baseURL= https://bedrock-mantle.eu-central-1.api.aws/openai/v1` then `PASS` (exit 0).

**Manual test.**

1. On main, `withBedrockDefaults({ apiKey: 'k', region: 'eu-central-1', endpoint: 'mantle' }, undefined, 'google.gemma-4-31b')` returns `.../v1`.
2. On this branch it returns `.../openai/v1`.
3. `createBedrockText('google.gemma-4-31b', 'k', { api: 'chat', endpoint: 'mantle' })` constructs a chat adapter.

**How this PR makes testing easy.** Unit tests cover config `mantlePath`, adapter `client.baseURL` for Gemma 4, and the factory chat/responses path.

## Risk / rollback

Wrong `mantlePath` on a catalog row sends OpenAI-compatible Mantle traffic to the wrong path (401). Revert the PR to undo.

## Public API change

**Before**

```ts
bedrockText('google.gemma-4-31b', { api: 'chat', endpoint: 'mantle' })
// throws: not available on Chat Completions
// if forced through, POST .../v1/chat/completions → 401
```

**After**

```ts
bedrockText('google.gemma-4-31b', {
  api: 'chat',
  endpoint: 'mantle',
  region: 'eu-central-1',
})
// POST https://bedrock-mantle.eu-central-1.api.aws/openai/v1/chat/completions
```


packages/ai-bedrock/src/utils/client.ts
```diff
@@ -27,16 +27,28 @@ const DEFAULT_REGION = 'us-east-1'
 /** OpenAI SDK requires a non-empty apiKey even when a signed fetch overrides Authorization. */
 const SIGV4_PLACEHOLDER_KEY = 'bedrock-sigv4'
 
-function buildBaseURL(region: string, endpoint: BedrockEndpoint): string {
+/** Gemma on mantle is served at `/openai/v1`, not `/v1` (AWS model card). */
+function mantlePathForModel(model: string | undefined): string {
+  return typeof model === 'string' && model.startsWith('google.gemma-')
```
jsve: this will not work for the gen 3 gemma models, for example (https://docs.aws.amazon.com/bedrock/latest/userguide/model-card-google-gemma-3-12b-it.html)

harshlocham: @jsve 
Good catch — you’re right. The current google.gemma- prefix also matches Gemma 3, which should remain on /v1.

I’ll narrow the matcher to google.gemma-4- and add a regression test covering google.gemma-3-12b-it → /v1 alongside the Gemma 4 cases → /openai/v1.

tombeckenham (approved):
Looks good. Gemma 4 on Mantle now hits `/openai/v1`. Other models stay on `/v1`. Explicit `baseURL` still wins. Chat-only catalog ids construct on `api: 'chat'` / `api: 'responses'` and throw on default Converse.

AlemTuzlak (approved):

Issue #1247, opened 2026-08-26
StreamProcessor drops the first TEXT_MESSAGE_CONTENT delta when a tool call precedes any text in the same message


ramigs:
### TanStack AI version

0.49.1

### Framework/Library version

N/A — not framework-specific. This is a bug in @tanstack/ai's StreamProcessor itself; reproduces with zero framework involved (see minimal reproduction link below). Originally surfaced via @tanstack/ai-vue 0.19.1, but that's incidental — @tanstack/ai-client: 0.28.0.
Node.js v22.22.3 used to run the standalone repro.

### Describe the bug and the steps to reproduce it

When a `TOOL_CALL_START` event's `parentMessageId` references an assistant message that hasn't had a `TEXT_MESSAGE_START` yet — the normal AG-UI shape for "the agent calls a tool, then explains the result, all as one assistant turn" — the **first** `TEXT_MESSAGE_CONTENT` delta of the text that follows the tool call is silently dropped from the rendered message. Every delta after the first renders correctly; only the first chunk goes missing.

This isn't an edge case — "tool call(s), then a final answer, all as one assistant turn" is one of the most common AG-UI event shapes, and it's exactly the shape any UI built to visualize live tool-call activity needs to handle.

## Root cause

In `StreamProcessor` (`packages/ai/src/activities/chat/stream/processor.ts`):

1. `TOOL_CALL_START` arrives with `parentMessageId: "msg-1"`, but no `TEXT_MESSAGE_START` for `"msg-1"` has been seen yet.
2. `handleToolCallStartEvent` calls `ensureAssistantMessage("msg-1")`. Since no message/state exists for that id yet, it falls through to the "backward-compat auto-create" branch (intended for callers using `process()` without ever sending `TEXT_MESSAGE_START`), which sets `this.pendingManualMessageId = "msg-1"`.
3. `handleToolCallStartEvent` then sets `state.hasToolCallsSinceTextStart = true` on that message's state.
4. Later, the **real** `TEXT_MESSAGE_START` for `"msg-1"` arrives (the text that follows the tool call). Because `pendingManualMessageId` is still set to `"msg-1"`, `handleTextMessageStartEvent` takes its **Case 1** branch ("a manual message was created via `startAssistantMessage()`") instead of **Case 2** ("message already exists"). Case 1 just clears `pendingManualMessageId` and returns.
[...]

PR #1271, opened 2026-08-30
 fix(ai): keep the first text delta when the text starts before the tool call

iRanadheer:
## 🎯 Changes

Follow-up to #1247. #1248 fixed the order where the tool call precedes the message's `TEXT_MESSAGE_START`. The first delta is still dropped in the other order, where the text starts first, because `hasToolCallsSinceTextStart` is cleared on the start paths and never once a segment has begun.

The first delta after a tool call cannot enter the `isNewSegment` branch, since `previousSegment` is empty at that point. The flag survives into the second delta, which then matches all three conditions: the accumulation resets and `updateTextPart` writes the remainder over the text part the first delta had created. Multi-segment behaviour is unaffected, since a later tool call sets the flag again in `handleToolCallStartEvent`.

Reproduces on 0.52.0 and on main.

What is in it:

- `packages/ai/src/activities/chat/stream/processor.ts`: clear the flag once a segment has started.
- `packages/ai/tests/stream-processor.test.ts`: a sibling to the test added with #1248, same fixture, opposite event order.
- `testing/e2e/`: a harness mirroring the one added with #1248, for the text-first order.
- A changeset, patch.

Verified locally:

- The new unit test fails on unmodified main (`'sunny.'` instead of `'It is sunny.'`) and passes with the change.
- The new spec fails in Chromium against a build without the fix (`'world.'` instead of `'Hello, world.'`) and passes with it.
- `pnpm run test:pr` passes across all 82 projects.
- Both #1247 specs pass under `playwright test`.

## ✅ Checklist

## 🚀 Release Impact

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed streamed assistant responses after tool calls so the first text segment is preserved when content arrives in multiple chunks.
  * Prevented the beginning of a response from being dropped or overwritten.

* **Tests**
  * Added regression and end-to-end coverage for text-first tool-call streaming scenarios.

### DIFF
--- a/.changeset/text-first-tool-drop.md
+++ b/.changeset/text-first-tool-drop.md
@@ -1,0 +1,6 @@
+---
+'@tanstack/ai': patch
+---
+
+Keep the first text delta when a message speaks after a tool call and the text
+arrives in more than one chunk.
--- a/packages/ai/src/activities/chat/stream/processor.ts
+++ b/packages/ai/src/activities/chat/stream/processor.ts
@@ -1357,6 +1357,12 @@ export class StreamProcessor {
       state.hasToolCallsSinceTextStart = false
     }
 
+    // A segment begins at its first delta, not its second. Left set, the flag
+    // makes the second delta look like a further segment: the accumulation
+    // resets and updateTextPart writes over the part the first delta created.
+    // A later tool call sets it again in handleToolCallStartEvent.
+    state.hasToolCallsSinceTextStart = false
+
     const currentText = state.currentSegmentText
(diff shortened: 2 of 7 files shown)


AlemTuzlak:

**Verdict:** ready
**Head SHA:** 0ae844dd59b75eb47ab5588e93584a47cc4ccfba
**Label:** `ai-ready`

**Findings**
- None

**Push**
Did not push.

AlemTuzlak (approved):

PR #1290, opened 2026-09-01
fix(ai): replay OpenAI reasoning items on tool follow-up

### LINKED ISSUE #1212: OpenAI Responses: tool loops intermittently 400 with "function_call ... provided without its required 'reasoning' item"

AlemTuzlak:
A gpt-5.x tool follow-up returns HTTP 400. OpenAI requires the paired `reasoning` item (`rs_*`) with each `function_call`. This PR stores the item id and encrypted blob on `thinking[].signature`. Then it sends that item on the next turn.

Default `include: ['reasoning.encrypted_content']` applies only to reasoning models (`o*`, `gpt-5*` except `*-chat-latest`, `codex-mini-latest`). Pre-5 chat models such as `gpt-4o-mini` keep the old request shape.

When `response.completed` carries that blob after `output_text` already closed the step, the adapter attaches the signature to the existing thinking step. It does not open a second empty thinking part.

Fixes #1212

## 🎯 Changes

- Capture Responses `reasoning` id and `encrypted_content`. Store both on `thinking[].signature`.
- Replay `{ type: 'reasoning', id, encrypted_content }` before `function_call` on the next turn.
- Keep empty thinking text when a signature is present (wire, UI parts, snapshot).
- `openaiText` sets `include: ['reasoning.encrypted_content']` only for reasoning models, and only when the caller does not pass `include`.
- After streamed reasoning text, attach a late encrypted blob to the same thinking step. Do not emit a second empty thinking part.
- Fill e2e generate prompts after hydration so `generate-button` is enabled before click.
- Give `@tanstack/ai-client` tests a 15s timeout so resume-join waits do not lose the default 5s race under `test:pr` load.
- Patch changeset for `@tanstack/ai`, `@tanstack/openai-base`, and `@tanstack/ai-openai`.

Related: #1266 stores the reasoning id only. This PR also keeps encrypted content, empty thinking, and the gated default `include`.

## ✅ Checklist

The docs checkbox is the skill reference at `packages/ai/skills/ai-core/adapter-configuration/references/openai-adapter.md`. There is no new `docs/` page.

## 🚀 Release Impact

## Root cause

**Issue.** A gpt-5.x tool loop dies with HTTP 400: `function_call` without its required `reasoning` item.

**Cause.** Four drops on the 400 path:

1. `convertMessagesToInput` omits reasoning items.
2. `convertMessagesToModelMessages` drops empty `role:reasoning` rows that only have `encryptedValue`.
3. `finalizeCurrentThinkingStep` drops empty thinking text even when a signature exists.
4. `openaiText` does not request `reasoning.encrypted_content` on reasoning models.

CI `Test` then failed `@tanstack/ai-client:test:lib`. Nx Cloud hid the vitest log. Locally 740 tests passed. Resume-join tests wait 2s inside `waitFor({ timeout: 5_000 })` under Vitest's default 5s `testTimeout`, which loses under parallel `test:pr` load.

**Fix.** Pack id plus encrypted blob into `thinking[].signature`. Keep empty thinking when a signature is present. Replay a `reasoning` item before `function_call`. Default `include` only when `openAIModelRejectsSamplingParams(model)` is true. Set ai-client `testTimeout` to 15s.

## Possible alternatives

- **previous_response_id.** Send the OpenAI response id on the next turn. This PR did not take it because the SDK is stateless per request and the client already sends full history.
- **New public thinking.id field.** This PR did not take it because `thinking.signature` already exists and JSON packing avoids a public type change.
- **Require callers to set include.** This PR did not take it because gpt-5.x tool follow-up fails unless the encrypted blob is present.
- **Re-run CI only for the ai-client failure.** This PR did not take it. The 5s vs 5s timeout race is deterministic under load.

## Testing

**Commands run**

- `pnpm --dir packages/ai-client exec vitest run`: 740 passed (after merge of `origin/main` `924d45e4`).
- `pnpm --dir packages/ai-client exec vitest run tests/resume-snapshot.test.ts`: 20 passed.
- Latest CI: E2E passed. `Test` failed `@tanstack/ai-client:test:lib` (1m 17s). Nx printed no vitest assertion.
- `pnpm test:pr`: not run locally on this follow-up.

**Manual test**

1. On main, convert a wire history with empty `role:reasoning` plus `encryptedValue` and a `function_call`. `thinking` is missing.
2. On this branch, run the same conversion. `thinking[].signature` is present. `convertMessagesToInput` emits `type: 'reasoning'` before `function_call`.
3. Call `openaiText('gpt-4o-mini')` with no `include`. The request must not set `include`.
4. Stream reasoning text, then assistant text, then a completed output that includes the reasoning item. The stream must emit one `REASONING_START`.

Unit tests in `packages/ai/tests/messages.test.ts`, `packages/ai/tests/ag-ui-wire.test.ts`, `packages/openai-base/tests/responses-text.test.ts`, and `packages/ai-openai/tests/openai-adapter.test.ts`.

## Public API change

**Before**

```ts
chat({
  adapter: openaiText('gpt-5.6'),
  messages,
  modelOptions: { include: ['reasoning.encrypted_content'] },
})
```

**After**

```ts
// Reasoning models only. gpt-4o-mini is unchanged.
chat({
  adapter: openaiText('gpt-5.6'),
  messages,
})

// Opt out on a reasoning model:
chat({
  adapter: openaiText('gpt-5.6'),
  messages,
  modelOptions: { include: [] },
})
```

## Risk / rollback

Reasoning models request extra reasoning bytes by default. Revert the PR, or set `include: []`, to undo that.

Pre-5 chat models do not get a new `include` default.

Old transcripts with no `thinking[].signature` still cannot replay gpt-5.x tools.


harshlocham (commented):
I checked this locally against the original #1212 case. This looks like a much more complete fix than #1266, especially with the encrypted reasoning content and empty reasoning case.

One thing I noticed while tracing the flow: captureReasoningItem() only keeps a single reasoning item, so if a response contains multiple reasoning items, the previous one gets overwritten. Also, convertMessagesToInput() currently puts all reasoning items before the function calls, so something like reasoning A → function_call A → reasoning B → function_call B wouldn’t be replayed in the same order.

I don’t think this needs to block the #1212 fix, but it seems worth having a regression test for multiple/interleaved reasoning items (or documenting the expected invariant if this can’t happen in this flow).

harshlocham:
I also checked the failing reasoning.spec.ts E2E.

It looks like this is now producing two thinking-blocks: the expected reasoning block and another empty one that only has the signature. This seems to come from keeping signature-only thinking through the message/UI conversion.

I don’t think we should just update the E2E to expect two blocks, since the second one has no visible reasoning content. Could we keep the signature for the follow-up request without exposing an empty thinking block in the UI?

The other E2E failures I looked at (multi-turn-structured and durable-takeover) seem unrelated to this PR.

tombeckenham (approved):

PR #1346, opened 2026-09-08
fix(ai-client): keep the reply when a rejoined run reasons before it answers

AlexDemzz:
## 🎯 Changes

`ChatClient.resumeInFlightRun` drops the hydrated in-flight assistant message the first time a chunk in `REJOIN_REBUILD_TRIGGERS` arrives, so the replay can rebuild it cleanly. That set only held `TEXT_MESSAGE_START`, `TEXT_MESSAGE_CONTENT`, `TOOL_CALL_START` and `MESSAGES_SNAPSHOT`.

A model that reasons before it answers (OpenAI gpt-5.x, Anthropic with extended thinking) opens its run with `REASONING_MESSAGE_START`, `STEP_STARTED`, `STEP_FINISHED` and `REASONING_ENCRYPTED_VALUE` before any text. The stream processor answers those with `ensureAssistantMessage()`, so the assistant message already exists when `TEXT_MESSAGE_START` arrives. The drop then fires on that first text chunk, removes the message from `processor.messages` through `setMessages()`, and leaves the processor's `messageStates` and `activeMessageIds` pointing at it. Every text chunk after that lands on a message the transcript no longer holds.

What the user sees on a `persistence: true` mount while a run streams (a reload mid-stream, or a route change that remounts the chat): the user message, `isLoading` true for the whole run, then nothing. No assistant message, and `onFinish` runs against a transcript without the reply. The server side is fine, the whole log replays; it is only the client that discards it. Verified against a real durability log where the client received every chunk through `onChunk` while `messages` never changed.

Fix: add `REASONING_MESSAGE_CONTENT`, `REASONING_ENCRYPTED_VALUE` and `STEP_FINISHED` to `REJOIN_REBUILD_TRIGGERS`, which are the other chunk types the processor creates the assistant message on. The drop then runs before any assistant message exists on a rejoin, and a hydrated partial is still rebuilt from the first content chunk as before.

Test: `packages/ai-client/tests/chat-client-join-run-reasoning.test.ts` rejoins a run whose log opens with reasoning chunks through a fake connection and asserts the reply is present. It fails on `main` (assistant text is empty, one message) and passes with this change.

No docs change: the behaviour matches what the rejoin docs already describe.

## ✅ Checklist

## 🚀 Release Impact

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where rejoining an AI run could lose the assistant’s reply when reasoning occurred before text generation.
  * Assistant messages now remain intact when a run is resumed after reasoning or completed-step events.

### DIFF
--- a/.changeset/rejoin-reasoning-first-reply.md
+++ b/.changeset/rejoin-reasoning-first-reply.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/ai-client': patch
+---
+
+Fix a `joinRun` rejoin losing the whole reply when the model reasons before it answers. The `REASONING_*` and `STEP_FINISHED` chunks that open a run already create the assistant message; the rebuild drop then ran on the later `TEXT_MESSAGE_START`, removed that message from the transcript while the processor kept its state, and every text chunk after it went to a message the UI no longer held. The transcript stayed on the user message with no reply and no `onFinish`. Those chunk types now count as rebuild triggers, so the drop runs before any assistant message exists.
--- a/packages/ai-client/src/chat-client.ts
+++ b/packages/ai-client/src/chat-client.ts
@@ -322,12 +322,22 @@ const REJOIN_CONNECT_DEADLINE_MS = 2000
  * in-flight partial is dropped only when one of these arrives — never on
  * `RUN_STARTED` alone — so a rejoin that connects but delivers no content cannot
  * leave an empty assistant bubble.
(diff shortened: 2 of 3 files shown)


AlexDemzz:
Closing: on our side the client no longer mounts while a run streams, so this path is not what we need any more. The change and the test are still valid if a maintainer wants to pick them up.

PR #1355, opened 2026-09-09
feat(ai): flush CUSTOM events immediately through durability

oztune:
CUSTOM progress events now reach the client as soon as they are emitted. A slow compaction or tool can show `started` while the work is still running.

Durability still batches high-volume output: `process.stdout`, `process.stderr`, `sandbox.file`, and `sandbox.file.diff`. Pass `{ batch: true }` on `emitCustomEvent` to keep one of your own events in that batch.

## Changes

`compaction:started` and other `emitCustomEvent` progress used to sit in the durability batch until 32 chunks, `RUN_FINISHED`, or `TOOL_CALL_END`. They then arrived together with later output.

This PR makes `CUSTOM` a durability flush boundary. The engine already yielded the chunk at emit time. Durability now appends and forwards it on that same pull.

High-volume adapter events stay batched by name. `{ batch: true }` is the opt-in for app events that should share appends. The hint is stripped before the log and the wire.

## Checklist

- [x] Followed the Contributing guide.
- [ ] Tested locally with `pnpm run test:pr`. I ran package tests instead (see Testing).
- [x] I understand all the code here.
- [x] Updated docs.
- [x] Changeset added (`@tanstack/ai` minor, `@tanstack/ai-sandbox` patch).

## Release impact

- [x] Affects published code; changeset added.

## Testing

**Commands run**

1. `oxfmt --check` on the changed files
2. `oxlint src --type-aware` for `@tanstack/ai` and `@tanstack/ai-sandbox`
3. `tsc --noEmit` for both packages
4. `vitest run tests/stream-to-response-durability.test.ts tests/strip-to-spec-middleware.test.ts` in `@tanstack/ai` (32 passed)
5. sandbox `tool-bridge` and `bridge-jsonrpc` tests (17 passed)
6. `pnpm --filter @tanstack/ai --filter @tanstack/ai-sandbox run build`

I did not run full `pnpm run test:pr`.

**Manual test**

1. Run a chat with `withCompaction` and durability SSE.
2. Use a slow `summarizeOldest` (a few seconds).
3. Confirm `compaction:started` arrives on the client before the summarize promise returns.
4. Confirm `compaction:ended` arrives when the strategy returns.
5. For a tool that emits `emitCustomEvent('progress', { step })` three times, confirm three separate durability appends.

`packages/ai/tests/stream-to-response-durability.test.ts` covers:

- a progress `CUSTOM` event flushing alone before model text
- `process.stdout` staying in the batch with later text
- `{ batch: true }` staying in the batch, with the hint stripped
- a middleware event appended while `onConfig` is still waiting
- three tool progress events in three separate appends

## Risk / rollback

High-volume `CUSTOM` events that are not in the batched-name list will flush on every chunk (more durability `append` calls). Revert this PR to restore the old batch-until-32 behavior.

## Public API change

**Before**

```ts
ctx.emitCustomEvent("job:started", { step: "prepare" });
```

A pre-model event waited in the durability batch and often arrived with later output.

**After**

```ts
ctx.emitCustomEvent("job:started", { step: "prepare" });
// flushes immediately

ctx.emitCustomEvent("noisy", { n }, { batch: true });
// stays in the durability batch
```

## Summary by CodeRabbit

- **New Features**
  - Custom events now reach clients immediately by default, improving visibility of live progress.
  - Added an optional batching setting for high-volume custom events and events that should remain grouped.
  - Tool, middleware, and sandbox event APIs support the batching option.

- **Documentation**
  - Updated guidance and API references to explain immediate flushing, batching behavior, and progress-event delivery.

- **Tests**
  - Added coverage for immediate delivery, grouped events, middleware activity, and tool progress updates.


oztune:
Hey let me know if there's anything I can do to help get this reviewed

packages/ai/src/types.ts
```diff
@@ -629,6 +639,26 @@ type RuntimeContextField<TContext> =
...
+   * fills or a flush boundary (`RUN_STARTED` / `RUN_FINISHED` / `TOOL_CALL_END`)
+   * fires, so a low-volume progress event emitted before the model produces
+   * output — a compaction/summarize pass, a sandbox boot, a retrieval step —
+   * ships bunched with later chunks instead of at emit time. Marking it `flush`
+   * makes the batching producer flush it on its own, so a live progress
+   * indicator can render. Off by default: high-volume events (e.g.
+   * `process.stdout`) should keep batching, so opt in per event rather than
+   * flushing every `CUSTOM` chunk.
+   */
+  flush?: boolean
```
AlemTuzlak: **[suggestion]** The new public `{ flush: true }` option encodes the wrong default. `emitCustomEvent` already yields from `chat()` at emit time (`drainMiddlewareCustomQueue` / `runWhileYielding` / tool `pendingEvents`). The durability producer in `packages/ai/src/stream-to-response.ts` is what holds an unmarked `CUSTOM` chunk: `isDurabilityFlushBoundary` is only `RUN_STARTED` / `RUN_FINISHED` / `RUN_ERROR` / `TOOL_CALL_END`, plus this new `metadata.tanstack.flush === true` check. Default batch size is 32. A pre-model event therefore sits in the batch until a later terminal or tool-call-end (or 32 chunks), then ships together with later output. That is why `compaction:started` needed the flag.

The flag is not required for the mechanism to work, and it is the wrong knob for this product:
1. Compaction, sandbox boot, retrieval, and tool progress need to reach the client at emit time. Batching those events defeats the live indicator. The library should flush them without a per-call opt-out.
2. Making `CUSTOM` a durability flush boundary, or flushing every `emitCustomEvent` result, removes `EmitCustomEventOptions`, `TanStackRunMetadata.flush`, and the compaction `{ flush: true }` calls.
3. The stated reason for opt-out is high-volume `process.stdout`. That name is a well-known adapter CUSTOM event (`CUSTOM_EVENT.PROCESS_STDOUT`). It is not produced by `emitCustomEvent`. Harness adapters yield those chunks on their own stream. Flushing every `emitCustomEvent` still lets `process.stdout` / `process.stderr` / `file.changed` batch. Flushing every `CUSTOM` type would turn each stdout chunk into its own `append()` (more durability I/O). If that cost matters, opt *in* to batch on those high-volume names, or pass `{ batch: true }` only on the emit path. Do not make every progress caller learn a hidden durability batch.
4. Riding the hint in `metadata.tanstack.flush` leaks a transport concern into the persisted run log and the wire. `toWireChunk` keeps `metadata` (it is a spec key). Replay does not read `isDurabilityFlushBoundary`, so the field is unused after the live producer. The client ignores it (`onCustomEvent` gets name, value, toolCallId). It still sits on stored chunks next to real TanStack extras (`model`, `finishReason`, `usage`, interrupts).
5. Other `emitCustomEvent` callers stay stuck in the batch if they do not set the flag: `sandbox.file` / `sandbox.file.diff` (`createCustomEventChunk` without options), `ui-resource`, `code_mode:*`, docs examples, and bridged tools. `createBridgeEventChannel` / `ToolBridgeCoreOptions.emitCustomEvent` only take `(name, value)` and drop a third argument, so `{ flush: true }` on a harness-bridged tool is silently ignored.
6. With `{ flush: true }`, compaction:started *does* flush while `summarizeOldest` awaits. This is a default/API problem, not a failed flush path.

**Suggestion:** Drop the public option. Prefer one of these, in order:
1. Treat `chunk.type === 'CUSTOM'` as a durability flush boundary (simplest; measure `process.stdout` append volume if that path is hot).
2. Flush every chunk from `createCustomEventChunk` / `emitCustomEvent`, and keep batching only for adapter-yielded CUSTOM (stdout, file watchers).
3. If high-volume emitters must batch, add opt-in `{ batch: true }` (or an internal list of names), not opt-out `{ flush: true }`.

Do not put a delivery hint on `metadata.tanstack`. If a per-chunk signal is still needed, keep it off the wire and off the log (strip after the producer reads it, or use a side channel that `toWireChunk` drops).

If the flag stays for this PR, at least flush compaction (and other library progress events) internally without teaching every app the durability batch, and add a test that holds `summarizeOldest` in flight and asserts `compaction:started` was appended before the summarize promise resolves — the current test uses sync `evictOldest` and only compares append order after a full drain.

AlemTuzlak:
@oztune I just went through the PR and your fix is the opposite of what should happen in the first place, instead of flushing the event with a config flag it's flushed by default and you can opt into batching now, that was the expected behavior from the beginning and the behavior that was on main was a bug!

AlemTuzlak (approved):

PR #1401, opened 2026-09-17
fix(ai): preserve cancelled and denied tool outcomes

### LINKED ISSUE #1386: Cancelled/denied tool calls are indistinguishable from genuine failures
### TanStack AI version

@tanstack/ai v0.54.0 (also v0.53.0)

### Framework/Library version

Framework-agnostic (plain Node.js)

A cancelled or user-denied tool call is emitted with the same shape as a genuine failure, so a consumer can't tell them apart. Both produce `state: 'output-error'` + `{ error: <hardcoded string> }` (`packages/ai/src/activities/chat/tools/tool-calls.ts`):

- Cancel (`:807-813`): `{ error: 'Tool execution cancelled' }`
- Deny (`:912-923`, `:1001-1009`, via `deniedApprovalResult` `:486-490`): `{ error: 'User declined tool execution' }`

That is identical to real failures — unknown tool (`:787`), arg-parse (`:826`), validation (`:848`). The only discriminator is the hardcoded English string; there's no distinct state, since `ToolResultState` is `'streaming' | 'complete' | 'error'` (`types.ts:76`).

The framework *does* know the outcome structurally — the `InterruptStore` record's status is `'cancelled'` — but that's discarded when the `TOOL_CALL_RESULT` is built, and only `{ error: <string> }` survives into the persisted `ModelMessage` a UI re-hydrates from. So after reload, matching that localizable string is the only option left.

Runnable repro (no API key) + verified output are in the linked sandbox: an approval-deny and a genuine `execute()` throw come back byte-identical except for the error string.

Expected: a structured, persistable signal for "user cancelled/declined", distinct from a failure — a terminal state, a discriminator on the result, or (minimally) exported constants for the two strings. Ideally carry through the outcome the `InterruptStore` already records.

https://stackblitz.com/github/oztune/tanstack-ai-cancel-repro

_No response_

_No response_

### Terms & Code of Conduct

mikemikimike:
Fixes #1386

Cancelled and denied tool results now keep structured outcomes across tool execution, streaming, persistence, and UI restoration. Consumers can distinguish a user decision from an ordinary tool execution failure without matching error text.

## Changes

- Preserve explicit `cancelled` and `denied` outcomes across tool execution, streaming, persistence, and UI restoration.
- Replace an existing denied tool result by `toolCallId` during resume instead of appending a duplicate result.
- Ignore lifecycle-looking fields in arbitrary tool output content. Restore outcomes only from namespaced metadata.
- Add regression coverage for denied-result replacement and ordinary output containing `{"outcome":"cancelled"}`.

## Checklist

- [x] I tested the code changes locally with focused package tests, type checks, lint, formatting, and documentation link checks.
- [x] Docs: I updated `docs/` for this user-facing change.
- [x] Changeset: I added a changeset for the published package.

## Release Impact

## Root cause

**Issue.** Approval denial and cancellation reached clients as generic tool errors, so callers could not distinguish a user decision from an execution failure.

**Cause.** Tool execution produced only the output error state. Resume, stream, and message conversion paths did not carry a structured outcome through all representations.

**Fix.** Add `ToolResultOutcome`, carry it in TanStack metadata, restore it on UI parts, and replace denied resume results by `toolCallId`.

## Possible alternatives

- **Match error text.** Callers can inspect fixed error strings, but text is not a stable lifecycle contract.
- **Add separate result states.** New states would distinguish outcomes, but would break the existing error-state compatibility contract.

### Commands

- `pnpm run test:ci` passed for the initial implementation.
- Focused Vitest run: 4 files, 434 tests passed.
- `tsc`, `oxlint`, `oxfmt --check`, `git diff --check`, and documentation link verification passed. Oxlint reported existing warnings only.
- Agent-written repro failed on clean `main` because the outcome was undefined, then passed on this branch with `denied outcome preserved`.

### Manual test

1. Start a chat with a tool that requires approval.
2. Deny the tool call.
3. Read the `tool-result` part and confirm that `state` is `error` and `outcome` is `denied`.
4. Cancel an approval request and confirm that `outcome` is `cancelled`.

### Testing artifacts

The branch includes unit tests for execution, message conversion, stream processing, and denied-result replacement. It also includes interrupt E2E coverage.

## Risk / rollback

Risk is low. The existing error state and wire format remain unchanged. Revert this PR to restore the previous behavior.

## Public API change

**Before**

```ts
const denied = message.parts.some(
  (part) => part.type === 'tool-result' && part.state === 'error',
)
```

**After**

```ts
const denied = message.parts.some(
  (part) => part.type === 'tool-result' && part.outcome === 'denied',
)
```

## Summary by CodeRabbit

* **New Features**
  * Tool results now preserve whether execution was **denied** or **cancelled** across approval, interruption, streaming, and AG-UI flows.
  * Chat interfaces display denied and cancelled results as error outcomes with appropriate status details.
  * Added public types and validation for tool-result outcomes.

* **Bug Fixes**
  * Prevented denied tools from re-executing or being duplicated in conversation history.
  * Preserved structured outcomes during message conversion and persistence.

* **Documentation**
  * Updated tool approval guidance to explain denied and cancelled results and outcome handling.


tombeckenham (changes requested):
Thanks for this. The approach is right: the outcome travels in `metadata.tanstack` and is never inferred from content, and replacing the denial message in place avoids duplicate tool results. A few things need to change before merge.

## Must fix

**1. A middleware `toolResume: 'cancel'` now turns a user's denial into "cancelled".**
`index.ts:3240-3243` stops counting denied ids as completed in `getPendingToolCallsFromMessages()`. The cancel sweep at `index.ts:4429-4432` adds every pending id to `resumeCancelledToolCallIds`, so denied ids get added too. In `tool-calls.ts:810` the cancelled check runs before the approval branch. So in a resume batch that has a denial plus a generic interrupt whose `onInterruptResolution` returns `'cancel'`, the model sees `Tool execution cancelled` / `outcome: 'cancelled'` and the denial payload is lost. On main that denial was left alone.
Fix: skip ids in `resumeDeniedToolResults` in the cancel sweep, or check denial before cancellation in `executeToolCalls`. Please add a test for this batch.

**2. Validate `toolResultOutcome` when reading it back.**
`tanstackMetadata()` only casts the value. At `messages.ts:819`, `messages.ts:1139` and `processor.ts:1675`, a persisted or client-sent value like `null`, `'foo'` or `42` turns a successful result into `'error'` (and the tool-call part too, in `modelMessagesToUIMessages`) and puts an invalid value into the typed `outcome` field. One guard used at all three sites is enough:
```ts
const isToolResultOutcome = (v: unknown): v is ToolResultOutcome =>
  v === 'cancelled' || v === 'denied'
```

**3. Make the stream path agree with the converters.**
The converters force `state: 'error'` whenever an outcome is present. `processor.ts:1673-1683` does not, so a chunk with an outcome but no `output-error` shows `{ state: 'complete', outcome: 'denied' }` while streaming and `'error'` after a reload. Set `isOutputError ||= toolResultOutcome !== undefined`.

## Should fix

- **Flaky e2e:** the cancel case in `per-item.spec.ts:140-159` waits a fixed 500ms and reads the messages once. Use `expect.poll`.
- **Weak assertion:** the replacement test in `chat.test.ts` (~L1741) only checks that one `call_denied` tool message is left. Also check that it is the new denial (its content and `metadata.tanstack.toolResultOutcome: 'denied'`), not the stale `{approved:false}` message.
- **JSDoc:** `types.ts:450`, `types.ts:545` and `tool-calls.ts:434` say "the user cancelled". Middleware can also cancel (`toolResume: 'cancel'`). It is also worth saying that `state` stays `'error'` when `outcome` is set.
- **Docs:**
  - Bump `updatedAt` for the `tool-approval` entry in `docs/config.json`.
  - `tool-approval.md:51` still says "cancelled (if denied)".
  - The new `###` section sits in the page intro, so the paragraphs after it read as part of the section. Move it to its own `##`, near "Handle denial gracefully".
  - The error text is fixed English, not "localized".
  - A short client snippet (`part.outcome === 'denied'`) would help readers.

## Simplifications (optional)

- `messages.ts:90-101`: `withTanstackMetadata` already does this merge. Use `withTanstackMetadata(part, { toolResultOutcome: part.outcome }).metadata`.
- `messages.ts:1154`: `resultState === 'error' ? 'error' : 'complete'` returns the same value. Use `toolCallPart.state = resultState`. The `resultState` derivation is also duplicated at L819 and L1139.
- `index.ts:3153-3163`: an outcome always comes with `output-error`, so spread `toolResultOutcome` into the original ternary and drop the `Object.keys` check.
- `index.ts:3182-3192`: `resumeDeniedToolResults.has(id) ? findIndex(...) : placeholderIdx` is enough.
- `testing/e2e/tests/interrupts-test/helpers.ts:165`: `getMessages` already exists in `tools-test/helpers.ts`. Reuse it. `toContainEqual(expect.objectContaining({...}))` gives a better failure diff than `.some()` plus `toBe(true)`.
- `messages.ts:1156`: a blank line was removed for no reason. Please restore it.

## Question

The legacy `ai-client` approval path (`chat-client.ts:2821-2874`) never produces `outcome: 'denied'`. Only the interrupt-resume path does. Is that intended? If so, please say so in the docs.

mikemikimike:
The current head `329b1732d4a02e6c7525b491d413daa9ee582304` includes the requested follow-up coverage and behavior fixes:

- denied tool results are excluded from middleware cancellation sweeps;
- persisted/client `toolResultOutcome` values are validated before being restored;
- streamed outcomes force the tool result state to `error`, matching the converters;
- added the mixed denied-plus-generic-cancel regression;
- strengthened denied-result replacement assertions and made the per-item e2e assertion poll for the result;
- updated the approval docs and outcome JSDoc.

Verification on the current head:
- focused Vitest: 4 files, 441 tests passed;
- `tsc -p packages/ai/tsconfig.json` passed;
- `oxfmt --check` passed;
- `git diff --check` passed;
- oxlint completed with the repository's existing `no-explicit-any` warnings only.

Please re-review the current head.

PR #1419, opened 2026-09-18
feat: add decide() and TypeSafe Jev adapters

AlemTuzlak:
You can now ask typed questions about shared state and branch on the answers in ordinary TypeScript.

Call `await decide({ adapter, state, questions })`. Each answer carries a value, a probability, and (for choice and score) a confidence number. Jev is the first evaluate model. It runs on TypeSafe, OpenRouter, Vercel AI Gateway, and Cloudflare.

## 🎯 Changes

A new `evaluate` activity in `@tanstack/ai`, a new `@tanstack/ai-typesafe` package, and Jev evaluate adapters on the three gateways.

`decide()` takes the adapter in the same call as the state and the questions, which matches `rerank()`. Questions come from `choice()`, `score()`, and `boolean()`. Answers sit on the result. Usage sits on `result.meta`.

The four adapter factories are `typesafeDecider`, `openRouterDecider`, `vercelGatewayDecider`, and `cloudflareDecider`. Each one has a `create*` variant that takes an explicit API key.

Docs: an evaluate guide, a TypeSafe adapter page, a tutorial, and an API reference update. The example app is `examples/react/evaluate`.

## ✅ Checklist

## 🚀 Release Impact

### Commands run

1. `pnpm test:pr` — passed. All targets green across 96 projects.
2. `pnpm --filter @tanstack/ai-e2e test:e2e` — passed. 703 tests passed, 14 skipped.
3. `nx run evaluate:test:types --skip-nx-cache` — passed. The example app is outside `test:pr`.
4. Live run on Vercel AI Gateway in the example app: queue `billing` p=1.00, urgency `high` score=2.00, refund `true` p=0.99, model `typesafe-ai/jev`.
5. Live run on OpenRouter in the example app: the same three answers, model `typesafe/jev-1.13-20260917`.

The two live runs happened before the adapter rename. The rename is identifier-only, and unit tests plus the typecheck cover it.

### Manual test

1. Copy `examples/react/evaluate/.env.example` to `examples/react/evaluate/.env.local`.
2. Set `OPENROUTER_API_KEY`, or another key that the file lists.
3. Run `pnpm --filter evaluate dev`.
4. Open http://localhost:3100/.
5. Pick the matching provider, paste a support ticket, then click Submit.

The panel shows the queue, the urgency, the refund answer, and the Jev model id.

Every adapter has unit tests that mock `fetch`. The example app runs a live `decide()` call against any of the four providers from one provider picker.

## Risk / rollback

Low risk. The evaluate activity is new code behind new exports. The chat, rerank, and embed paths do not change. To undo, revert the PR.

## Public API change

**Before**

```ts
// no evaluate activity
```

**After**

```ts
import { decide, choice, score, boolean } from '@tanstack/ai'
import { typesafeDecider } from '@tanstack/ai-typesafe'

const result = await decide({
  adapter: typesafeDecider('jev-latest'),
  state: ticket,
  questions: {
    queue: choice({
      instructions: 'Which team should handle this ticket?',
      options: { billing: 'Payments', tech: 'Bugs', sales: 'Pricing' },
    }),
    urgency: score({
      instructions: 'How urgent is this ticket?',
      levels: ['low', 'medium', 'high'],
    }),
    refund: boolean({
      instructions: 'Is the customer asking for a refund?',
    }),
  },
})

result.queue.value
result.queue.probability
result.meta.usage
```

## Summary by CodeRabbit

- **New Features**
  - Added typed evaluation through the direct `decide()` API.
  - Added `choice()`, `score()`, and `boolean()` question helpers with typed answers, confidence, probabilities, scores, and usage metadata.
  - Added evaluation support for TypeSafe, OpenRouter, Vercel AI Gateway, and Cloudflare.
  - Added a support-ticket evaluation example with provider selection.

- **Documentation**
  - Added evaluation guides, tutorials, adapter references, and API documentation.

- **Observability**
  - Added evaluation events and OpenTelemetry activity reporting.


tombeckenham (commented):
Review of the evaluator activity and the four Jev adapters. I'm leaving comments only, not requesting changes. Several review passes (code, silent failures, tests, types, docs, simplification) ran over the diff, and I checked the top findings against the code.

**Looks good**
- Middleware order, abort/error classification and devtools events match `rerank`.
- The value unions inferred from `const` option keys are a real type-safety gain.
- Dependency ranges, the README banner, the `docs/config.json` dates and the model ids all follow repo rules.
- API keys stay on the server in the example.

**Main themes (details inline)**
1. **Bad provider answers aren't caught.** The core never checks answer fields, Cloudflare casts the response unchecked, and Vercel makes up `confidence: 0` / `{}`. A malformed answer becomes a typed value the caller branches on. One set of checks in `mapWireAnswer` would cover every adapter.
2. **Proxy / transport options.** OpenRouter ignores `baseURL`. Vercel silently falls back to the public gateway on an invalid `baseURL` and drops path prefixes. TypeSafe's `timeout` does nothing.
3. **Docs.** The tutorial has `as` casts and a route file that isn't a Start server route. The "every adapter takes `baseURL`" claim is false for OpenRouter.

**Not inline**
- **No E2E test.** Nothing in `testing/e2e/` changed. TypeSafe and Vercel take a `baseURL`, so an aimock `mount()` fixture could cover them today. If E2E is skipped on purpose, a note in the PR body would help.
- **Tests replace the global `fetch`.** The TypeSafe, OpenRouter and Vercel adapter tests use `vi.stubGlobal('fetch')`. TypeSafe already has a `fetch` option.
- **Error paths untested.** Nothing covers a wire-type mismatch, a missing answer key or a missing probability (`index.ts:233, 297-339`). Nothing checks that `abortSignal` reaches `fetch`.
- **Agent skill.** `packages/ai/skills/ai-core` isn't updated for `evaluator()`.
- **"raw fraction" wording.** `score` is described as a "raw fraction" in `adapter.ts:84`, `index.ts:75`, and `evaluate.md:190` and `:266`. It's a fractional level index (0 to levels - 1).
- **Smaller items**
  - `evaluate.md:276` leaves out the `start` hook.
  - `evaluate.md:477-479` repeats "add a key for the adapter you pick".
  - The TypeSafe README sample uses `ticket` without defining it.
- **Simplification.** The same `!response.ok` block appears in 3 adapters, the same record guards in 3, and a usage mapper in 4. A shared `postJson`/response parser in `adapter-internals` would cut roughly 100 lines.

packages/ai/src/activities/evaluate/index.ts
```diff
@@ -0,0 +1,640 @@
...
+  if (keys.length === 0) {
+    throw new Error('decide() requires at least one question')
+  }
+  if (Object.hasOwn(questions, RESERVED_QUESTION_KEY)) {
+    throw new Error('decide() reserves the question key "meta"')
+  }
+  return keys
+}
+
+function mapChoiceAnswer(wire: WireChoiceAnswer, key: string) {
```
tombeckenham, line 232: **Provider answers are never checked here.** Every adapter routes through this code, so one set of checks here covers all four.

- `mapChoiceAnswer` never checks that `wire.choice` is one of `question.criteria`'s keys. `.value` is typed as the option union but can be any string, so `switch` exhaustiveness in user code is a lie.
- `confidence` is never checked (it can be `undefined`), and `score` is never checked with `Number.isFinite`.

Suggest: `Object.hasOwn(question.criteria, wire.choice)`, `typeof confidence === 'number'`, and `Number.isFinite(score)`, all throwing with the question key. OpenRouter's `isWireAnswer` could move here, and the per-adapter copies could be deleted.

packages/ai/src/activities/evaluate/index.ts
```diff
@@ -0,0 +1,640 @@
...
+    value,
+    probability,
+    confidence: wire.confidence,
+    score: wire.score,
+    legend: wire.legend,
+    probabilities: wire.probabilities,
+  }
+}
+
+function mapBooleanAnswer(wire: WireNoulAnswer) {
```
tombeckenham, line 286: If `noul` is missing, `undefined >= 0.5` is `false`, so the caller gets a confident "no" with `probability: undefined`. Suggest throwing when `!Number.isFinite(wire.noul)`.

packages/ai/src/activities/evaluate/index.ts
```diff
@@ -0,0 +1,640 @@
...
+  }
+}
+
+function mapScoreAnswer(
+  question: WireScoreQuestion,
+  wire: WireScoreAnswer,
+  key: string,
+) {
+  const levels = question.criteria
+  if (levels.length < 2) {
```
tombeckenham, line 254: This levels check runs after the provider call, so it fails after the request has already been paid for. `decide` accepts hand-built `WireQuestion`s, not just `score()` output. Suggest moving it into `assertQuestions` (line 221), together with a check that choice questions have at least one option (`choice({ options: {} })` gives `ChoiceAnswer<never>`).

packages/ai/src/activities/evaluate/index.ts
```diff
@@ -0,0 +1,640 @@
...
+>(config: EvaluatorConfig<TAdapter>) {
+  /**
+   * Ask typed questions about `state`. This is the network call.
+   *
+   * Pass questions built with `choice`, `score`, and `boolean`. The question
+   * key `meta` is reserved. Throws if `questions` is empty or uses that key.
+   *
+   * Per-call `debug` and `middleware` replace the values from `evaluator()`.
+   */
+  async function decide<TQuestions extends Record<string, WireQuestion>>(
```
tombeckenham, line 522: `decide` has no return annotation, so the exported `EvaluateResult` can drift from what it actually returns. Suggest annotating `Promise<EvaluateResult<TQuestions>>`.

`meta` could also be blocked at compile time with `questions: TQuestions & { meta?: never }`, not only at runtime.

packages/ai/src/activities/evaluate/index.ts
```diff
@@ -0,0 +1,640 @@
...
+    })
+
+    logger.request(`activity=evaluate provider=${adapter.name}`, {
+      provider: adapter.name,
+      model,
+      questionCount: keys.length,
+    })
+
+    try {
+      const result = await adapter.evaluate({
```
tombeckenham, line 561: Unlike summarize and the media activities, this doesn't wrap the call in `raceWithAbort(..., abortSignal)`. The Cloudflare binding path (`utils/run.ts`) never forwards `signal`, so aborting a binding-backed `decide()` does nothing.

Also nit: there's no `evaluate:request:error` devtools event (embedding has `embedding:request:error`), and `runGenerationStart` (line 544) runs outside the `try`.

packages/ai/src/activities/evaluate/index.ts
```diff
@@ -0,0 +1,640 @@
...
+
+// ===========================
+// Helper Functions
+// ===========================
+
+function createId(prefix: string): string {
+  return `${prefix}-${Date.now()}-${Math.random().toString(36).slice(2, 9)}`
+}
+
+function isAbortError(error: unknown, signal?: AbortSignal): boolean {
```
tombeckenham, line 205: Nit: this is a copy of `isAbortError` in `rerank/index.ts:102`. It could move into `error-payload.ts` next to `isAbortShapedError`. The comment also names the OpenRouter SDK's `RequestAbortedError`, but these adapters use plain `fetch`.

packages/ai/src/activities/evaluate/adapter.ts
```diff
@@ -0,0 +1,212 @@
...
+  readonly kind = 'evaluate' as const
+  abstract readonly name: string
+  readonly model: TModel
+
+  // Type-only property - never assigned at runtime
+  declare '~types': {
+    providerOptions: TProviderOptions
+  }
+
+  protected config: EvaluateAdapterConfig
```
tombeckenham, line 198: Nit: `config`, `EvaluateAdapterConfig` (L7-12) and `generateId` (L209) aren't used. All four adapters call `super({}, model)`. The constructor could take only `model`.

packages/ai-cloudflare/src/adapters/evaluate.ts
```diff
@@ -0,0 +1,97 @@
...
+      logger.request(`activity=evaluate provider=${this.name} model=${model}`, {
+        provider: this.name,
+        model,
+      })
+      const output = (await runModel(
+        this.cfConfig,
+        model,
+        { state, questions },
+        { signal: abortSignal },
+      )) as CloudflareEvaluateOutput
```
tombeckenham, line 49: **This casts the output with no check.** If `usage` is missing, you get an unclear `Cannot read properties of undefined (reading 'input_tokens')`. If the usage fields are missing, token counts are `NaN`. On REST, a response that isn't JSON comes back from `runModel` as a `Uint8Array`, and the cast lets it through.

Suggest a shape guard, like TypeSafe's `isTypesafeEvaluateResponse`. TypeSafe, Cloudflare and OpenRouter all parse the same `{ model, answers, usage }` envelope, so one shared parser in `adapter-internals` would cover all three.

packages/ai-openrouter/src/adapters/evaluate.ts
```diff
@@ -0,0 +1,232 @@
...
+import { buildHeaders, getOpenRouterApiKeyFromEnv } from '../utils'
+import type { EvaluateOptions, WireAnswer } from '@tanstack/ai/adapters'
+import type { OpenRouterClientConfig } from '../utils/client'
+import type {
+  OpenRouterEvaluateModel,
+  OpenRouterEvaluateProviderOptions,
+} from '../evaluate/evaluate-provider-options'
+
+/** OpenRouter Decisions API. Jev is not chat completions. */
+const OPENROUTER_DECISIONS_URL = 'https://openrouter.ai/api/alpha/decisions'
```
tombeckenham, line 13: **`baseURL` is ignored.** `OpenRouterClientConfig.baseURL` is accepted and then dropped, so a user routing through a proxy sends their key directly to openrouter.ai. The image and video adapters pass `baseURL` through as `serverURL`.

Suggest resolving it next to the `/v1` root: `new URL('../alpha/decisions', \`${baseURL ?? 'https://openrouter.ai/api/v1'}/\`)`. That also keeps a proxy's path prefix, and it would let an aimock E2E test hit this adapter.

`docs/adapters/typesafe.md:118` says `baseURL`/`defaultHeaders` work "on every TanStack AI adapter", which is false here.

packages/ai-openrouter/src/adapters/evaluate.ts
```diff
@@ -0,0 +1,232 @@
...
+    const { model, state, questions, abortSignal, logger } = options
+    const questionKeys = Object.keys(questions)
+
+    logger.request(
+      `activity=evaluate provider=${this.name} model=${model} questions=${questionKeys.length}`,
+      { provider: this.name, model },
+    )
+
+    try {
+      const response = await fetch(OPENROUTER_DECISIONS_URL, {
```
tombeckenham, line 139: There's no `fetch` seam, so the tests have to use `vi.stubGlobal('fetch')`. An optional `fetch?: typeof fetch` on `OpenRouterEvaluateConfig` would let the tests inject it, like Cloudflare's test does.

packages/ai-vercel-gateway/src/adapters/evaluate.ts
```diff
@@ -0,0 +1,382 @@
...
+  for (const key of keys) {
+    const value = defaultHeaders[key]
+    if (typeof value === 'string') headers[key] = value
+  }
+  return headers
+}
+
+function resolveEvaluateUrl(baseURL: string | null | undefined) {
+  if (!baseURL) return DEFAULT_EVALUATE_URL
+  try {
```
tombeckenham, line 93: **A bad `baseURL` silently falls back to the public gateway.** If `new URL(baseURL)` throws, the catch returns `DEFAULT_EVALUATE_URL`, so a mistyped proxy URL sends the user's key to ai-gateway.vercel.sh. `.origin` also drops any path prefix, so `https://proxy.example.com/vercel/v1` becomes `https://proxy.example.com/v4/ai/evaluation-model`.

Suggest letting the invalid URL throw, and resolving relative to the base: `new URL('../v4/ai/evaluation-model', \`${baseURL}/\`)`.

packages/ai-vercel-gateway/src/adapters/evaluate.ts
```diff
@@ -0,0 +1,382 @@
...
+
+function confidenceFor(
+  key: string,
+  answer: Record<string, unknown>,
+  byKey: Record<string, number>,
+) {
+  if (typeof answer.confidence === 'number') return answer.confidence
+  const fromMeta = byKey[key]
+  if (typeof fromMeta === 'number') return fromMeta
+  return 0
```
tombeckenham, line 179: This makes up a `confidence: 0` when neither the answer nor `providerMetadata` has one. Missing `probabilities` also become `{}` (L198, L214), and the core then throws "missing probability for choice", which blames the wrong layer.

Per the AI SDK's `GatewayEvaluationModel` schema, `probabilities` is optional and there's no `confidence` or `legend` on the answer. So either throw a clear Vercel-specific error here, or make `confidence` optional on the public types. A made-up 0 looks like a real, very low confidence.

packages/ai-vercel-gateway/src/adapters/evaluate.ts
```diff
@@ -0,0 +1,382 @@
...
+  const keys = Object.keys(questions)
+  for (const key of keys) {
+    const question = questions[key]
+    if (question === undefined) continue
+    mapped[key] = toGatewayQuestion(question)
+  }
+  return mapped
+}
+
+function readCount(candidates: Array<unknown>) {
```
tombeckenham, line 132: Nit: `readCount` reads four spellings of each usage field. The gateway schema only has `inputTokens`/`outputTokens`. Also, `VercelGatewayEvaluateConfig` extends OpenAI `ClientOptions`, so `fetch` and `timeout` type-check but are ignored (L260-266, L281 uses global `fetch`).

packages/ai-typesafe/src/utils/client.ts
```diff
@@ -0,0 +1,62 @@
...
+  /**
+   * Headers sent with every request. Same option name as the other adapters.
+   * Wins over `headers` when both are set.
+   */
+  defaultHeaders?: Record<string, string>
+
+  /** Alias of `defaultHeaders`. */
+  headers?: Record<string, string>
+
+  /** Request timeout in milliseconds (default: 30_000). */
```
tombeckenham, line 30: **`timeout` is documented ("default: 30_000") but nothing reads it**, so requests can hang forever. It's also listed in `docs/adapters/typesafe.md:147`. Either implement it (`combineAbortSignals` / `AbortSignal.timeout`) or delete the option and the doc line.

Nit: the `baseUrl` and `headers` aliases (L18-28) add precedence rules for no clear gain.

packages/ai-typesafe/src/adapters/evaluate.ts
```diff
@@ -0,0 +1,169 @@
...
+    })
+
+    let response: Response
+    try {
+      response = await fetchFn(`${this.baseUrl}/v1/systemone`, {
+        method: 'POST',
+        headers: {
+          Authorization: `Bearer ${this.apiKey}`,
+          'Content-Type': 'application/json',
+          ...this.headers,
```
tombeckenham, line 84: `...this.headers` is spread after `Authorization`, so a `defaultHeaders` entry silently replaces the auth header. Vercel spreads its extra headers first. Suggest matching that.

packages/ai-typesafe/src/adapters/evaluate.ts
```diff
@@ -0,0 +1,169 @@
...
+        }`,
+      )
+      logger.errors(`${this.name}.evaluate fatal`, {
+        error,
+        source: `${this.name}.evaluate`,
+      })
+      throw error
+    }
+
+    const json: unknown = await response.json()
```
tombeckenham, line 111: `response.json()` and the shape-check throw are outside the `try`, so these failures skip `logger.errors`. That breaks the contract in `adapter.ts:118-121`. OpenRouter and Vercel wrap the whole body. (`decide()` already logs every adapter throw, so the per-adapter try/catch-log blocks could arguably all go.)

docs/tutorials/evaluate.md
```diff
@@ -0,0 +1,146 @@
...
+  subject: string
+  body: string
+}
+
+const ticketEval = evaluator({
+  adapter: typesafeEvaluator('jev-latest'),
+})
+
+export async function POST(request: Request) {
+  const body = (await request.json()) as { ticket: Ticket }
```
tombeckenham, line 61: CLAUDE.md docs rule: no `as` casts in doc samples. This line and L107 (`res.json() as Promise<...>`) both use one. `docs/evaluate/evaluate.md:311-319` already narrows the input properly.

docs/tutorials/evaluate.md
```diff
@@ -0,0 +1,146 @@
...
+type Ticket = {
+  subject: string
+  body: string
+}
+
+const ticketEval = evaluator({
+  adapter: typesafeEvaluator('jev-latest'),
+})
+
+export async function POST(request: Request) {
```
tombeckenham, line 60: A bare `export async function POST` in `src/routes/api.evaluate.ts` isn't a TanStack Start server route. It needs `createFileRoute(...)({ server: { handlers: { POST } } })`, the same as `docs/evaluate/evaluate.md:307`.

examples/react/evaluate/src/lib/server-functions.ts
```diff
@@ -0,0 +1,94 @@
...
+          questions,
+        })
+      case 'vercel':
+        return await evaluator({
+          adapter: vercelGatewayEvaluator('typesafe-ai/jev'),
+        }).decide({
+          state: ticket,
+          questions,
+        })
+      case 'cloudflare':
```
tombeckenham, line 80: `const res` is declared in a `case` with no braces, which is a `no-case-declarations` lint risk. `return await ...` directly, like the other cases.

examples/react/evaluate/src/components/EvaluatePanel.tsx
```diff
@@ -0,0 +1,163 @@
...
+      </div>
+
+      {error !== null && (
+        <div className="flex gap-3 rounded-xl border border-red-800 bg-red-950/50 p-4 text-red-200">
+          <TriangleAlert className="w-5 h-5 shrink-0 mt-0.5" />
+          <div>
+            <p className="font-medium">Evaluate failed</p>
+            <p className="text-sm text-red-300/90 mt-1">{error}</p>
+            <p className="text-sm text-red-300/70 mt-2">
+              Set <code className="font-mono">{envVarsLabel(provider)}</code> in{' '}
```
tombeckenham, line 114: The "Set <env vars> in .env and restart" hint shows for every error, including 5xx, shape errors, aborts and network failures. Suggest showing it only when the message matches the missing-key error from `getApiKeyFromEnv`.

.changeset/evaluate-typesafe.md
```diff
@@ -0,0 +1,19 @@
+---
```
tombeckenham, line 1: `@tanstack/ai-event-client` is missing from this changeset. It's published and gains three evaluate events. Also, "Answers sit on the result (value, probability, confidence)" isn't true for `boolean`, which has no `confidence`.

